Issue #55, opened 2022-11-15
make readthedocs documentation awesome


drbenvincent:
#53 added sphinx documentation and got it working reliably on readthedocs. This issue focuses on improving the docs and making them look nicer!

- [ ] Better page title
- [ ] Add some more compelling content, taking inspiration from the GitHub README. This should include images.
- [ ] Rather than itemise individual examples in the sidebar, switch to having one page of example. So there will be one Examples entry in the sidebar which leads to a single page which will expand with examples over time. 
- [ ] Remove module index and search page 
- [ ] Add info relating to PyMC Labs supporting the package

drbenvincent:
After a lot of faffing about with remote build of readthedocs, we have the new structure (with separated small notebooks) up and running https://causalpy.readthedocs.io

PR #64, opened 2022-11-16
Split the notebooks into separate examples + update toctree

drbenvincent:
Partial progress on #55.


drbenvincent:
Trying it when we provide the `requirements.txt` file into the admin area of readthedocs.
Screenshot 2022-11-16 at 11 34 20

That alone does not work...

Experimenting with removing the mocks (`autodoc_mock_imports`) from `conf.py`.

drbenvincent:
No, removing `autodoc_mock_imports` from conf.py doesn't work as that was autodoc specific. This is a _new_ problem in executing the notebooks.

Issue #66, opened 2022-11-16
type hints + mypy


drbenvincent:
Add type hints and run mypy.

Padarn:
Hyy @drbenvincent I'd be happy to lend a hand here if you like. I assume this issue is meant to cover adding mypy testing? 

>  Fix numpy type hint here

Presumably to use `nptyping`?

drbenvincent:
Hi @Padarn. Sure, I've not used mypy so I was going to learn that and improve the type hinting. But if you wanted to submit a PR, that would be great. It probably makes sense to add some short usage instructions in `CONTRIBUTING.md` as well.

My comment about numpy typing is a little cryptic, but I think I must have meant to add shape information (if required). But perhaps that's optional.

Padarn:
Haha got it. Sure let me take a stab at it.

Issue #22, opened 2022-10-26
replace PyMC linear model with Bambi model


drbenvincent:
**We are not removing custom PyMC models.** It makes a lot of sense to be able to write custom PyMC models, for maximum flexibility.

But for the majority of cases, a linear model will be used. Because of this, it doesn't make sense to duplicate all the work that Bambi does in terms of specifying custom priors and handling hierarchical model formulae.

So need to figure out how to use Bambi models ~instead~ in addition.

* To fit with the `scikit-learn` API we need to be able to pass in a blank model object, so the creation of that will have to happen behind the scenes. So maybe we have a `PymcModel` class as a wrapper around a Bambi model.

* A better idea would be to have `ModelBuilder` subclass the Bambi model class, not `pm.Model`.

drbenvincent:
Having chatted with @NathanielF, we'll close this issue. The original motivation was the desire to provide custom priors, and Bambi already has the machinery for that. Though we think that integrating Bambi into the code structure will involve some pretty big API changes.

I'm putting together an alternative issue focussed on allowing the user to define custom priors.

PR #109, opened 2022-12-03
Add tests Model Builder

juanitorduz:
This PR adds some basic test for the model builder class. I also fixed the problem of having mutable structures as default, see for example https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/
I also started adding more strict type hints as I strongly recommend to start using `mypy` for static type checking now as it s still a young project :)

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,2 +1,2 @@
 [settings]
-known_third_party = arviz,matplotlib,numpy,pandas,patsy,pymc,scipy,seaborn,setuptools,sklearn,statsmodels,xarray
+known_third_party = arviz,matplotlib,numpy,pandas,patsy,pymc,pytest,scipy,seaborn,setuptools,sklearn,statsmodels,xarray
--- a/causalpy/pymc_models.py
+++ b/causalpy/pymc_models.py
@@ -1,5 +1,5 @@
-from typing import Dict
-
+from typing import Any, Dict, Optional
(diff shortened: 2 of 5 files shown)


drbenvincent:
Thanks @juanitorduz, I'll take a proper look at this very soon, but from a quick glance it looks good. 

I was planning on working on #22 soon. So Bambi models will be used for most/all linear model situation. But that won't make `ModelBuilder` defunct - we still want the ability to have custom PyMC models.

And agree with `mypy`, it's on the list (#66) 🙂

juanitorduz:
ok! When you start working in https://github.com/pymc-labs/CausalPy/issues/22, maybe this PR can be used as a initial gide to write test while developing :) I will of course be happy to support with the tests!

drbenvincent (commented):
This all looks good. The only thing I don't really follow is what's going on in `causalpy/tests/conftest.py`. Could you just briefly explain, then happy to merge.

juanitorduz:
Sure! The conftest.py file contains global pytest fixtures. In this example rng is a random generator which can be used in all tests and is just defined once. Other use cases to store in this files are toy data as pandas data frames for example. More info https://stackoverflow.com/questions/34466027/in-pytest-what-is-the-use-of-conftest-py-files

drbenvincent (approved):

Issue #25, opened 2022-10-26
Add tests


drbenvincent:
- [x] integration tests for sklearn examples in the docs
- [x] integration tests for pymc examples in the docs. If these are slow we might not want to run these tests remotely for every push? (addressed by #119)
- [ ] ~~bunch of unit tests~~ Maybe best wait until the code base is more stable. Otherwise it becomes harder to make change.
- [ ] ~~test for custom exceptions etc~~ Not specific enough
- [x] test we can load in the datasets (addressed by #119)

PR #118, opened 2022-12-07
add integration tests covering the sklearn examples from the docs

drbenvincent:
My attempt at starting #25. Here are some integration tests to cover just the sklearn based models. So far I've added a test for every example in the docs.

I've also given these a custom mark of `integration` and we can just run those tests alone with `pytest -v -m integration`.

All tests pass.

Let me know if you have suggested changes. After the approach for integration tests is sorted then I'll add equivalent tests for the pymc based examples.

### DIFF
--- a/causalpy/tests/test_integration_skl_examples.py
+++ b/causalpy/tests/test_integration_skl_examples.py
@@ -10,14 +10,14 @@ import causalpy as cp
 @pytest.mark.integration
 def test_did():
     data = cp.load_data("did")
-    assert type(data) is pd.DataFrame
     result = cp.skl_experiments.DifferenceInDifferences(
         data,
         formula="y ~ 1 + group + t + treated:group",
         time_variable_name="t",
         prediction_model=LinearRegression(),
     )
-    assert type(result) is cp.skl_experiments.DifferenceInDifferences
+    assert isinstance(data, pd.DataFrame)
+    assert isinstance(result, cp.skl_experiments.DifferenceInDifferences)
 
 
 @pytest.mark.integration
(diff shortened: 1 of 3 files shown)


causalpy/tests/test_integration_skl_examples.py
```diff
@@ -0,0 +1,109 @@
...
+from sklearn.gaussian_process.kernels import ExpSineSquared, WhiteKernel
+from sklearn.linear_model import LinearRegression
+
+import causalpy as cp
+
+
+@pytest.mark.integration
+def test_did():
+    data = cp.load_data("did")
+    assert type(data) is pd.DataFrame
```
juanitorduz, line 13: I think is better to use (apply to all of the above where `type(...)` is used)
```python
assert isinstance(data, pd.DataFrame)
```

causalpy/tests/test_integration_skl_examples.py
```diff
@@ -0,0 +1,109 @@
...
+
+@pytest.mark.integration
+def test_rd_drinking():
+    df = (
+        cp.load_data("drinking")
+        .rename(columns={"agecell": "age"})
+        .assign(treated=lambda df_: df_.age > 21)
+        .dropna(axis=0)
+    )
+    assert type(df) is pd.DataFrame
```
juanitorduz, line 31: Ideally, all the `assert` statements should be at the end of the test function: Prepare all input and finally have the assert statements.

juanitorduz (changes requested):
Hey! Nice that we are adding tests. Left a few suggestions.

causalpy/tests/test_integration_skl_examples.py
juanitorduz: Or even consider splitting in two tests :)

causalpy/tests/test_integration_skl_examples.py
drbenvincent: Sure. That's done now

causalpy/tests/test_integration_skl_examples.py
drbenvincent: Done 👍

juanitorduz (approved):

PR #119, opened 2022-12-08
Add more tests: loading of package datasets and integration tests of pymc examples

drbenvincent:
Contributes to #25.

Following the merge of integration tests of the sklearn examples, I've added integration tests for the pymc examples in the docs.

I've also added tests for loading of the datasets included in the package.

The problem with the pymc integration tests is that they take a long time to run. Remotely it was taking over 10 mins **(I ended up cancelling the tests)**, and that's done across multiple python versions. So as things stand, this is going to eat up a lot of the GitHub actions time budget.

I think there are a few options. Happy to take advice:
1. Add the ability to provide sampling kwargs and get these to only generate a few samples and only 1 chain. I guess we do need this anyway. Despite adding the Bambi model #22 we still want custom pymc models.
2. Change it so that the tests only get triggered on a pull request. Even so the tests take too long to be useful. It's less bad running them locally because we have more cores and fast machines, but running the tests remotely takes way too long.

I think the first option is probably the way forward. What do you think?

### DIFF
--- a/causalpy/tests/test_integration_pymc_examples.py
+++ b/causalpy/tests/test_integration_pymc_examples.py
@@ -3,6 +3,8 @@ import pytest
 
 import causalpy as cp
 
+sample_kwargs = {"tune": 20, "draws": 20, "chains": 2, "cores": 2}
+
 
 @pytest.mark.integration
 def test_did():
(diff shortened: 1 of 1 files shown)


drbenvincent:
@juanitorduz So far I've implemented all the suggestions aside from checking the loaded datasets have the required columns. Happy to do that if you think it's important, but tests will fail anyway if they don't have the required columns, just not from a specific test for column presence. Would probably be pretty easy to tell why from the test logs. Let me know what you think.

Tests pass, and in total they seem to take ~4 mins remotely. That compares to ~2 min historical runs before the pymc integration tests. I think this is reasonable.

juanitorduz (approved):
This is a great start to increasing our test cover! In a future PR we could look into testing the logic. For example, have a dummy dataset where there is no effect and test we are not detecting anything... Also edge cases ;)

Issue #120, opened 2022-12-09
Evaluate docstring coverage with `interrogate`


drbenvincent:
The idea is to prompt increased docstring coverage by empirically measuring it.
- [ ] Use [`interrogate`](https://interrogate.readthedocs.io/en/latest/) to check for missing docstrings.
- [ ] Set up to run in pre-commit, including badge generation
- [ ] Show badge in README

PR #124, opened 2022-12-18
use interrogate to assess docstring coverage

drbenvincent:
Addresses #120.
* adds `interrogate` to `requirements-dev.txt`
* adds options to `pyproject.toml`
* adds badge to `img/` and displayed in the README
* added to pre-commit checks


.pre-commit-config.yaml
```diff
@@ -29,3 +29,7 @@ repos:
       rev: 3.9.2
       hooks:
           - id: flake8
+    - repo: https://github.com/econchick/interrogate
+      rev: 1.5.0
+      hooks:
+          - id: interrogate
```
juanitorduz, line 35: I also think we need to add this into the Makefile (`check_lint`?) to make it run inside github acations.

juanitorduz (changes requested):

drbenvincent:
At the moment the lint tests fail due to `interrogate`. I _think_ this is because we have under the threshold level of docstring coverage as opposed to a shoddy job of putting this into the GitHub actions in `.github/worksflows/ci.yml`.

Note, at the moment the badge is generated, but it is left on the random ubuntu instance. It would need to be committed to the repo. See here https://github.com/marketplace/actions/python-interrogate-check for info on how to do that.

But perhaps it is easier for that to be done locally, so contributors upload edits to the interrogate badge svg. This will happen because interrogate is in the pre-commit.

drbenvincent:
interrogate still failing. I think the remote run of interrogate is not paying attention to the config options in `pyproject.toml`?

drbenvincent:
pfft. interrogate got removed from the `requirements-lint.txt`. So I've remove the GitHub action and just call it from command line in the makefile

drbenvincent:
I think this is now good @juanitorduz 🎉

juanitorduz (approved):

PR #125, opened 2022-12-18
fix notebook formatting

### LINKED ISSUE #40: black formatting of notebooks not happening in pre-commit checks
- [x] add `black-jupyter` to pre-commits
- [ ] check `isort` working in notebooks
  - [ ] update to use https://github.com/pycqa/isort. The current one listed is archived.
  - [ ] remove https://github.com/asottile/seed-isort-config. This is deprecated.
  - [ ] might be best to use [`nbqa`](https://github.com/nbQA-dev/nbQA)

drbenvincent:
Should close #40.

* updated `.pre-commit-config.yaml`
* now uses `nbqa` for notebooks. `nbqa` added to `requirements-dev.txt`
* includes formatting of the notebooks with `nbqa`


juanitorduz (changes requested):
I think we also need to update the Makefile (so that this is also checked in github actions).
We need to add in the `check_lint` the same steps you added in the pre-commit change. Something like:
```
nbqa black --check .
nbqa isort --check-only .
```

drbenvincent:
That should hopefully do it

juanitorduz (approved):
great! :) It seems there is a conflict of a notebook from the master branch. Once you resolve it I think we can merge this one.

Issue #143, opened 2022-12-27
Be consistent with plotting CI’s


drbenvincent:
At the moment we have a mix of posterior expectations and posterior predictive distributions being plotted. And these are not labelled, so interpretation is not clear. Switch to posterior expectation by default. Maybe allow posterior predictive as an option?

this also goes for the calculation of causal impact.

PR #145, opened 2022-12-27
Improve plot clarity with combined mean and HDI legend elements

### LINKED ISSUE #142: Improve plots with combined line + credible region in legend
Combining the line and credible intervals would significantly simplify the plots, and improve their clarity.

drbenvincent:
Closes #142 and #143.

* Previously the figure legends were quite messy. The main thing this PR does is to create single legend entries which include both the central tendency and the HDI. This is actually a bit involved and so we now build up a list of handles and labels in the plotting functions. With this we can call `ax.legend` once at the end of the plot functions. See my [StackOverflow question](https://stackoverflow.com/questions/74930795/is-it-possible-to-iteratively-build-up-legend-entries-through-multiple-plot-func/74931052?noredirect=1#comment132232774_74931052) and [post on Twitter](https://twitter.com/inferencelab/status/1607751718011375618?s=20&t=mUj9-CFRVakTDbdZugznpQ).
* I updated the plot methods for all the pymc experiments to use this new approach. There is certainly still scope for slight improvement, but I think this is a good step forward in improving the plots.

# Examples for synthetic control plots...

## Before
Screenshot 2022-12-27 at 20 50 32

## After
Note the combined line+shaded region in the legend
Screenshot 2022-12-27 at 20 45 28

# Examples for difference in differences

## Before
notebooks_did_pymc_banks_17_0-1

## After
notebooks_did_pymc_banks_17_0


drbenvincent:
- [ ] I should update the plots in README

drbenvincent:
I originally tried to do this on a better way, with everything handled internally in the plot function. But I ran into problems. It may be that I discovered a bug in matplotlib. See the solution here https://stackoverflow.com/a/74935897 If this can be made to work, it would be much cleaner.

causalpy/plot_utils.py
```diff
@@ -13,28 +14,46 @@ def plot_xY(
     ax: plt.Axes,
     plot_hdi_kwargs: Optional[Dict[str, Any]] = None,
     hdi_prob: float = 0.94,
+    label: Optional[str] = "",
     include_label: bool = True,
-) -> None:
+):
```
juanitorduz, line 19: type hints are missnig

causalpy/plot_utils.py
```diff
@@ -13,28 +14,46 @@ def plot_xY(
     ax: plt.Axes,
     plot_hdi_kwargs: Optional[Dict[str, Any]] = None,
     hdi_prob: float = 0.94,
+    label: Optional[str] = "",
```
juanitorduz, line 17: Remember `Optional[str]` is equivalent to `Union[str, None`. I think we probably do not want this so we should remove `Optional`. See https://github.com/pymc-labs/CausalPy/pull/117

causalpy/plot_utils.py
```diff
@@ -13,28 +14,46 @@ def plot_xY(
...
+        label=f"{label}" if include_label else None,
+    )
+    ax_hdi = az.plot_hdi(
         x,
         Y,
         hdi_prob=hdi_prob,
         fill_kwargs={
             "alpha": 0.25,
-            "label": f"{hdi_prob*100}% HDI" if include_label else None,
+            "label": " ",  # f"{hdi_prob*100}% HDI" if include_label else None,
```
juanitorduz, line 38: shall we start removing commented code? this makes the code hard to maintain.

causalpy/plot_utils.py
```diff
@@ -13,28 +14,46 @@ def plot_xY(
...
+        filter(lambda x: isinstance(x, PolyCollection), ax_hdi.get_children())
+    )[-1]
+
+    # if include_label:
+    #     handles, labels = ax.get_legend_handles_labels()
+    #     ax.legend(
+    #         handles=[(h1, h2) for h1, h2 in zip(handles[::2], handles[1::2])],
+    #         # labels=[l1 + " + " + l2 for l1, l2 in zip(labels[::2], labels[1::2])],
+    #         labels=[l1 for l1 in labels[::2]],
+    #     )
```
juanitorduz, line 58: same here regarding commented code

causalpy/pymc_experiments.py
```diff
@@ -118,31 +118,47 @@ def plot(self):
...
             y1=az.extract(
-                self.post_pred, group="posterior_predictive", var_names="y_hat"
+                self.post_pred, group="posterior_predictive", var_names="mu"
             ).mean("sample"),
             y2=np.squeeze(self.post_y),
-            color="C0",
+            color="C2",
             alpha=0.25,
-            label="Causal impact",
+            # label="Causal impact",
```
juanitorduz, line 157: remove this duplicated line

causalpy/pymc_experiments.py
```diff
@@ -189,10 +209,14 @@ def plot(self):
                 ls="-",
                 lw=3,
                 color="r",
-                label="Treatment time",
+                # label="Treatment time",
```
juanitorduz, line 212: remove commented line?

juanitorduz (changes requested):
I added some small suggestions. Mostly regarded commented code. This is very hard to maintain and we always have the git history to come back to them anyway. In addtion,  what if we put the legends outside the figure as described in https://stackoverflow.com/questions/4700614/how-to-put-the-legend-outside-the-plot ?

causalpy/plot_utils.py
drbenvincent: updated in an upcoming commit

causalpy/plot_utils.py
drbenvincent: Changed to `Union[str, None]`. This also allowed me to remove a kwarg

causalpy/plot_utils.py
drbenvincent: Fixed in an upcoming commit

causalpy/plot_utils.py
drbenvincent: Fixed in an upcoming commit

causalpy/pymc_experiments.py
drbenvincent: Fixed in an upcoming commit

causalpy/pymc_experiments.py
drbenvincent: Fixed in an upcoming commit

drbenvincent:
Thanks for the comments @juanitorduz. I believe I've dealt with all those changes _except_ for an external axis legend. I agree that it's suboptimal for the synthetic control plots, but I'll improve it in a separate PR.

I also re-ran some notebooks because I was not fond of the green shaded region for the causal impact.

causalpy/pymc_experiments.py
```diff
@@ -137,7 +136,6 @@ def plot(self):
             self.datapost.index,
             self.post_pred["posterior_predictive"].mu,
             ax=ax[0],
-            include_label=False,
             # label="Synthetic control",
```
juanitorduz, line 139: shall we remove this line?

juanitorduz (approved):

juanitorduz:
Just left a comment regarding a commented line: either remove or uncomment ;) Feel free to merge.

causalpy/pymc_experiments.py
drbenvincent: will do

PR #152, opened 2023-01-05
Add input validation, custom exceptions, and associated tests

### LINKED ISSUE #148: Synthetic control: `IndexError` in `.plot()` method when using a data frame with `datetime` index
Hi,

first thanks for your great work with CausalPy!

I experienced the following issue:

When using a data frame with `datetime`-typed index the `.plot()` method of `cp.pymc_experiments.SyntheticControl` class returns `IndexError`.

Code:

```python
data = ... # A data frame with datetime-type index

treatment_index = '2022-10-28'

# Build the model
model = cp.pymc_models.WeightedSumFitter(sample_kwargs={'target_accept': 0.95})

# Run the experiment
results = cp.pymc_experiments.SyntheticControl(
    data,
    treatment_index,
    formula="y ~ 0 + x1 + x2 + x3",
    model=model,
)

# Plot results
results.plot(plot_predictors=True)
plt.show()

```

This results in the following error:

```python
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\axis.py:1665, in Axis.convert_units(self, x)
   1664 try:
-> 1665     ret = self.converter.convert(x, self.units, self)
   1666 except Exception as e:

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\dates.py:1980, in _SwitchableDateConverter.convert(self, *args, **kwargs)
   1979 def convert(self, *args, **kwargs):
-> 1980     return self._get_converter().convert(*args, **kwargs)

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\dates.py:1909, in DateConverter.convert(value, unit, axis)
   1903 """
   1904 If *value* is not already a number or sequence of numbers, convert it
   1905 with `date2num`.
   1906 
   1907 The *unit* and *axis* arguments are not used.
   1908 """
-> 1909 return date2num(value)

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\dates.py:449, in date2num(d)
    448     return d
--> 449 tzi = getattr(d[0], 'tzinfo', None)
    450 if tzi is not None:
    451     # make datetime naive:

IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed

The above exception was the direct cause of the following exception:

ConversionError                           Traceback (most recent call last)
Cell In[49], line 2
      1 # Plot results
----> 2 results.plot(plot_predictors=True)
      3 plt.show()

File ~\anaconda3\envs\causal-pymc\lib\site-packages\causalpy\pymc_experiments.py:231, in SyntheticControl.plot(self, plot_predictors)
    229 def plot(self, plot_predictors=False):
    230     """Plot the results"""
--> 231     fig, ax = super().plot()
    232     if plot_predictors:
    233         # plot control units as well
    234         ax[0].plot(self.datapre.index, self.pre_X, "-", c=[0.8, 0.8, 0.8], zorder=1)

File ~\anaconda3\envs\causal-pymc\lib\site-packages\causalpy\pymc_experiments.py:200, in TimeSeriesExperiment.plot(self)
    198 # Intervention line
    199 for i in [0, 1, 2]:
--> 200     ax[i].axvline(
    201         x=self.treatment_time,
    202         ls="-",
    203         lw=3,
    204         color="r",
    205     )
    207 ax[0].legend(
    208     handles=(h_tuple for h_tuple in handles),
    209     labels=labels,
    210     fontsize=LEGEND_FONT_SIZE,
    211 )
    213 return (fig, ax)

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\axes\_axes.py:825, in Axes.axvline(self, x, ymin, ymax, **kwargs)
    822 xmin, xmax = self.get_xbound()
    824 # Strip away the units for comparison with non-unitized bounds.
--> 825 xx, = self._process_unit_info([("x", x)], kwargs)
    826 scalex = (xx < xmin) or (xx > xmax)
    828 trans = self.get_xaxis_transform(which='grid')

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\axes\_base.py:2603, in _AxesBase._process_unit_info(self, datasets, kwargs, convert)
   2601             if dataset_axis_name == axis_name and data is not None:
   2602                 axis.update_units(data)
-> 2603 return [axis_map[axis_name].convert_units(data)
   2604         if convert and data is not None else data
   2605         for axis_name, data in datasets]

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\axes\_base.py:2603, in <listcomp>(.0)
   2601             if dataset_axis_name == axis_name and data is not None:
   2602                 axis.update_units(data)
-> 2603 return [axis_map[axis_name].convert_units(data)
   2604         if convert and data is not None else data
   2605         for axis_name, data in datasets]

File ~\anaconda3\envs\causal-pymc\lib\site-packages\matplotlib\axis.py:1667, in Axis.convert_units(self, x)
   1665     ret = self.converter.convert(x, self.units, self)
   1666 except Exception as e:
-> 1667     raise munits.ConversionError('Failed to convert value(s) to axis '
   1668                                  f'units: {x!r}') from e
   1669 return ret

ConversionError: Failed to convert value(s) to axis units: '2022-10-28'
```

drbenvincent:
If the index of the input dataframe is of type `pd.DatetimeIndex` and `treatment_time` is _not_ `pd.Timestamp` then we get an error when it comes to plotting the intervention time. `treatment_time` has to be `pd.Timestamp` to know where to place the vertical line (representing treatment time) on the x axis.

This PR adds some input validation and a regression test to confirm that we get an `AssertionError`.

Closes #148


drbenvincent:
Tests pass locally. But fail remotely, under 3.8. I think this is because it doesn't like the following type hints? `treatment_time: int | float | pd.Timestamp`.

My personal preference is to take advantage of new language features and drop support for older versions, but I realise there might be advantages in supporting older versions. Otherwise I guess we change the type hint to `treatment_time: Union[int,float,pd.Timestamp]`. What do you think @juanitorduz?

juanitorduz:
I think python 3.8 is still active (until almost two years see https://endoflife.date/python) so I strongly suggest using the valid type hints for now (i.e. `Union[int,float,pd.Timestamp]`). Otherwise, we could lose some users.

causalpy/pymc_experiments.py
```diff
@@ -54,12 +54,23 @@ class TimeSeriesExperiment(ExperimentalDesign):
...
         data: pd.DataFrame,
-        treatment_time: int,
+        treatment_time: int | float | pd.Timestamp,
         formula: str,
         model=None,
         **kwargs,
     ) -> None:
         super().__init__(model=model, **kwargs)
+
+        # Input validation
```
juanitorduz, line 66: I suggest we add this into a private function

```python
def _validate_input_indices(self, ...):
    ...
```
and simply call it inside the `__init __` method.

causalpy/pymc_experiments.py
```diff
@@ -54,12 +54,23 @@ class TimeSeriesExperiment(ExperimentalDesign):
     def __init__(
         self,
         data: pd.DataFrame,
-        treatment_time: int,
+        treatment_time: int | float | pd.Timestamp,
```
juanitorduz: I suggest using `Union[int,float,pd.Timestamp]` to keep up the support of python 3.8

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -191,6 +191,21 @@ def test_sc():
     assert len(result.idata.posterior.coords["draw"]) == sample_kwargs["draws"]
 
 
+@pytest.mark.integration
+def test_sc_input_error():
+    """Test that an error is raised if the data index is not datetime and the
+    treatment time is pd.Timestamp."""
+    with pytest.raises(AssertionError):
```
juanitorduz, line 198: you can use the `match` argument to specify the `AssertionError` string so that is clear which error is expected.

causalpy/pymc_experiments.py
```diff
@@ -54,12 +54,23 @@ class TimeSeriesExperiment(ExperimentalDesign):
...
         formula: str,
         model=None,
         **kwargs,
     ) -> None:
         super().__init__(model=model, **kwargs)
+
+        # Input validation
+        if isinstance(data.index, pd.DatetimeIndex):
+            assert isinstance(
```
juanitorduz, line 68: I suggest to raise an error than using `assert`. In general is not a good practice, see https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html#b101-test-for-use-of-assert

you could define
```
class BadIndexException(Exception):
    pass
```
and raise it when the desired condition is violated.

juanitorduz (changes requested):

causalpy/pymc_experiments.py
drbenvincent: Already made this change in a commit just now :)

causalpy/pymc_experiments.py
drbenvincent: God idea. 

We already have some input validation in the other classes, so I'm thinking a more general input validation method.

drbenvincent: Done

causalpy/tests/test_integration_pymc_examples.py
drbenvincent: I'm assuming this suggestion is moot, given the suggestion below to raise a custom exception?

causalpy/pymc_experiments.py
drbenvincent: Done. This is my first time using custom exceptions, but it seems to all be working ok.

causalpy/pymc_experiments.py
```diff
@@ -361,6 +360,30 @@ def __init__(
             if "post_treatment" in label and self.group_variable_name in label:
                 self.causal_impact = self.idata.posterior["beta"].isel({"coeffs": i})
 
+    def _input_validation(self):
+        """Validate the input data for correctness"""
+        assert (
```
juanitorduz, line 365: Here we could also have custom exceptions as above.

drbenvincent:
I've now replaced all the asserts in the input validation code with custom exceptions, `FormulaException` and `DataException`.

These are evaluated with utility functions. These themselves are tested.

I will create another issue about adding tests to check that these exceptions are raised in cases where bad data and/or model formulas are provided.

drbenvincent:
I think we're getting some conflicting order of imports changes in `pymc_experiments.py`. 

Running the pre-commit checks locally reports that imports in the file are updated. But something is then undoing those changes. Pfft.

From the Makefile
```
lint:
    pip install -r requirements-lint.txt
    isort .
    black .
```
So it's likely that `isort` is doing something which `black` is then reversing.

causalpy/pymc_experiments.py
drbenvincent: done :)

drbenvincent:
Tests for the new custom exceptions for data validation are pending...

drbenvincent:
I think this is reasonable now. We now have good input validation, done in private methods, and we now have custom exceptions rather than asserts. 

We also have tests that these exceptions are thrown when the data and/or formula are not correct.

We did not have any input validation for `RegressionDiscontinuity`, but that is now added.

juanitorduz (approved):
Nice!

Issue #155, opened 2023-01-06
try to get dynamic version in `pyproject.toml`


drbenvincent:
AFTER #105...

In theory we should be able to have ONE version number in `causalpy/version.py` and get `pyproject.toml` to extract that. See the instructions here, https://packaging.python.org/en/latest/guides/single-sourcing-package-version/

But when working on #105 this was not working. So at the moment we have the version number in both `causalpy/version.py` and and `pyproject.toml`.

OriolAbril:
I use flit in xarray-einstats and the dynamic version is working: https://github.com/arviz-devs/xarray-einstats/blob/main/pyproject.toml. (I used https://github.com/astrojuanlu/cookiecutter-pylib as template in case it helps)

jpreszler:
I first tried to get setuptools to work, but the imports in `__init__.py` caused issues (either `pip install -e .` would fail with missing modules, or it would finish but the usage of the package would need to change, i.e. `result = cp.pymc_experiment.SyntheticControl(...)` wouldn't work).

Changing to flit, based on the above suggestion seems to work with minimal changes but I'm not sure show to test besides `pip install -e .` and `python -m build` and checking.

Also it seems @drbenvincent and @lucianopaz didn't like a previous attempt that changed the build system to `setuptools-scm`.  It seems the same issues don't apply to Flit, but there could be other concerns.

drbenvincent:
There is also [bump2version](https://github.com/c4urself/bump2version/#installation) as a possibility.

anevolbap:
Gave this a try in #850 using `[tool.setuptools.dynamic]` with the existing setuptools backend.

cetagostini:
Completed on `pymc6_and_pymcmarketing1_migration` via #850 (`15173c15`): setuptools project metadata now derives from `causalpy.version.__version__`, leaving one version source. The separate 1.0.0 value bump remains part of #1048.

PR #156, opened 2023-01-06
`setup.py` is dead, long live `pyproject.toml`

### LINKED ISSUE #105: `setup.py`: fix or kill
Changes/improvements to `setup.py` from @juanitorduz in #79 ended up breaking the pypi build when I released version 0.0.4. I quickly fixed this with #99 by rolling back changes to `setup.py`.

At the moment, it's not clear if the error was coming from how the version is dealt with (line below works fine and is good)

https://github.com/pymc-labs/CausalPy/blob/127f384df34879dfc86c33aa9de9f49ea3b941b9/setup.py#L17

or changes to how requirements files were being read in and dealt with.

Thoughts:
- Can we actually entirely remove `setup.py` and just have `pyproject.toml`?
- If not, then:
  -  We should improve versioning, perhaps by using https://github.com/peritus/bumpversion, which will make it easier to automate the version release process.
  - See if we can revert back to @juanitorduz's improvements to how requirements were deal with, and not break the pypi build.

drbenvincent:
Closes #105. This PR greatly simplifies things. We've moved to a more modern workflow where all the config stuff is contained in `pyproject.toml`. 

So I've been able to delete:
- `setup.py`
- `.isort.cfg`
- `manifest.in`
- `requirements.txt`
- `requirements-dev.txt`
- `requirements-docs.txt`
- `requirements-lint.txt`
- `requirements-test.txt`

The requirements are now in `pyproject.toml`. Devs can install by running `pip install causalpy[dev]` for example. 

Instructions to developers have been updated in `CONTRIBUTING.md`

Tests pass locally and remotely.

I've been through a build, upload to pypi, install process and can confirm it's pip installable in a temp local environment.

Oh, and the dev environment now uses pymc>=5.0.0 🎉

PS. When I was testing, I uploaded to pypi with version 0.0.10a, thinking that is counted as a pre-release. But I think it's the case that anyone doing `pip install causalpy` will already be getting this version not yet merged into `main`.

## Remaining things
* The only downside to the current situation is that the version number is in _two_ places. But this may be fixable with #155. If not, it will be easy if we use some auto version bump tool. Otherwise manual is fine for the moment.
* `setup.cfg` remains now because `flake8` doesn't play ball with `pyproject.toml`. There may be something we can do about this in the future.


juanitorduz (approved):
great!

PR #164, opened 2023-01-20
Fix rtd builds

### LINKED ISSUE #158: remote build of docs is broken
After switching to `pyproject.toml` (in #156) the remote build of the docs is broken. Fix it :)

OriolAbril:
I noticed docs weren't being build from main. This updates the readthedocs config so it takes into account pyproject and extra pip installs with `[]` are used now instead of requirements files. closes #158

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -26,7 +26,7 @@ sphinx:
 # Optionally declare the Python requirements required to build your docs
 python:
    install:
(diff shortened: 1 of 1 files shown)


drbenvincent (approved):
Great stuff. I initially missed this getting broken when moving over to `pyproject.toml`, so fixing this was on my list. So this is great.

PR #206, opened 2023-06-07
Add the csv data files into the package data discovery

### LINKED ISSUE #202: `.csv` files not included in automated pypi release
Greetings to the pymc community, yesterday I installed causalpy to my local machine (windows 11) via anaconda prompt using the pip install github link to get the latest version, to my surprise when i try to load any of the datasets from the example I receive an error, the image will be attached below, assistance will be greatly appreciated. do let me know if I need to upload something.
cp_load

no-dataset_cp

lucianopaz:
Closes #202

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,11 +5,11 @@ requires = ["setuptools>=61.0"]
 
 # This is configuration specific to the `setuptools` build backend.
 # If you are using a different build backend, you will need to change this.
-[tool.setuptools]
-# define all packages here
-packages = ['causalpy', 'causalpy.data']
-# include package data, such as .csv files
-include-package-data = true
+[tool.setuptools.packages.find]
+exclude = ["causalpy.test*", "docs*"]
+
+[tool.setuptools.package-data]
+"causalpy.data" = ["*.csv"]
 
 [project]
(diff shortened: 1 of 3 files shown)


drbenvincent:
You should just be able to `pre-commit run --all-files` to fix the interrogate test. Hopefully

pyproject.toml
```diff
@@ -5,11 +5,11 @@ requires = ["setuptools>=61.0"]
 
 # This is configuration specific to the `setuptools` build backend.
 # If you are using a different build backend, you will need to change this.
-[tool.setuptools]
-# define all packages here
-packages = ['causalpy', 'causalpy.data']
```
drbenvincent, line 10: Are you sure about removing this `packages` line? I think this is required.

lucianopaz: Yes, this now taken care of by `packages.find`. It recursively looks for packages (folders that have an `__init__.py` in them) and includes them into the sdist and wheel distributions. This means that if at a later stage you decide to add another folder to causalpy with special functionality, you won’t have to explicitly add it here, it will get included automatically.
The only thing that I had to do was say that `find` needed to exclude the docs and tests folders, or those would have been included too.

PR #224, opened 2023-07-22
Add an `epsilon` parameter to `RegressionDiscontinuity` classes

### LINKED ISSUE #220: Regression Discontinuity: add an `epsilon` parameter
At the moment the `RegressionDiscontinuity` class has a fixed epsilon of ±0.001. That is, the estimated causal impact is assessed at ±0.001 relative to the discontinuity. Ideally we'd want this to be under user control.

- [x] add an optional `epsilon` parameter to replace the hard-wired 0.001 value.
- [x] document it
- [x] add kwarg to tests

drbenvincent:
Epsilon is the (small) distance above and below the threshold that we use to calculate the discontinuity. Previously this was hard-wired to be 0.001, but now we have a kwarg `epsilon` that can be used to override the default (still 0.001).

Added a short description in the docstring.

Added use of the kwarg into the integration tests.

Closes #220.

### DIFF
--- a/causalpy/skl_experiments.py
+++ b/causalpy/skl_experiments.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
@@ -372,7 +370,7 @@ class RegressionDiscontinuity(ExperimentalDesign):
         treatment_threshold,
         model=None,
         running_variable_name="x",
-        epsilon: Optional[float] = 0.001,
+        epsilon: float | None = 0.001,
         **kwargs,
     ):
         super().__init__(model=model, **kwargs)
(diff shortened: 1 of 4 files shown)


causalpy/skl_experiments.py
```diff
@@ -362,13 +372,15 @@ def __init__(
         treatment_threshold,
         model=None,
         running_variable_name="x",
+        epsilon: Optional[float] = 0.001,
```
juanitorduz, line 375: If the default is not `None` , the type should not have `Optional`. Remember `Optional[float]` is equivalent to `float | None`

juanitorduz (changes requested):

causalpy/skl_experiments.py
drbenvincent: Well spotted. I didn't know about that.

juanitorduz (approved):

PR #228, opened 2023-08-03
Use dynamic versioning (#155)

nkoep:
Randomly spotted this open issue and figured I'd have a go at it. Hope there are no objections to using `setuptools-scm`.


lucianopaz:
Thanks @nkoep. I have to say that I'm strongly against `setuptools-scm`. I wanted to use it to get the version automatically from git tags but its main problem is that it assumes that the files you want to package are **all the files that are tracked by version control**. This includes:

- `.github/workflow`
- `docs`
- `causalpy/tests`
- All the files in the root folder like `codecov`, `readthedocs`, `.gitignore`, etc

In practice, `setuptools-scm` completely ignores the arguments of the `[tool.setuptools.packages.find]` section of pyproject.toml and does whatever it wants. From reading a bit of issues, this was [documented back in 2017](https://github.com/pypa/setuptools_scm/issues/190) and flagged as wont-fix because they say it's `setuptools` fault to opt to include all files from git. Until there is a sane solution to this problem, I will remain strongly against `setuptools-scm`.

drbenvincent:
Definitely appreciate the effort @nkoep, but as @lucianopaz points out, `setuptools-scm` looks to have some problems, so I'll close this PR for now. Don't let that put you off making any other contributions if you like :)

Issue #129, opened 2022-12-19
increase docstring coverage


drbenvincent:
After #124 
- [ ] increase the number of docstrings, now evaluated with `interrogate`
- [ ] ratchet up `fail-under = 40` to a higher number in `pyproject.toml`

* Don't include type information because that is given in the type hints which are (or will be in the function signature)

jpreszler:
I've begun work on this to learn the package better. I should have it done within the next week - getting the example usages will take a little time.

drbenvincent:
Fantastic. Expecting the birth of our first child any moment now, so excuse any delays in reviewing the PR.

jpreszler:
There's the PR. In retrospect, I probably should have done it in chunks because it's fairly large.

Between the new addition to your family (I hope everything will go, or has gone, smoothly) and the size of this, I'm not expecting a quick review. Hopefully the fact that you wrote the code this documents means the review isn't as bad as if I was adding 1K lines of code :) 

Please don't hesitate to flag changes you would like or inconsistencies you see.

drbenvincent:
Closed by #232

PR #213, opened 2023-06-30
Add Bayesian instrumental variable estimation

NathanielF:
Working through the repo to figure out how to best integrate the multivariate approach to Bayesian instrumental regression as seen in @juanitorduz 's [blog](https://juanitorduz.github.io/iv_pymc/). This is very early stages. Just want to make sure to test the basic idea (seems to work!) and get to know the Repo. 

In addition happy to take feedback and have a conversation about choices. Will be experimenting a bit this weekend. 

Relates to Add functionality for Bayesian Instrumental Variables #212

PR #232, opened 2023-08-24
Increase docstring coverage and add doctests

jpreszler:
This addresses issue #129:
1. All methods in the api documentation have docstrings with working examples. Documentation was checked locally.
2. Interrogate has 97% coverage - only things missing are file level docstrings on some test files.
3. Interrogate failure threshold moved up to 85%

Potential future work:
* some docstrings are still minimal, so they could be expanded - such as examples of data simulation functions
* various plot methods don't have output examples in the API reference, links to static images could be added

Main items for review:
- wording of documentation
- consistency
- docstring style


causalpy/pymc_experiments.py
```diff
@@ -70,6 +175,8 @@ def __init__(
         self._input_validation(data, treatment_time)
 
         self.treatment_time = treatment_time
+        # set experiment type - usually done in subclasses
+        self.expt_type = "Pre-Post Fit"
```
jpreszler, line 179: This is the only code change. The experiment type had to be added here in order for the summary method to be run on an instance of the PrePostFit class (rather than the SyntheticControl or InterruptedTimeSeries classes).

causalpy/pymc_models.py
```diff
@@ -1,3 +1,14 @@
+"""
+Defines generic PyMC ModelBuilder class and subclasses for
+
+- WeightedSumFitter model for Synthetic Control experiments
+- LinearRegression model
+
+Models are intended to be used from inside an experiment
+class (see pymc_experiments.py). This is why the examples require some extra
```
drbenvincent, line 8: Could we change this to `(see :code:`pymc_experiments.py`)`. Or even better, link to the `causalpy.pymc_experiments` API page.

causalpy/pymc_models.py
```diff
@@ -39,12 +57,59 @@ def build_model(self, X, y, coords) -> None:
...
+        Sampling: [beta, sigma, y_hat]
+        Sampling: [y_hat]
+        Inference data with groups:
+                > posterior
+                > posterior_predictive
+                > sample_stats
+                > prior
+                > prior_predictive
+                > observed_data
+                > constant_data
```
drbenvincent, line 112: So I am not 100% sure about this, but I think one possible good idea would be:
* Not include the text output in the docstring examples. My feeling is that this adds clutter and could make things harder to keep up to date.
* The possible exception to this would be for specific methods, such as `result.summary()` whose sole job is to provide text output.

I'll invite thoughts from @lucianopaz and @juanitorduz on this.

drbenvincent (changes requested):
Firstly, thanks so much for this @jpreszler. This is amazing work. Timing wise I wanted to do a relatively quick pass to keep the ball rolling. But I've not done a thorough read of all the parameter explanations in the docstrings for example, so I will almost certainly have a few minor points about that before we merge.

One thought I had about the docstring examples was about the maintainability. I've not used this yet, but my feeling is that we will probably avoid major headaches in the future if we test these docstring examples with `doctest` (e.g. https://realpython.com/python-doctest/; https://docs.python.org/3/library/doctest.html). That may also require changes to the GitHub actions so that these tests are run locally and remotely. If we run into trouble there, then @lucianopaz may be able to give you pointers.

There's a comment in there about when/if we include text output in the docstring examples. I've noted down my first thought, but I'd definitely appreciate input from @juanitorduz and @lucianopaz on this. As I say, I've not used `doctest`, so it might be that we do need to include (some) text output.

PS: We're likely to merge #213 very soon, so there will be a couple of new classes added there.

I've rendered the docs locally with your changes and they look really great. This is an excellent contribution.

twiecki:
Wow, this is amazing -- much appreciated @jpreszler!

causalpy/pymc_models.py
jpreszler: I like this idea and agree it would reduce clutter and be easier to maintain.

causalpy/pymc_experiments.py
```diff
@@ -855,14 +1209,48 @@ def plot(self):
         ax[1].set(title="Estimated treatment effect")
         return fig, ax
 
-    def _causal_impact_summary_stat(self):
+    def _causal_impact_summary_stat(self) -> str:
+        """Computes the mean and 94% credible interval bounds for the causal impact."""
         percentiles = self.causal_impact.quantile([0.03, 1 - 0.03]).values
-        ci = r"$CI_{94\%}$" + f"[{percentiles[0]:.2f}, {percentiles[1]:.2f}]"
+        ci = r"$CI_{94%}$" + f"[{percentiles[0]:.2f}, {percentiles[1]:.2f}]"
```
jpreszler, line 1215: Same as in the DiD experiment summary above.

jpreszler:
I've cleaned up the example output, added doctest to the CI workflow and all examples pass locally. I've also added examples for the new IV model and experiment added in #213 . 

This is ready for a new review @drbenvincent when you have a chance.

drbenvincent:
Hi @jpreszler. Thanks for the updates. I'll try to carve out some time (I now have an 11 day old son now!) to review properly. But in the mean time I triggered the remote checks and it looks like we've got a failure. The test output looks like it could be just a missing import of `statsmodels`.

causalpy/pymc_models.py
juanitorduz: I also agree with Ben's comment :)

jpreszler:
@drbenvincent , it looks like statsmodels wasn't being installed into the remote environment so I added it to the dependencies. That should fix the test failure.

Congratulations on the baby boy!

jpreszler:
Looks like there's a little instability in the summary output that I'm looking into fixing.

jpreszler:
I should have all the instability addressed, the doctests passed in the remote environment on my fork as well as locally.

causalpy/pymc_models.py
```diff
@@ -1,3 +1,17 @@
+"""
+Defines generic PyMC ModelBuilder class and subclasses for
+
+- WeightedSumFitter model for Synthetic Control experiments
+- LinearRegression model
+
+Models are intended to be used from inside an experiment
+class (see `pymc_experiments.py
+<https://causalpy.readthedocs.io/en/latest/api_pymc_experiments.html>`_).
```
drbenvincent, line 9: If this link can be done using sphinx rather than a hard-coded URL, that would be even better. Not 100% sure, but this might help: https://stackoverflow.com/questions/20059953/linking-back-to-a-source-code-file-in-sphinx

causalpy/pymc_models.py
```diff
@@ -11,6 +25,13 @@
 class ModelBuilder(pm.Model):
     """
     This is a wrapper around pm.Model to give scikit-learn like API.
+
+    Public Methods
+    ---------------
+    - build_model: must be implemented by subclasses
+    - fit: populates idata attribute
+    - predict: returns predictions on new data
+    - score: returns Bayesian R^2
```
drbenvincent, line 34: We should be able get the R^2 rendered as math, https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#role-math

causalpy/pymc_models.py
```diff
@@ -73,6 +161,33 @@ def score(self, X, y) -> pd.Series:
...
+        ...             pm.Normal("y_hat", mu=mu, sigma=sigma, observed=y_)
+        >>> rng = np.random.default_rng(seed=42)
+        >>> X = rng.normal(loc=0, scale=1, size=(200, 2))
+        >>> y = rng.normal(loc=0, scale=1, size=(200,))
+        >>> model = MyToyModel(
+        ...         sample_kwargs={"chains": 2, "draws": 2000, "progressbar": False}
+        ... )
+        >>> model.fit(X, y)
+        Inference...
+        >>> round(model.score(X, y),2) # using round() to simplify doctest
```
drbenvincent, line 187: The results should be more stable, and the doctest less fussy if you pass in random_seed=rng (I believe) in the sample_kwargs dict

causalpy/pymc_models.py
```diff
@@ -85,10 +200,42 @@ def score(self, X, y) -> pd.Series:
...
+
+    .. note: Generally, the `.fit()` method should be rather than calling
+    `.build_model()` directly.
+
+    Defines the PyMC model:
+
+    - y ~ Normal(mu, sigma)
+    - sigma ~ HalfNormal(1)
+    - mu = X * beta
+    - beta ~ Dirichlet(1,...,1)
```
drbenvincent, line 214: It would be even better to get this as LaTeX formatted eqnarray. I believe we can do multi line equations with the math directive, https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#math

My preference would be to reorder the equations so the priors come first and the likelihood term is on the bottom line.

causalpy/pymc_models.py
```diff
@@ -85,10 +200,42 @@ def score(self, X, y) -> pd.Series:
...
 
     def build_model(self, X, y, coords):
-        """Defines the PyMC model"""
+        """
+        Defines the PyMC model:
+
+        - y ~ Normal(mu, sigma)
+        - sigma ~ HalfNormal(1)
+        - mu = X * beta
+        - beta ~ Dirichlet(1,...,1)
```
drbenvincent, line 236: Same point about LaTeX eqnarray here

causalpy/pymc_models.py
```diff
@@ -106,10 +253,45 @@ def build_model(self, X, y, coords):
...
+
+    .. note: Generally, the `.fit()` method should be rather than calling
+    `.build_model()` directly.
+
+    Defines the PyMC model
+
+    - y ~ Normal(mu, sigma)
+    - mu = X * beta
+    - beta ~ Normal(0, 50)
+    - sigma ~ HalfNormal(1)
```
drbenvincent, line 267: Same point about LaTeX eqnarray here

causalpy/pymc_models.py
```diff
@@ -106,10 +253,45 @@ def build_model(self, X, y, coords):
...
 
     def build_model(self, X, y, coords):
-        """Defines the PyMC model"""
+        """
+        Defines the PyMC model
+
+        - y ~ Normal(mu, sigma)
+        - mu = X * beta
+        - beta ~ Normal(0, 50)
+        - sigma ~ HalfNormal(1)
```
drbenvincent, line 293: Same point about LaTeX eqnarray here

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -27,6 +36,18 @@ def test_did():
 
 @pytest.mark.integration
 def test_did_banks_simple():
+    """
+    Test simple Differences In Differences Experiment on the 'banks' data set.
+
+    formula="bib ~ 1 + district * post_treatment"
```
drbenvincent, line 42: I think we can get this line formatted as code (i.e. monoface font) using a sphinx role or directive

drbenvincent (changes requested):
Great stuff. I did a quick pass and left a few more change requests. It's really coming along.

causalpy/pymc_models.py
jpreszler: Thanks for the suggestion, switched to using the `:doc:` role which is much cleaner.

causalpy/pymc_models.py
jpreszler: Thanks, with the seed and the `+NUMBER` doctest option the `round` call is no longer needed.

jpreszler:
@drbenvincent Thanks for the helpful comments. I've made the improvements so this is ready for another look when you have time.

causalpy/pymc_models.py
```diff
@@ -85,10 +204,52 @@ def score(self, X, y) -> pd.Series:
...
+
+        .. math::
+
+            sigma \sim HalfNormal(1)
+
+            beta \sim Dirichlet(1,...,1)
+
+            mu = X * beta
+
+            y \sim Normal(mu, sigma)
```
drbenvincent, line 250: Delete. I think we only need the model maths once, and the best place for that is in the main class docstring, we don't need to repeat it for the `build_model` method. Same comment goes for the other classes

causalpy/pymc_models.py
```diff
@@ -23,29 +43,63 @@ def __init__(self, sample_kwargs: Optional[Dict[str, Any]] = None):
...
             pm.set_data({"X": X})
 
     def fit(self, X, y, coords: Optional[Dict[str, Any]] = None) -> None:
-        """Draw samples from posterior, prior predictive, and posterior predictive
-        distributions.
+        """Draw samples fromposterior, prior predictive, and posterior predictive
+        distributions, placing them in the model's idata attribute.
+
+        .. note::
+            Calls the build_model method
```
drbenvincent, line 79: Remove I think. Users don't need to know about the build_model method, that's not called by end-users. We should consider making this a private class, but that can be discussed elsewhere.

drbenvincent (changes requested):
This is really getting there! Not far to go I think. 

Sorry for the iterative nature of this, but having had time to stare at this for a while, I've had some thoughts.

I think the docstring examples are great. But perhaps we've overdone it. I feel that one example for each class makes a lot of sense, but examples for accessing properties (like `result.idata`) and some/all methods (e.g. `result.plot()`) are overkill. My feeling is that these are relatively self explanatory and we can perhaps leave it to the example notebooks for users to see more complete worked examples in its full context.

I'm open to counter-arguments, but if you agree then let's just keep one core example per class.

jpreszler:
I think there's definitely a bit of redundancy in the examples, and with doctest that adds a lot of time to running tests.

I've reduced the redundancy and moved all meaningful examples (like `summary()` calls) to a single example for each class and removed plot and a few other low value examples.

drbenvincent (changes requested):
Great stuff. I just noticed 2 minor issues in `pymc_models.py`:

1. Under the `LinearRegression`, there seems to be something going wrong with the note admonition where it says `Generally, the `.fit()` method should be used rather than calling `.build_model()` directly.` This could just be an issue with my local build of the docs, but if it's an issue for you also, let's see if we can fix it.
2. Under the `InstrumentalVariableRegression` class, the `:code:`priors = {"mus":...` has a line break so it's separate from the `:param priors:` block.

**I'm sure there are a bunch of other small errors that we might find or slight improvements, but I'm happy to merge after these minor updates :)**

jpreszler:
Those issues were not just local to you. I've fixed them and looked for other problems, but didn't spot much besides the same issue in `WeightedSumFitter` as in `LinearRegression`. 

The tests have also passed on my fork after some small adjustments.

drbenvincent:
Quick question as I've never used `doctest` before... 

As far as I can tell doctests are run with `pytest --doctest-modules causalpy/`.  At the moment I can only see these being triggered with the remote tests. Do you think it makes sense to add brief instructions to `CONTRIBUTING.md` either under "Pull request checklist" or "Building the documentation locally" to tell people that they should run the doctests locally (and how to do that) before a PR.

jpreszler:
Good call. The same command can run all doctests locally, but I also added a `make doctest` command to the makefile. I've also added some details to the contributing doc about running all doctests or individual ones. This might be too much for the PR checklist.

This is my first venture into doctests, but the system is pretty straightforward. A good thing to note is that the `+NUMBER` option that is used by a number of the tests is a pytest extension of doctest. If you use doctest directly (not through pytest) these tests will fail.

drbenvincent:
This is great. When running the doctests I just noticed that it produces new files, `ancova_data.csv` and `regression_discontinuity.csv`. We should ideally either not produce those files, or have some clean-up.

jpreszler:
That's from a few of the doctests for the simulated data, I've skipped the lines that write out the csvs, but leaving the example of how to do so.

drbenvincent (approved):
Approving. Thanks for the contribution! The first of many perhaps :)

PR #234, opened 2023-08-27
[docs] Fix broken instrumental variable regression image link

alexmalins:
Minor bug that sneaked through from #213. The second image of the two created in this [commit](https://github.com/pymc-labs/CausalPy/pull/213/commits/dd2201667404335531151fea470c0f9a30b6d115) seemed the best option.


NathanielF (approved):
Thanks, good catch

PR #239, opened 2023-09-15
Issue 155: single source of version, setuptools -> flit

jpreszler:
This should resolve issue #155 , but I couldn't get setuptools to do dynamic versioning because of the imports in `__init__.py`. Switching to flit seems to work with minimal changes.

I tested by changing the version locally and using `pip install -e .` in a clean environment. The version was correct. Also using `python -m build` produced targ.gz and whl files with the correct versions in names.

Not sure if other testing would be good or if there's further discussion needed before changing the backend build system.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
           mkdir -p test-sdist
           cd test-sdist
           python -m venv venv-sdist
-          venv-sdist/bin/python -m pip install ../dist/CausalPy*.tar.gz
+          venv-sdist/bin/python -m pip install ../dist/causalpy*.tar.gz
           echo "Checking import and version number (on release)"
           venv-sdist/bin/python -c "import causalpy; assert causalpy.__version__ == '${{  github.ref_name }}' if '${{ github.ref_type }}' == 'tag' else causalpy.__version__; print(causalpy.__version__)"
           cd ..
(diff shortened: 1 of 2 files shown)


drbenvincent:
Just checking that this is still a draft PR? When it's ready can you get me _and_ @lucianopaz to review?

jpreszler:
@drbenvincent and @lucianopaz , I've cleaned this up and it's ready for review. 

I looked into `bump2version` but it seemed like it would involve more changes because it's an external command line tool, but would work if there's strong desire to keep `setuptools` as the build system.

drbenvincent:
## Automated triage

**Recommendation:** `review:high` — **maintainer decision needed.**

**Why:**
- Proposes changing the build backend from `setuptools` to `flit` for single-source versioning (issue #155).
- This is a project-infrastructure decision. PR #850 (by @anevolbap) already implements the same goal using `setuptools dynamic versioning` without changing the build backend — that approach is more conservative and already CI-verified.
- Only `codecov/patch` has run; no CI checks are present. PR is very old (September 2023) and conflicted (DIRTY).

**Review focus:**
1. **Maintainer decision needed:** Does the project prefer the `setuptools dynamic versioning` approach (PR #850) or the `flit` approach (PR #239)? These are competing solutions for the same issue.
2. If the `setuptools` approach is preferred, this PR should be closed in favour of #850.
3. If the `flit` approach is preferred, the PR needs significant updates and CI re-validation.

**Confidence:** high

Issue #238, opened 2023-09-15
Improve math rendering of models in the docs


drbenvincent:
#232 introduced better docstrings which include mathematical specifications of the models, like this:
<img width="569" alt="Screenshot 2023-09-15 at 21 03 01" src="https://github.com/pymc-labs/CausalPy/assets/6765047/a5230300-07e2-4ec2-aef9-4c4ae244b99c">

In an ideal world we could pimp these maths blocks, e.g.
* get the equations aligned on the $\sim$ or $=$
* use greek symbols such as $\beta, \sigma, \mu$ etc
* get operators to be formatted properly, e.g. `\mathrm{Normal}`

We could optionally add in the model math for the `WeightedProportion` model:
https://github.com/pymc-labs/CausalPy/blob/234a0cd6a94f3e6effd28c11c63a41f8dcf89bdc/causalpy/skl_models.py#L15

PR #240, opened 2023-09-16
improve math rendering of pymc model docs

jpreszler:
This resolves issue #238 
Screenshot 2023-09-16 at 2 25 54 PM

### DIFF
--- a/causalpy/pymc_models.py
+++ b/causalpy/pymc_models.py
@@ -152,13 +152,13 @@ class WeightedSumFitter(ModelBuilder):
 
     .. math::
 
-        sigma \sim HalfNormal(1)
+        \sigma &\sim \mathrm{HalfNormal}(1)
 
-        beta \sim Dirichlet(1,...,1)
+        \\beta &\sim \mathrm{Dirichlet}(1,...,1)
 
-        mu = X * beta
+        \mu &= X * \\beta
 
-        y \sim Normal(mu, sigma)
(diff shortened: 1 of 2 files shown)


drbenvincent:
Not sure why the doctests are failing - they pass when I run then locally. Sometimes remote tests randomly fail and then pass when they are re-run, but that hasn't worked here (yet).

jpreszler:
@drbenvincent The failure is coming from numpy when conftest is getting initialized, looks like it's the same issue on your website change pr. A new version was just released and the may have changed something. I've pinned version 1.24.4 (the most recent that github has before 1.26.0, I had 1.25.2 locally) and the tests are starting to run now.

jpreszler:
The issue is actually coming from https://github.com/pymc-devs/pytensor/blob/main/pytensor/link/c/cmodule.py#L2720 in pytensor. Since CausalyPy and PyTensor don't have upper limits on the numpy version, this line throws an exception since v1.26.0 (released on Sept. 16th) no longer has the `get_info()` method.

drbenvincent:
Cool. I see you've dropped an issue in the pytensor repo https://github.com/pymc-devs/pytensor/issues/441

What do you think the best way forward is? Something like this?
* Temporarily pin `numpy<1.26.0` like you've done **and** create an issue to resolve this in the future. 
* That future resolution would also presumably require a version of pytensor which resolves your issue https://github.com/pymc-devs/pytensor/issues/441 ?

Issue #237, opened 2023-09-15
Get logo to show on readthedocs


drbenvincent:
## Get the logo to display in the docs
The `sphinx_rtd_theme` theme is in use for the docs currently. This theme allows for the use of a logo to be displayed in the top left area.

As far as I understand we should be able to get this to work using`docs/source/conf.py`

https://github.com/pymc-labs/CausalPy/blob/234a0cd6a94f3e6effd28c11c63a41f8dcf89bdc/docs/source/conf.py#L89-L98

We have a logo at `docs/source/_static/logo.png` but I can't get it to display. I did try some experiments being more specific at the path to the logo image in `conf.py` but I didn't have any luck when rendering the docs locally.

## Optional bonus

We could also use the logo for `html_favicon`

PR #245, opened 2023-09-19
fix logo path and add favicon option

jpreszler:
This resolves issue #237 . 

I'm using the same image for the favicon, which looks ok according to http://www.colinkeany.com/favicon-checker/, but in general I think the logo isn't the best on dark themed browsers, so it may be worth having a specialized favicon version of the logo that looks good at the small size in all themes.

 
Screenshot 2023-09-19 at 1 00 51 PM

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -91,7 +91,8 @@ myst_enable_extensions = [
 
 html_theme = "sphinx_rtd_theme"
 html_static_path = ["_static"]
-html_logo = "logo.png"
+html_logo = "_static/logo.png"
+html_favicon = "_static/logo.png"
 html_theme_options = {
     "logo_only": True,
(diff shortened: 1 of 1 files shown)


jpreszler:
I took a stab at a specialized favicon that is a bit more distinctive in browser tabs (especially dark themed). I've added a white background, extracted the rocket and dag parts, and done some color and sharpness adjustments to help the contrast.

Screenshot 2023-09-21 at 9 45 46 AM

jpreszler:
@drbenvincent this is ready when you get a chance.

drbenvincent:
Nice! Sorry for the delay - I've had some newborn parenting challenges. I'll merge this and see how things go. We can always iterate on the logo or light/dark theme if needed, but this looks good already.

Issue #161, opened 2023-01-11
double check input validation for PrePostNEGD


drbenvincent:
At the moment we test with `_series_has_2_levels`. But double check if we instead need to enforce `_is_variable_dummy_coded`. If so then:
- Update `PrePostNEGD._input_validation`
- Update corresponding tests in `test_input_validation.py` to check we get the appropriate exception 
- Update the example notebook to clarify the data requirements

PR #247, opened 2023-09-21
fix PrePostNEGD input validation

jpreszler:
This resolves issue #161 

I tried the PrePostNEGD example with `group` being binary (current), boolean, or categorical with levels 'A' and 'B'. All three pass the `_series_has_2_levels` check but in the prediction after model fitting, errors result in the 'A'/'B' case because 'untreated' is given values of 0 (and 'treated' 1). Switching to `_is_variable_dummy_coded` moves the error to a data exception in the input validation, which makes more sense.

A more complicated fix would be to assign the two levels in lines 1059 and 1070 rather than 0's and 1's, but then we would need to know which is the 'treatment' level so probably best to leave those transformations to users.

I also fixed a bug that requires the group variable to be called 'group' and adjusted some re-used variable names whose names didn't match what was being stored.


jpreszler:
@drbenvincent also ready when you get a chance

PR #254, opened 2023-09-26
split pymc_experiments into smaller files

jpreszler:
No functional code changes, just splitting into smaller files.

This relates to discussion #253 .


drbenvincent:
So this was a good idea, but has been superseded by #381

PR #260, opened 2023-10-17
update README

drbenvincent:
* Remove the _comparison to related packages_ section. The comparison table was pretty limited in terms of only comparing to a few other packages. There are a very large number of causal inference packages, so this table wasn't cutting it. We can bring back a section on related packages but take a different approach.
* Add section on Instrumental Variable Regression. When this feature was added by @NathanielF, it was added to the readthedocs page but not the GitHub readme. We don't have an idiomatic IVR plot output from that notebook, so I've just used a DAG already used in the example notebook.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,20 +17,6 @@
 
 A Python package focussing on causal inference in quasi-experimental settings. The package allows for sophisticated Bayesian model fitting methods to be used in addition to traditional OLS.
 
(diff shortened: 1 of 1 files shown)


twiecki:
I liked the comparison table. It's never perfect but I think it does provide a good overview of the added features here compared to the other packages that are most similar and well-known.

Issue #261, opened 2023-10-17
Improve the CausalPy PyPI page


drbenvincent:
The [PyPI](https://pypi.org/project/CausalPy/) page currently displays the github `README.md` however none of the images display.

Either:
1. Fix it so that the images are displayed on the PyPI page

or 

2. Change the PyPI page to be a more minimal text-only description of the package.

PR #270, opened 2023-11-11
add absolute paths to images in README to fix PiPy description

krz:
I replaced the image links in the `README.md` with their absolute paths. This should fix issue #261 .
The `codecov` link is unchanged and will still not be displayed properly on PyPi (probably because of the required token id).


twiecki (approved):

twiecki:
Thanks @krz!

Issue #273, opened 2023-11-23
Splitting dataframe into pre-post is inconsistent with covid example


rlaker:
In the COVID excess deaths [example](https://causalpy.readthedocs.io/en/latest/notebooks/its_covid.html), the data is split into pre and post treatment with 

```
self.datapre = data[data.index <= self.treatment_time]
self.datapost = data[data.index > self.treatment_time]
```

However, when inspecting result.datapre we see that 2020-01-01 is included with the label `pre=False`

![image](https://github.com/pymc-labs/CausalPy/assets/20686171/efe8f61f-4e36-4c9c-9513-e0e95a80d0bb)

If 2020-01-01 should be in the post set, as the df says, then the splitting should be changed to:
```
self.datapre = data[data.index < self.treatment_time]
self.datapost = data[data.index >= self.treatment_time]
```

If it should be in the pre set, then `deaths_and_temps_england_wales.csv` needs to be updated so that `pre=True` for this date

drbenvincent:
Thanks for this @rlaker. 

I agree with the second code snippet - `date < treatment_time` should be classed as pre, and `date >= treatment_time` should be classed as post.

It could even be worth adding some input validation and add include that under test coverage. Did you want to have a go at this and submit a PR?

drbenvincent:
Though it could be worth thinking if we want treatment _always_ defined with `>=`, or if there will be some occasions where a user might want `>` instead.

rlaker:
Happy to have a go!

rlaker:
Perhaps the confusion is that the example dataframe defined its own `pre` column, which was then ignored by the `PrePostFit` class which used `treatment_time` instead. Maybe the user should create the `pre` column and the class should do 

```
self.datapre = data[data['pre'] == True]
self.datapost = data[data['pre'] == False]
```

?

PR #275, opened 2023-11-24
changed splitting behaviour for PrePostFit

rlaker:
Fix for #273

### DIFF
--- a/causalpy/pymc_experiments.py
+++ b/causalpy/pymc_experiments.py
@@ -164,8 +164,8 @@ class PrePostFit(ExperimentalDesign):
         # set experiment type - usually done in subclasses
         self.expt_type = "Pre-Post Fit"
         # split data in to pre and post intervention
-        self.datapre = data[data.index <= self.treatment_time]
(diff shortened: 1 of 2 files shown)


drbenvincent:
Thanks very much for this @rlaker! Just letting you know that this is on my list. Things are hectic but I will be able to request changes or approve/merge in the first few days of Dec. So around the end of this week.

drbenvincent:
Sorry for the delay on this @rlaker. I think this is good to merge. We just have a failing test. Tests pass for me locally, so it could perhaps be a version issue, or sometimes remote tests randomly just fail. I'll try to trigger those tests again and see what happens.

drbenvincent:
Remote failing tests should hopefully be fixed whenever https://github.com/pymc-devs/pytensor/pull/550 is fixed and make its way into a new pytensor release. Or when https://github.com/pymc-labs/CausalPy/issues/279 is resolved. So don't sink any effort into trying to fix failing remote tests right now. You can of course ensure that all local tests pass - they are not affected by the issue. Not on my machine anyway.

Issue #278, opened 2023-12-12
check what is run with `make doctest` and `make test`


drbenvincent:
* Running `make doctest` locally took over 5 mins and there were 84 passing tests. 
* Running `make test` locally took just over 2 mins with 62 passing tests.

Something about `make doctest` doesn't sound right. I don't think we have anywhere close to 84 docstrings that need doctest. Maybe it's additionally running the regular tests on top of the doctests? Either way, it would be great to get to the bottom of this to save time and local and remote compute time.

Tagging @jpreszler just in case you're interested in taking a look into this.

jpreszler:
Here's a PR for an improvement. https://github.com/pymc-labs/CausalPy/pull/280

The tests were getting run twice, but doctests still take a while. This is mainly because some of the tests require doing enough MCMC sampling for numeric stability in the examples. There's likely ways to save time here also, but it's not a simple as just reducing the pytest test collection.

drbenvincent:
Closed by #280

Issue #279, opened 2023-12-13
install pymc in the recommended manner for remote workflows


drbenvincent:
Following a discussion with @ricardoV94, it seems that we are pip installing pymc. We should instead install pymc in the [recommended manner](https://www.pymc.io/projects/docs/en/stable/installation.html). 

I'm not 100% sure how to implement this currently, but I'm assuming we need to edit files in `.github/workflows`. Perhaps...

https://github.com/pymc-labs/CausalPy/blob/83cb28cfd728ce01ed175c1ddd300b91308bf410/.github/workflows/ci.yml#L43

which may well require changes to the `pyproject.toml` file

https://github.com/pymc-labs/CausalPy/blob/83cb28cfd728ce01ed175c1ddd300b91308bf410/pyproject.toml#L29-L43

drbenvincent:
Is this something that @maresb might have expertise with?

anevolbap:
Heads up: opened #910 to switch the CI test job to micromamba + the autogenerated environment.yml. Both 3.12 and 3.14 legs pass.

PR #280, opened 2023-12-14
Ignore `causalpy/tests/` when running doctest to save a little time

jpreszler:
Ignore `causalpy/tests/` when running doctest to save a little time. Resolves issue #278.

This cuts out the tests from being run twice, and doesn't doctest the docstrings in the tests - but none of them have examples so the doctests weren't doing much.

Locally, this went from 4m 50s to 3m 20s, so ~40% reduction in time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Run doctests
         run: |
           pip install -e .[test]
-          pytest --doctest-modules causalpy/
+          pytest --doctest-modules --ignore=causalpy/tests/ causalpy/
       - name: Run tests
         run: |
           pip install -e .[test]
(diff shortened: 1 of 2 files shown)


drbenvincent:
Thanks @jpreszler, this looks great. The tests and doctests work for me locally. @ricardoV94 do you think we can get away with merging this despite the failing remote test, which I believe is being caused by the issue (https://github.com/pymc-labs/CausalPy/issues/279) we discussed the other day?

drbenvincent:
Remote tests seem to pass now that https://github.com/pymc-devs/pytensor/pull/550 is addressed in [pytensor 2.18.4](https://github.com/pymc-devs/pytensor/releases/tag/rel-2.18.4)

PR #281, opened 2023-12-17
Use conda-lock to create test environments

### LINKED ISSUE #279: install pymc in the recommended manner for remote workflows

maresb:
Closes #279, if I'm understanding correctly.

This needs to be tested, but it uses lockfiles (which are generated by a separate CI) to set up a Conda environment for running the tests.

We need to set up a [fine-grained GitHub token](https://github.com/settings/tokens?type=beta) with the following permissions to the CausalPy repo:
* Read access to metadata
* Read and Write access to code and pull requests
and make it available as a secret named `GH_PAT_FOR_PR`. (I can't do it myself since I'm not admin here.)


maresb:
It looks like the Python 3.8 doctests are hanging right after:

```
 causalpy/tests/test_integration_pymc_examples.py::test_geolift1 PASSED   [ 69%]
```

I'm not sure what's the next test that's running here.

drbenvincent:
Cool. This is outside my comfort zone at the moment, so let's jump on a quick call at some point when you have time.

maresb:
I'm happy to discuss when I find a moment, but let me write a few words about how this works, because I think ideally these sorts of things should be documented somewhere.

The starting point is a bunch of test environment definitions. They include a particular version of Python, and pip. For completeness, I define environments for Python 3.11 and 3.12, even though they aren't used.

The next step is generating/updating the lockfiles via the `update-lockfile.yml` workflow. For each of the Python versions in `["3.8", "3.9", "3.10"]`, we generate lockfiles based on the dependencies in `pyproject.toml`, the `test` extras defined in `pyproject.toml`, and the environment definition file which selects the particular Python version. In more detail, conda-lock merges these dependency specifications, solves them, and writes the solution into the respective lockfile. The workflow generates a pull request once per week with the updated dependency solutions.

The final step is updating the test CI to install and cache the environment based on the lockfile.

This guarantees that the installed dependencies are deterministic, reproducible, regularly updated, and done in the recommended Conda way.

drbenvincent:
We're currently running tests under 3.10, 3.11, 3.12, so I think we can remove 3.8 and 3.9 related stuff. Happy to chat if it's time to revisit this.

maresb:
Ya, let's discuss next week and move this forward.

maresb:
Will probably use pixi instead

PR #283, opened 2023-12-22
User specified number of significant figures for numerical outputs

### LINKED ISSUE #282: User specified number of significant figures for numerical result outputs
* Similar to #272
* Allow user to provide `round_to` kwarg to the `summary` method. That will in turn involve changes to the `print_coefficients` method

drbenvincent:
* Closes #282 
* Doctests and tests pass locally.
* Docstrings updated to reflect the new output you get.

### DIFF
--- a/causalpy/pymc_experiments.py
+++ b/causalpy/pymc_experiments.py
@@ -64,7 +64,7 @@ class ExperimentalDesign:
         Prints the model coefficients
 
         :param round_to:
-            Number of decimals used to round results. Defaults to 2. Use "none" to return raw numbers.
+            Number of decimals used to round results. Defaults to 2. Use "None" to return raw numbers.
 
         Example
         --------
@@ -234,7 +234,7 @@ class PrePostFit(ExperimentalDesign):
         Plot the results
 
         :param round_to:
-            Number of decimals used to round results. Defaults to 2. Use "none" to return raw numbers.
+            Number of decimals used to round results. Defaults to 2. Use "None" to return raw numbers.
         """
         fig, ax = plt.subplots(3, 1, sharex=True, figsize=(7, 8))
 
@@ -339,7 +339,7 @@ class PrePostFit(ExperimentalDesign):
         Print text output summarising the results
 
         :param round_to:
-            Number of decimals used to round results. Defaults to 2. Use "none" to return raw numbers.
+            Number of decimals used to round results. Defaults to 2. Use "None" to return raw numbers.
         """
 
         print(f"{self.expt_type:=^80}")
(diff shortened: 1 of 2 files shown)


drbenvincent:
Pfft. These doctests are brittle.

jpreszler:
There are definitely some that are particularly problematic, and will continue to be so without some significant changes. I think the value comes over the long-term, but it will be hard to get there without some improvements. The options I see:
1. turn them off, at least temporarily.
2. Split the example and output into separate docstrings (could be difficult for some longer examples). This would flag the biggest problems (an exception from input changes) but would make the actual example output static. This could be a good temporary fix in some places.
3. Replace the more problematic examples. I generally grabbed things from the other docs which are nice realistic example but have coefficients near 0 or HDI's near or including 0. This is perhaps the biggest issue since the output may change sign in a way the makes even the pytest numeric accuracy checks too strict. It's likely possible to have examples that show the right output, but sacrifice realism for basic numeric stability that make the doctests far more reliable. 

My vote would be to temporarily turn off the tests and replace the examples for at least PrePostFit (flakey), RegressionDiscontinuity (flakey) and the Instrumented Variables (slow) and turn doctests back on. I should be able to work on this soon, perhaps even getting better examples over the next week.

jpreszler:
created https://github.com/pymc-labs/CausalPy/issues/284. For now, I think you can just comment out the doctest call in the CI untill that issue is resolved.

drbenvincent:
Note to self: I'll finish and merge this PR after #286 is merged

drbenvincent:
Well, I guess I have to increase test coverage

causalpy/pymc_experiments.py
```diff
@@ -59,10 +59,13 @@ def idata(self):
 
         return self.model.idata
 
-    def print_coefficients(self) -> None:
+    def print_coefficients(self, round_to=None) -> None:
         """
         Prints the model coefficients
 
+        :param round_to:
+            Number of decimals used to round results. Defaults to 2. Use "none" to return raw numbers.
```
juanitorduz, line 67: None (upper case)

causalpy/pymc_experiments.py
```diff
@@ -1362,20 +1376,23 @@ def _causal_impact_summary_stat(self, round_to) -> str:
...
-        causal_impact = f"{self.causal_impact.mean():.2f}, "
+        causal_impact = f"{round_num(self.causal_impact.mean(), round_to)}, "
         return f"Causal impact = {causal_impact + ci}"
 
     def summary(self, round_to=None) -> None:
         """
         Print text output summarising the results
+
+        :param round_to:
+            Number of decimals used to round results. Defaults to 2. Use "none" to return raw numbers.
```
juanitorduz, line 1387: None (upper case)

juanitorduz (changes requested):
Minor change request ;) Otherwise LGTM.

causalpy/pymc_experiments.py
drbenvincent: Done: changed this for all instances

PR #286, opened 2024-01-05
Make results fully reproducible 

### LINKED ISSUE #284: Make docstring examples more stable
Several of the docstring examples have numeric output that changes too much leading to doctest failures.

We should change the data and models in those examples (particularly PrePostFit and RegressionDiscontinuity) to have more stability even if the examples become unrealistic or 'silly' - they will illustrate basic usage and ensure documentation is up to date with code, leaving the main docs for real examples and detailed instruction.

The main goal is that doctests should always pass unless there's been a clear change in how a function works.

drbenvincent:
Problems with doctests passing revealed that we were only passing the `random_seed` kwarg to the `pm.sample` method and that this was not being passed into `pm.sample_prior_predictive` or `pm.sample_posterior_predictive`. 

This will hopefully avoid any future issues in doctest fragility. But it is also a pretty useful (if not essential) fix to make.

Closes #284.

### DIFF
--- a/causalpy/tests/test_pymc_models.py
+++ b/causalpy/tests/test_pymc_models.py
@@ -125,7 +125,11 @@ def test_idata_property():
     assert isinstance(result.idata, az.InferenceData)
 
 
-def test_result_reproducibility():
+seeds = [1234, 42, 123456789]
+
+
+@pytest.mark.parametrize("seed", seeds)
+def test_result_reproducibility(seed):
     """Test that we can reproduce the results from the model. We could in theory test
     this with all the model and experiment types, but what is being targetted is
     the ModelBuilder.fit method, so we should be safe testing with just one model. Here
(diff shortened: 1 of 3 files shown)


juanitorduz (approved):

causalpy/tests/test_pymc_models.py
```diff
@@ -123,3 +123,38 @@ def test_idata_property():
...
+
+def test_result_reproducibility():
+    """Test that we can reproduce the results from the model. We could in theory test
+    this with all the model and experiment types, but what is being targetted is
+    the ModelBuilder.fit method, so we should be safe testing with just one model. Here
+    we use the DifferenceInDifferences experiment class."""
+    # Load the data
+    df = cp.load_data("did")
+    # Set a random seed
+    sample_kwargs["random_seed"] = 42
```
juanitorduz, line 136: Maybe you can parametrize this test, passing maybe the other two seeds?

drbenvincent: Great idea. Have pushed changes to do this.

PR #290, opened 2024-02-05
Expand data validation to also cover the scikit-learn experiment classes

### LINKED ISSUE #78: Add input validation to experiment objects
Adding some input validation to the experiment objects is a sensible idea. Taking a more defensive programming stance is probably sensible now that we've announced the beta release. So this could head off a lot of simple dataset or user input issues.

- [ ] DiD / scikit-learn model
- [x] DiD / pymc model (done in https://github.com/pymc-labs/CausalPy/commit/e33ce25f18e6916d58ac32c8c8054c43e9395f44)
- [ ] SC / scikit-learn model
- [x] SC / pymc model
- [ ] RD / scikit-learn model
- [x] RD / pymc model
- [ ] ITS / scikit-learn model
- [x] ITS / pymc model
- [x] Instrumental variables / pymc model
- [ ] Instrumental variables / skit-learn (not implemented, see #229)
- [x] Regression Kink design / pymc model
- [ ] Probably a good idea to raise exceptions (rather than just add in asserts) to help with testing.

drbenvincent:
* Closes #78 
* We currently have `_input_validation` methods in just the pymc experiment classes
* This PR extracts these out into new Mixin classes in `data_validation.py`
* Now we can use the same data validation code for both the pymc and the scikit-learn versions of the experiment classes.


jpreszler (approved):
Sorry it took me so long to get to this. I've been starting work at 6am to overlap with coworkers in the UK and have been wiped out in the evenings. 

I like splitting out the input validation, it's good progress to making things more streamlined.

There are a few suggestions I would make:
1.  codecov has flagged a couple of things that don't have tests that would be easy to add but they seem fairly low value.
2. There's a lot of redundancy here. For example, there are 4 checks for `_is_variable_dummy_coded` with the only difference being the argument. Perhaps there could be a generic `DataValidator` class with methods for common checks that each experiment validator inherits from and the `_input_validation` methods just run the checks needed for that experiment.

I don't think either of these should hold up merging this PR. If you like the idea of 2, I could do that in the near future.

drbenvincent:
Thanks @jpreszler. I've added tests for the scikit-learn versions of the experiments in `test_input_validation.py`

PR #291, opened 2024-02-12
Fix #288: remove deprecated 'model' property in tests.

anevolbap:
Closes issue #288.

Removes the `FutureWarning` in tests caused by a deprecated property:
```
FutureWarning: Model.model property is deprecated. Just use Model.
    warnings.warn("Model.model property is deprecated. Just use Model.", FutureWarning)
```

PR #293, opened 2024-02-21
Replacing self.model with self

peeeffchang:


drbenvincent:
Closing as this is fixed by #291. But thanks nevertheless for the PR @peeeffchang

PR #296, opened 2024-03-01
Fix causal impact calculation error in synthetic control and interrupted time series

### LINKED ISSUE #294: Ensure causal impact is calculated from posterior predicted observations, not expected values
Stemming from a comment on the PyMC discourse https://discourse.pymc.io/t/causalpy-evaluating-uncertainty-for-interrupted-time-series/13491/8 it looks like there might be a mistake in some of the causal impact calculations.

For example, in the `PrePostFit` class we have this

https://github.com/pymc-labs/CausalPy/blob/6283c7625b724956378c7a7586c0b09430567e25/causalpy/pymc_experiments.py#L199-L209

So the problem here is that we are calculating the causal impact on the expected value of the outcome rather than the posterior predicted observation. This will likely lead to causal impacts being underestimated.

We should check to see if this is done anywhere else.

drbenvincent:
* There was an error in the causal impact calculations which affected Interrupted Time Series and Synthetic control. The causal impact was being calculated using the expectation of the posterior predictive distribution. But it should have been based on the actual posterior predictive values which have additional likelihood-related observation noise. Therefore, previous estimates of causal impact would have slightly underestimated the uncertainty.
* I also re-ran the affected notebooks
* Closes #294


drbenvincent:
I'll cut a point release when this is merged.

juanitorduz (approved):
Good catch!

PR #304, opened 2024-03-05
Move to a conda only install of pymc

### LINKED ISSUE #287: Remove the `pip install` lines in `Makefile` and add the `--no-deps` flag for the local install
@lucianopaz pointed out that having `pip install` commands in the `Makefile` could potentially lead to version conflicts that can't be resolved. So we should perhaps remove these lines in the `Makefile`.

Add the `--no-deps` flag when doing the local import of the CausalPy package. The purpose of this is to ensure that we have a clean condo-only install of pymc. 

Double check the `CONTRIBUTING.md` file to ensure instructions are up to date.

### LINKED ISSUE #297: Move to a conda-only install of pymc
* Related to #279
* The recommended install instructions for PyMC is via conda, not pip. We are already doing that, but it seems that at a later step in the setup of the environment that we are basically overwriting that with a pip installed pymc.
* Based on experience with other repo's, it seems that we'll have to:
  * Remove the package dependencies from `pyproject.toml` and instead place them in `environment.yml`
  * Build the environment from this new `environment.yml`
  * Update the instructions in `CONTRIBUTING.md`
* We shou also re-run the notebooks which show `WARNING (pytensor.tensor.blas): Using NumPy C-API based implementation for BLAS functions.` because this is indicative of pymc being pip installed rather than conda installed. 

- [x] Re-running of diff in diff notebooks is blocked by #299. Fix this first.

drbenvincent:
* Closes #297 
* ~Closes #287~ (Addressed by #309)
* Core dependencies are ~~taken out of `pyproject.toml` and moved into `environment.yml`~~ duplicated in `environment.yml`
* Instructions for setting up dev environment are updated in `CONTRIBUTING.md`
* I re-ran the notebooks which show `WARNING (pytensor.tensor.blas): Using NumPy C-API based implementation for BLAS functions`. which is indicative of pymc being pip installed rather than conda installed.


drbenvincent:
Looks like the ci workflow is failing. I'm not strong on GitHub actions (see `ci.yml` I believe), so would appreciate any pointers. But it looks like the changes have broken the remote environment setup, which I suspected would happen :)

pyproject.toml
```diff
@@ -26,21 +26,7 @@ requires-python = ">=3.8"
...
-    "pandas",
-    "patsy",
-    "pymc>=5.0.0",
-    "scikit-learn>=1",
-    "scipy",
-    "seaborn>=0.11.2",
-    "statsmodels",
-    "xarray>=v2022.11.0",
-]
+dependencies = []
```
juanitorduz, line 29: I think the problem comes because you are deleting these dependencies and you are pip installing the package in the CI

juanitorduz: I think you can try to mimic the way is done in https://github.com/pymc-devs/pymc-experimental/blob/main/.github/workflows/test.yml

drbenvincent: I'm not that experienced with GitHub workflows. What am I looking for? Are you referring to the `environment-file:` in the `with:` block?

twiecki (approved):

pyproject.toml
maresb: Hey, I'm on vacation now, but we should discuss (and update) #281.

As for Conda vs pyproject dependencies, ideally we keep both. Pyproject dependencies define the python-only dependencies that you have. On the other hand, Conda takes care of both Python and non-Python dependencies.

I haven't read carefully any of the details, so I may be missing stuff, but the way I like to handle this is to:
1. Create a Conda environment that satisfies all the dependencies, Python or otherwise
2. `pip install` the project into the Conda environment with the `--no-dependencies` flag so that `pip` installs only the project
3. Run `pip check` to ensure all of the project's Python dependencies are satisfied within the Conda environment. This will trip on the case when you declare a Python dependency in `pyproject.toml` that is missing from your `environment.yaml` file. I don't really know of any tests that go in the other direction, and I don't know that it would even make sense.

Note that this means that your primary source of truth for a project's dependencies should be the `pyproject.toml` file, not the `environment.yaml` file! I think this is how it should be. It enables masochistic "experts" to do an unsupported `pip install` if they are so inclined.

Regarding #281, `conda-lock` can [parse](https://github.com/pymc-labs/CausalPy/blob/b92084fc2f5b1f924a899ad5ec00269f89ef687e/.github/workflows/update-lockfile.yml#L34) the `pyproject.toml` for the requirements and generate a lockfile, so apart from specifying special requirements, we may be able to replace the `environment.yml` with a lockfile.

drbenvincent:
Just updated requirement to `pymc>=5.14.0`, in order to keep up with #325

drbenvincent:
When building the environment (locally) using the methods in this PR (i.e. conda):
* all tests pass
* 12 doctests fail

That suggests that it's not a pip/conda environment construction issue

We also get 12 failing doctests in the remove tests.

drbenvincent:
Would be good to get a sanity check to ensure I won't break everything before we merge this :)

maresb:
At a glance this looks sound, but this seems to be completely user-facing, meaning it doesn't seem like anything is being validated/tested in the CI.

maresb:
(At worst you're breaking something minor in the development workflow.)

maresb (approved):

maresb:
On the CI front, it would be nice to move forward on #281.

drbenvincent:
Yes, this one is user-facing. Will try to progress the ci PR very soon.

PR #306, opened 2024-03-05
Fix #259: Annotation in `did_skl` notebook plot .

### LINKED ISSUE #259: Error in sci-kit learn version of `DifferenceInDifferences.plot` method
When running the `did_skl.ipynb` example we get an error when calling the plot function.

    ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 2 dimensions. The detected shape was (2, 1) + inhomogeneous part.

I believe this relates to line 453 here where we calculate the position of the text annotation associated with the causal effect arrow.

https://github.com/pymc-labs/CausalPy/blob/f1a522d14b223eed9ec397069f59c4a1b35b01ee/causalpy/skl_experiments.py#L451-L459

anevolbap:
Closes #259.

- `self.y_pred_counterfactual` was a list within a list.

### DIFF
--- a/causalpy/skl_experiments.py
+++ b/causalpy/skl_experiments.py
@@ -475,7 +475,7 @@ class DifferenceInDifferences(ExperimentalDesign, DiDDataValidator):
         )
         ax.annotate(
             "causal\nimpact",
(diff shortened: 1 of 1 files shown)


drbenvincent:
Thanks for this @anevolbap 👍🏻

PR #309, opened 2024-03-16
Fix #287 Update Makefile

### LINKED ISSUE #287: Remove the `pip install` lines in `Makefile` and add the `--no-deps` flag for the local install

anevolbap:
Closes #287.

To test it, I created a new virtual environment (with `pyenv`) from scratch and ran the installation commands and the `Makefile` commands as well.

This PR seems to overlap a bit with [this one](https://github.com/pymc-labs/CausalPy/pull/304).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -67,7 +67,7 @@ For more instructions see the [Pull request checklist](#pull-request-checklist)
 
 	Install development dependencies
 
-	```
+	```bash
 	pip install 'causalpy[dev]'
 	pip install 'causalpy[docs]'
 	pip install 'causalpy[test]'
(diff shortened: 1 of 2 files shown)


drbenvincent (approved):
Great. This addresses #287. Thanks for pointing out that this overlaps with #304 - that PR got a bit unwieldy because it simultaneously tried to make the changes to the makefile and move to a conda only install. I think I'll merge this PR now because it focusses #287 solely.

PR #317, opened 2024-04-06
Enable manual triggering of the PyPI release workflow

drbenvincent:
When I released 0.2.3, the `release.yml` workflow (which uploads a new version to PyPI) failed. The cause of this was fixed in #316. I want to manually trigger this action so that 0.2.3 gets uploaded to PyPI without having to do another version bump. 

It should be possible to manually trigger this workflow through the GitHub webpage https://github.com/pymc-labs/CausalPy/actions/workflows/release.yml but currently there is no "Run Workflow" button. I followed the instructions here https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow?tool=webui and have basically added `workflow_dispatch` as another trigger for this action. If I've done this correctly then this should enable manual triggering of this action through the GitHub webpage.

Though I'm not sure this alone will work, because there are additional requirements for some of the steps, e.g. `if: github.event_name == 'release' && github.event.action == 'published'` for the `publish` job.


drbenvincent:
If this is tricky, it could be surplus to requirements. I am always able to manually upload to PyPI with my API tokens.

lucianopaz:
Sorry that I missed the request for reviewing some days ago. If what you want is to be able to push a release to pypi if the release workflow failed, then you don’t need this.
Pypi releases are different from GitHub releases. Pypi releases cannot be amended. If you make a mistake, you can remove a release number from pypi, but you’ll never be able to use that release number again. That’s why pypi forces you to have to push a new release number always.
GitHub releases are completely different. You can amend their release notes, you can delete them and then push them again using the exact same number as before. The only limitation is that if your upload to pypi succeeded, then you shouldn’t try to change the contents of the GitHub release (you still can do it, but the changes won’t be pushed up to pypi).
So you can remove the 0.2.3 release and tag from GitHub, and then start the 0.2.3 release again using the fixed commit. The workflow should run successfully and finally push to pypi.

drbenvincent:
Thanks, and sorry for the delayed reply @lucianopaz. From what I recall, the PyPI workflow failed so the `0.2.3` never really got initiated on PyPI. So I just manually ran through the PyPI release workflow and now there is a CausalPy 0.2.3 on PyPI.

So if I understand correctly we can just close this PR?

anevolbap:
Suggest closing this one: `workflow_dispatch` landed on main in d77653a, and after the merge from main the diff here is empty (0 files changed).

The original goal does not work yet though: the `test` and `publish` jobs are gated on `if: github.event_name == 'release' && github.event.action == 'published'`, so a manual run only executes `build` and never uploads. If you think it is still worth fixing, I can open an issue and a small PR. The fix would be to extend those `if:` conditions to also accept `workflow_dispatch`, and document that manual runs must be dispatched from the release tag so the version assertions pass. The `environment: release` approval gate would stay as is.

Issue #327, opened 2024-05-05
Second iteration on DAGS for quasi-experiments knowledge base page


drbenvincent:
The first iteration of the knowledge base page is currently a PR, #321. But we already have plans to improve on it further.

## Confounding & RCT's
Provide a concrete and numerical example of confounding, and then follow that up with the RCT solution using the do operator.

## IV's
@NathanielF suggests:
> Optional: But you might want to mention that IV and RCT allow unconfounded inference for slightly different estimands. You can refer to the paper for details, but you might also want to distinguish ATE, ATT, ATC, and LATE kinds of estimands and how they relate to different DAGs?

@juanitorduz follows up with:
> And maybe comment about the need of a 'strong' instrument. Because a weak relationship IV -> Z can lead to high variance (right?)
Also, add an example. My favorite on is https://matheusfacure.github.io/python-causality-handbook/09-Non-Compliance-and-LATE.html Maybee we can borrow it (with the proper citation) 

## Regression discontinuity
We could potentially add discussion of ATE and LATE and ATC etc if we can do that while keeping the section clear and concise. Maybe with [drop down admonition boxes](https://mystmd.org/guide/admonitions#admonition-dropdown) or something.

## Synthetic control
The first iteration of the knowledge base page will not have a DAG for synthetic control. It would be nice to add this if possible.

PR #332, opened 2024-05-06
Bound MyST NB to fix the compressed images.

juanitorduz:
See https://github.com/pymc-labs/CausalPy/issues/327#issuecomment-2096256657

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,7 +54,7 @@ docs = [
     "ipykernel",
     "daft",
     "linkify-it-py",
-    "myst-nb",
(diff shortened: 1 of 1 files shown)


drbenvincent (commented):
Thanks for this :) 

Just wondering if there's a more future proof solution than pinning `myst-nb<=1.0.0`? Currently it's on `1.1.0` and we might want to use new versions at some point.

drbenvincent:
Happy to pull the trigger now, because readthedocs is currently set up to render the latest on `main`. So the page is already available, with the poor images.

Issue #336, opened 2024-05-07
Fix failing test `test_inverse_prop`


drbenvincent:
Slightly strange because all tests passed for #311. But we have a failing remote test for https://github.com/pymc-labs/CausalPy/actions/runs/8974649334/job/24647471387

On `main` I re-built the environment from scratch following instructions in `CONTRIBUTING.md` and can reproduce this bug locally:

```
FAILED causalpy/tests/test_pymc_experiments.py::test_inverse_prop - ValueError: autodetected range of [nan, nan] is not finite
```

This was fixed by adding in the `rand_seed` code to the `PropensityScore.fit` method which overrides the `ModelBuilder.fit` method. So the method should now be:

```python
def fit(self, X, t, coords):
    """Draw samples from posterior, prior predictive, and posterior predictive
    distributions. We overwrite the base method because the base method assumes
    a variable y and we use t to indicate the treatment variable here.
    """

    # Ensure random_seed is used in sample_prior_predictive() and
    # sample_posterior_predictive() if provided in sample_kwargs.
    random_seed = self.sample_kwargs.get("random_seed", None)

    self.build_model(X, t, coords)
    with self:
        self.idata = pm.sample(**self.sample_kwargs)
        self.idata.extend(pm.sample_prior_predictive(random_seed=random_seed))
        self.idata.extend(
            pm.sample_posterior_predictive(
                self.idata, progressbar=False, random_seed=random_seed
            )
        )
    return self.idata
```

Though it's worth double checking the details of the test and error message which I didn't do 🤷🏻‍♂️😜

drbenvincent:
closed by #337

PR #337, opened 2024-05-07
Add Propensity Score DAG to Knowledge Base

NathanielF:
Adding random seed to fix bug https://github.com/pymc-labs/CausalPy/issues/336

and adding DAG write up re propensity scores as per https://github.com/pymc-labs/CausalPy/issues/327


NathanielF:
This should be good now @drbenvincent!

causalpy/tests/test_pymc_experiments.py
```diff
@@ -63,7 +63,7 @@ def test_regression_kink_gradient_change():
 
 def test_inverse_prop():
     df = cp.load_data("nhefs")
-    sample_kwargs = {"tune": 100, "draws": 100, "chains": 2, "cores": 2}
+    sample_kwargs = {"tune": 100, "draws": 500, "chains": 2, "cores": 2}
```
drbenvincent, line 66: Could add a `random_seed` kwarg to `sample_kwargs` in order to make test output deterministic. 

So far this is not generally done in tests because we don't test for specific results, but it might be relevant here if you're getting numerical stability issues.

We do do it however in the docstring examples (under test by `doctest`) because there we do examine numerical results.

I'll leave it up to you, but this might be the more defensive play.

drbenvincent (changes requested):
## Comments about the DAG notebook
* Please add the `remove-input` cell tag to the new cell which builds the propensity DAGs in `quasi_dags.ipynb`
* Could be worth briefly explaining $Y(0), Y(1)$ because so far we've got mentioned it and have nothing from the potential outcomes framework. Can be very brief - we'll likely end up with a specific page on the potential outcomes framework
* Nit: Use of PS as an acronym without defining it

causalpy/tests/test_pymc_experiments.py
NathanielF: Updated. Added Random seed

PR #338, opened 2024-05-07
Add example analysis of multiple geo lift test analysis

### LINKED ISSUE #165: Allowing inclusion of multiple 'sites' as treatment group for synthetic control
I'm not sure if this is already possible and I haven't worked it out yet, but it would be great to have a way to easily include multiple 'sites' to make up the treatment group. In the example of geolift experiments, sometimes you would want to include multiple cities as the treatment group. 

This is something easily done in Meta's GeoLift, so it would be great to see it here as well. 
And if it is already possible....it would be great to have instructions included in the documentation

Thanks

### LINKED ISSUE #320: Analysis of data with multiple treatment regions
We already have the ability to run geo-testing (with synthetic control methods), but this assumes that we have just a single treated region.

But often we may have (or plan to have) multiple regions which receive treatment. This treatment may either be the same, or different, if it is different, it could be different in kind (e.g. store refurbishment vs price discounts) or degree (level of price discount).

Below are notes which sketch out some approaches we could take here:

## 1. Individual synthetic control analyses for each treated geography 
It would be perfectly valid to treat each treated region as its own experiment. For each treated region, the set of control regions would be the complete set of remaining untreated regions. I don't see any problem in a control region being used as a donor in multiple synthetic control experiments.

This method may make most sense if the treatments were different in kind or magnitude. That is, when we do not expect the effects of the intervention to be similar in each treated region.

Optionally, if we have a natural hierarchy then we can choose the control regions based upon this. For example, if we have one test store per geographical state, then we could use the relevant store data for each of the separate tests.

## 2. Aggregated synthetic control
If the treatments are similar and we expect the effects of the intervention to be similar across treated regions, then it may make sense to create a new aggregate (e.g. `mean` or `median`) test region

## TODO
This issue can be closed by the addition of a new notebook example which demonstrates each of these approaches. It should operate on simulated data where we simulate known intervention effects on multiple regions.

Though it could also be useful to create new functionality, so a function or class called something like `MultiCellSyntheticControl` and we can pass kwargs to provide the test region and method (independent vs aggregated)

drbenvincent:
* Closes #320
* Closes #165 
* Adds a notebook to demonstrate 2 methods (pooled and unpooled) of analysing multi-cell geo lift tests.
* Also adds code to generate a synthetic dataset.

## Create new classes?
The PR currently does not add any additional code or classes. All the new functionality is embedded within the example notebook. It is relatively simple, we are just creating an aggregate treated region or iterating through each treated geo. So at the moment this uses an approach of transparency so that users can see what is being done etc.

However, part of the point of CausalPy is to provide a simple to use API, not requiring the user to do that much manual python coding. So the question is whether we should create new classes / an API. It would be something along the lines of:

For the pooled approach:

```python
result = multi_cell_geo_test_aggregate(
    df, 
    agg_func=median,
    experiment=cp.pymc_experiments.SyntheticControl,
    expt_kwargs={"treatment_time": treatment_time,
                 "formula": formula},
    model=cp.pymc_models.WeightedSumFitter,
    model_kwargs={sample_kwargs: {"target_accept": 0.95, "random_seed": seed}},
)
```

For the unpooled approach:

```python
results = multi_cell_geo_test_unpooled(
    df,
    treated_geos=treated,
    untreated_geos=untreated,
    experiment=cp.pymc_experiments.SyntheticControl,
    expt_kwargs={"treatment_time": treatment_time,
                 "formula": formula},
    model=cp.pymc_models.WeightedSumFitter,
    model_kwargs={sample_kwargs: {"target_accept": 0.95, "random_seed": seed}},
)
```

So the trade-off would be having a relatively clean API, but at the expense of making the operation a little more opaque. The manual python code in the notebook (as it stands right now) is not that complex. So I think it's not overwhelmingly obvious which we should go with.

## TODO's based on feedback so far

- [ ] ~~Check the pre-commit checks are applying ruff formatting to notebooks.~~ We'll do this in that in #340 and apply before the next release.
- [x] We're currently getting negative sales in the synthetic dataset. So need to check the synthetic data generation code and potentially consider a weighted sum model operating on log outcome data.
- [ ] ~~Fix the legends overlapping with plot content~~. We'll deal with that in #341 and run it on this notebook (and others) before the next release.
- [x] Add a section which compares the results from the pooled and unpooled approaches. The similarity or difference will be dependent on the nature of the synthetic data of course - are we simulating with identical causal impacts in all test geos, or heterogeneous causal impacts in the test geos.
- [ ] ~~Think about using fixed effects approaches https://matheusfacure.github.io/python-causality-handbook/14-Panel-Data-and-Fixed-Effects.html. And/or thinking about "What if you de-mean the data (in time and in unit) as in the book and apply the synthetic control model instead of the classic linear regression"~~ This is a great suggestion, but I think we are holding off and waiting until @juanitorduz investigates. We can certainly update this notebook in the future.


drbenvincent:
<p>Fixed. Created <a href="https://github.com/pymc-labs/CausalPy/issues/340" target="_blank">https://github.com/pymc-labs/CausalPy/issues/340</a> to double check ruff for notebooks.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/CausalPy/pull/338/discussion/#comment-2100519793'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Applied the tab20 colormap.</p><p> </p><p>Good spot with the negative sales. This is not expected. I'll add this to the todo list - I'll go back to the synthetic data generation and ponder whether we need to operate on log sales etc. </p>

---

 <a href='https://app.reviewnb.com/pymc-labs/CausalPy/pull/338/discussion/#comment-2100519816'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Good point. I'll actually create an issue about this because this is an important plot type in CausalPy and we want a decent solution which will be useful in many contexts.</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/CausalPy/pull/338/discussion/#comment-2100519834'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>Agreed. Will add this to the todo list and update very soon</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/CausalPy/pull/338/discussion/#comment-2100519847'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
> Actually! What if you de-mean the data (in time and in unit) as in the book and apply the synthetic control model instead of the classic linear regression💡
> 
> This could be the better way 🤔

So this is interesting @juanitorduz. But I think it will have implications in terms of interpolation/extrapolation and kind of change the nature of the model we are using - in that now we might need to use a `ZeroSumNormal` prior for the weights, rather than a `Dirichlet`?

In the situation where we have some large geos (with many sales) and some small geos (few sales), this scaling would also be scaling up/down the observation noise. I can't quite think through the implications of this at the moment, but is there a reason why this isn't done in situations where extrapolation seems to be required? Eg when a target geo is outside the convex hull.

juanitorduz:
Maybe we can leave this out from this PR and I can try to test it myself? 😄

drbenvincent:
> Maybe we can leave this out from this PR and I can try to test it myself? 😄

Sounds good. We can update the example at a later date with more content.

Also saw this...

> Kim, S., Lee, C., & Gupta, S. (2020). Bayesian Synthetic Control Methods. Journal of Marketing Research, 57(5), 831-852. https://doi.org/10.1177/0022243720936230

drbenvincent:
@juanitorduz... So I addressed many of the comments. Today I fixed the negative sales (whoops) and added a section comparing the approaches. The comments I've not yet addressed, I felt were appropriate to bundle up into separate issues (see the PR description up top).

Let me know what you think.

juanitorduz:
I left a small comment regarding a plot. We can merge this one an iterate. I think the killer feature would be to add a hierarchical model. This will close the gap between the pooled and unpooled models :D

drbenvincent:
<p>Good point. I've changed the colours</p>

---

 <a href='https://app.reviewnb.com/pymc-labs/CausalPy/pull/338/discussion/#comment-2179110400'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

drbenvincent:
<p>I agree. I'll add an issue for that.</p>

drbenvincent:
Replaced manual multiple forest plots with the nice feature of plot_forest that allows you to compare multiple models. Eg.
Screenshot 2024-06-19 at 20 38 17

@juanitorduz this should be good for approval + merging now?

juanitorduz (approved):

PR #346, opened 2024-06-04
Remove `<BLANKLINE>` from docstrings

### LINKED ISSUE #331: Remove the `<BLANKLINE>` from doctests when we are able
* The PR #330 fixes some failing doctests. There were a couple of things going on there, but one of the main reasons was because (probably) the progress bar was outputting empty lines.
* Turns out that this is the case and @tomicapretto is already on it with https://github.com/pymc-devs/pymc/pull/7290
* Once this PR is merged and the changes filter through into a new PyMC release, then we can go back to the docstrings and remove all the `<BLANKLINE>` instances.

drbenvincent:
* Removes a bunch of `<BLANKLINE>` instances in the docstrings. They were put there in #330 in order to fix failing tests. And that was happening because the progress bar was misbehaving.
* This was fixed in https://github.com/pymc-devs/pymc/pull/7290 so we no longer need most `<BLANKLINE>` instances.
* This PR removes those, **and increments the required version of PyMC to `5.15.1`** which is the version to include https://github.com/pymc-devs/pymc/pull/7290
* Closes #331 
* This should also fix failing remote tests that we are seeing in other PR's.

### DIFF
--- a/causalpy/pymc_experiments.py
+++ b/causalpy/pymc_experiments.py
@@ -102,11 +102,6 @@ class ExperimentalDesign:
         ...                 "progressbar": False
         ...             }),
         ...  )
-        <BLANKLINE>
-        <BLANKLINE>
-        <BLANKLINE>
-        <BLANKLINE>
-        <BLANKLINE>
         >>> result.print_coefficients(round_to=1) # doctest: +NUMBER
         Model coefficients:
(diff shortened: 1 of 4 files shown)


NathanielF (approved):
Looks good to me. Be nice to get rid of those blanklines

PR #352, opened 2024-06-14
Fix up ruff pre-commit checks to operate on notebooks

### LINKED ISSUE #340: Check pre-commit ruff formatting for notebooks is set up correctly
I'm not 100% sure the pre-commit checks are fully/properly working for ruff formatting of notebooks. Currently we have:

https://github.com/pymc-labs/CausalPy/blob/9a4daf7cb4b86c56c3f569a18f0212482c3a1986/.pre-commit-config.yaml#L26-L35

Maybe we don't need `nbQA` at all and solely use `ruff-pre-commit`? The README at https://github.com/astral-sh/ruff-pre-commit shows that it's pretty simple to make it apply to Jupyter notebooks.

drbenvincent:
* Closes #340
* Removes nbQA pre-commit check - was failing to properly ruff format notebooks
* Replaced with ruff-pre-commit hook, https://github.com/astral-sh/ruff-pre-commit
* Also includes the formatting applied to existing notebooks


NathanielF (approved):
Makes sense

PR #353, opened 2024-06-14
Remove all remaining traces of `nbqa` from the repo

drbenvincent:
* In #352 I forgot to nuke all mention of nbqa from the repo
* This PR does that, removing it from `pyproject.toml` and the `Makefile`


juanitorduz (approved):
+1

Issue #305, opened 2024-03-05
Have more coherent testing of the `summary` method


drbenvincent:
At the moment it seems that we only test the `summary` method for the `pymc_experiments.DifferenceInDifferences` class, and that is in the `test_pymc_experiments.py` file...

https://github.com/pymc-labs/CausalPy/blob/3bb81a702a19e725b1b180b8ddbc1aa3e453204c/causalpy/tests/test_pymc_experiments.py#L10-L23

We should make sure that we are testing the summary method for all of the `pymc_experiment` classes. Though I believe the `InstrumentalVariable` class doesn't have a `summery` method.

This _could_ be done by adding more to the current `test_pymc_experiments.py` file, but it could be worth thinking about simply adding to existing tests in `test_integration_pymc_examples.py`. My thinking is that this would involve less setup / computation time as we wouldn't be doing any additional model fitting.

Issue #174, opened 2023-03-28
self.summary should return an object


matekadlicsko:
I think self.summary should return a Summary object that has an HTML representation. Is this something you thought about already? If not, I'd love to implement a sa

matekadlicsko:
It would look a little something like this:
![image](https://user-images.githubusercontent.com/6461373/228217918-95e46f0e-527d-455f-9ae3-e38409c452cb.png)

PR #355, opened 2024-06-14
Enable `summary` method for all currently implemented frequentist experiments

### LINKED ISSUE #19: SC + ITS: quantitative outputs
Need to provide quantitative outputs/reports for synthetic control and interrupted time series.

The [Causal Impact](https://google.github.io/CausalImpact/CausalImpact.html) package provides these summary stats:
Screenshot 2022-11-19 at 12 49 32

For the frequentist version: Add ability to test for presence/absence of causal impact. There is a traditional way of doing this, but we could also envisage bootstrap on the pre-intervention data.

TODO
- [x] add summary stats for Bayesian ITS + SC (first pass)
- [ ] add relative impact as well as absolute
- [ ] Implement a better API
- [ ] add summary stats for Frequentist ITS + SC

drbenvincent:
* We can now call the `summary` method on all the existing frequentist experiments
* Notebooks have been updated to demonstrate the `result.summary()` functionality
* Closes #19
* Basic test coverage of the `summary` method just to get code coverage to pass. But note that #305 exists to do better testing on this front


juanitorduz:
Some tables based on `print` statements look out of phase. What if we use something like in the `pyfixest`  package https://py-econometrics.github.io/pyfixest/quickstart.html ?

drbenvincent:
> Some tables based on `print` statements look out of phase. 

Very good point. For the moment I've kept the same basic approach, but have improved the formatting. The function that prints the model coefficients now evaluates the length of the longest coefficient label so that it can align the coefficient values appropriately.

I'm very open to rethinking the nature of the summary outputs, but I'm tempted to deal with that in a separate issue. For example, there is already https://github.com/pymc-labs/CausalPy/issues/174 as one option.

I'll do a similar formatting improvement for the pymc experiments in a different issue now.

What do you think?

juanitorduz (approved):
ok! let's follow up the next iteration on the associated issue :)

PR #359, opened 2024-06-19
Improve formatting of model coefficients in `summary` method in pymc-based experiments

drbenvincent:
Similar to an improvement made in #355 to the `summary` method in scikit-learn based experiments, but for the pymc based experiments.

Printing of the model coefficients is now better horizontally aligned. The position of the coefficient values is now governed by the width of the longest coefficient label string.

I also re-ran the pymc based notebooks so that we get the updated summary output in the docs.

----
📚 Documentation preview 📚: https://causalpy--359.org.readthedocs.build/en/359/


juanitorduz (approved):
👍

PR #367, opened 2024-06-21
Add a temporal validation period to synthetic control and interrupted time series experiments

### LINKED ISSUE #364: Add validation period to synthetic control
At the moment the model parameters are estimated on the whole pre-intervention period. What we could do instead is to do parameter estimation up to a certain validation window, and produce synthetic control predictions for the validation period as well as the post-intervention period.

I believe that a motivation for this from a Frequentist perspective is possible overfitting of the training period, but this may be less of a concern when doing Bayesian parameter estimation because there can be a degree of regularisation from the priors. Nevertheless, it could be a useful feature to have.

We might want to automatically compute some goodness-of-fit metrics to the validation period.

Screenshot 2024-06-20 at 19 39 54

Inspired by https://netflixtechblog.com/round-2-a-survey-of-causal-inference-applications-at-netflix-fd78328ee0bb?gi=7d795057528e

drbenvincent:
* Closes #364
* Updates one of the synthetic control and one of the interrupted time series notebooks with quick demos of using the validation period
* This also corrects an error I found in the tests where `test_its` was testing synthetic control rather than interrupted time series 

## TODO
- [X] Implement for synthetic control
- [X] Implement for interrupted time series
- [x] Calculate Bayesian $R^2$ value separately for training period and validation period
- [x] Report Bayesian $R^2$ (including validation period where relevant) in the summary method instead of the plot title
- [x] Streamline the implementation - do we really need separate classes, or can we just add a `intervention_time` kwarg and add in additional logic to the existing classes?
- [x] Ensure class diagram is updated before requesting a review
- [x] Check docstrings are ok and doctests pass
- [x] Add test coverage
- [x] Test for `ValueError` when `validation_time` >= `treatment_time`

----
📚 Documentation preview 📚: https://causalpy--367.org.readthedocs.build/en/367/


drbenvincent:
Good progress so far. But we are missing separate Bayesian $R^2$ for training and validation phases.

Synthetic control figure:
Screenshot 2024-06-21 at 14 51 07

Interrupted time series figure:
Screenshot 2024-06-21 at 14 50 53

drbenvincent:
Based on one of @cetagostini 's PR's (https://github.com/pymc-labs/CausalPy/pull/368), I'm wondering if we should add a small feature to calculate a ROPE based on the validation period. Something a bit like this: 
 
Screenshot 2024-06-24 at 14 36 21

Any thoughts/comments welcome

causalpy/pymc_experiments.py
```diff
@@ -203,8 +219,22 @@ def __init__(
         COORDS = {"coeffs": self.labels, "obs_indx": np.arange(self.pre_X.shape[0])}
         self.model.fit(X=self.pre_X, y=self.pre_y, coords=COORDS)
 
-        # score the goodness of fit to the pre-intervention data
-        self.score = self.model.score(X=self.pre_X, y=self.pre_y)
+        if self.validation_time is None:
+            # We just have pre and post data, no validation data. So we can score the pre intervention data
+            self.score = self.model.score(X=self.pre_X, y=self.pre_y)
```
cetagostini, line 224: Do you think we could replace on validation score by `bayesian tail prob` instead of R2? So, the interpretation here is about how much the real mean during the validation diverge from the posterior mean.

drbenvincent: Sure. Just looking into it so that I get it right - I just need the high level algorithm because I've not heard that much about it. Google search for "bayesian tail probability" shows _very_ few hits. Is this a widely used approach? Doesn't matter if not, as long as it does what we want :)

PR #371, opened 2024-06-25
Change docs theme

### LINKED ISSUE #361: Restructure the API section of the docs
Right now the structure of the API docs is very flat, in that all functions and classes within a given module appear on the same page. This does give rise to quite a lot of scrolling and an increased sense of clutter.

Perhaps better ways to structure the API docs are seen in [pymc-marketing](https://www.pymc-marketing.io/en/stable/api/index.html) or [pymc](https://www.pymc.io/projects/docs/en/stable/api.html) where there is more hierarchy and each function or class has it's own page/url.

### LINKED ISSUE #369: Harmonise docs theme with `pymc-marketing` repo
Even though I like the default readthedocs theme, we should use the same theme as [pymc-marketing](https://www.pymc-marketing.io)

drbenvincent:
* This PR changes from the default readthedocs to `pydata_sphinx_theme`. So this now matches with both `pymc` and `pymc-marketing` repos.
* This is _virtually_ a straight swapsie of the themes with minimal changes to docs content. Just enough to make it work.
* Closes #369
* Closes #361

Worth checking the API docs. For example, the API docs for the class `LinearRegression` (in `pymc_model.py`) includes a LOT of stuff because we inherit from `ModelBuilder` which in turn inherits from `pm.Model`. Not saying we don't, but do we want all that stuff?

----
📚 Documentation preview 📚: https://causalpy--371.org.readthedocs.build/en/371/


OriolAbril:
> Worth checking the API docs. For example, the API docs for the class LinearRegression (in pymc_model.py) includes a LOT of stuff because we inherit from ModelBuilder which in turn inherits from pm.Model. Not saying we don't, but do we want all that stuff?

It is probably enough for many of them to mention the parent class and link to its docs (which should have those methods). That being said, the approach initially taken for pymc-marketing was to automate completely the generation of API docs. I think it would be possible to modify the template so inherited methods are not shown, but given the templates are only type specific (class/method/module) customization of the hierarchy or content depending on the module or class will end up complicating things more than generating all the rst/markdown files once and update them whenever a new object is added/removed to the library.

Everything else looks good, but at some point (especially if a pymc-analytics is created) it might help to externalize the common customizations into a labs theme

drbenvincent:
Thanks @OriolAbril. If you think this is good to go, could you approve in a review so I can merge?

OriolAbril (approved):

juanitorduz (approved):
Go 🚀 !

PR #376, opened 2024-06-26
fix docstring example rendering

### LINKED ISSUE #373: Examples in docstrings not appearing in the API docs
Following the theme change, I noticed that the example python code in the docstrings are not showing up in the rendered docs.

OriolAbril:
closes #373

----
📚 Documentation preview 📚: https://causalpy--376.org.readthedocs.build/en/376/


drbenvincent (approved):
Thanks. This is great. I guess in the future we can have a discussion about whether we want to use numpydoc. But for now, this is great :)

Issue #116, opened 2022-12-06
Add Causal Impact Wrapper from `sts-jax `


juanitorduz:
Once https://github.com/probml/sts-jax is released we could add the Causal Impact implementation in [sts-jax](https://github.com/probml/sts-jax).

drbenvincent:
Should we close this given the new BSTS functionality? @cetagostini @juanitorduz

Issue #356, opened 2024-06-17
Add test for API stability for a CausalPy example in Causal Inference and Discovery in Python by Aleksander Molak


drbenvincent:
On pages 304-307 there is an example using CausalPy in Causal Inference and Discovery in Python by @AlxndrMlk. We should create a new pytest file which tests the API to ensure that people running this code in the future are still able to replicate the results.

It should probably be done in a new test file, something like `test_api_stability.py`.

A choice will have to be about reading in the data remotely or placing it internally in this repo.

To be clear, I'm not proposing that we add that example to our docs, at least not at this point :)

drbenvincent:
Sounds great @arthurmello! Feel free to make a start - I'll assign you to this issue now.

arthurmello:
Thanks @drbenvincent! I just did the PR, looking forward to your review.

PR #381, opened 2024-07-02
Major code refactor to unify quasi experiment classes

### LINKED ISSUE #305: Have more coherent testing of the `summary` method

drbenvincent:
This is a relatively major code refactor with minor breaking changes to the API. The main purpose is to eliminate the parallel class hierarchy we had. Namely, we had virtually identical experiment classes which worked with either the PyMC or scikit-learn models. There were only slight differences here to deal with the fact that PyMC models produce `InferenceData` objects and the scikit-learn models would produce numpy arrays, for example.

We don't have an immediate intention of expanding beyond PyMC or scikit-learn models, _however_ the new code structure would make it much much easier to expand the kinds of models used. The main appeal of this is to focus on high level description of quasi-experimental methods and to abstract away from model-related implementation issues. So you _could_ add in non-PyMC Bayesian models (see #116), or use `statsmodels` (see #8) to use OLS but also get confidence intervals (which you don't get from scikit-learn models).

We should have 100% passing doctests and tests, and I re-ran all the notebooks to check that we have stable performance.

* Splits the experiment classes from one module into separate modules, so closes #253
* Adds test coverage of the `summary` method, so closes #305

## Before
classes

## After (at time of initial PR)
classes

## After (after dealing with review comments)
classes

So now we just have a single set of quasi experiment classes, all inheriting from `BaseExperiment`.

## Other changes
* I renamed `ModelBuilder` to `PyMCModel`. This seems to make more sense as it contrasts better with a new `ScikitLearnAdaptor` class/mixin which gives some extra functionality to scikit-learn models.
* I increased test coverage
* Plotting is done by experiment classes, though either the `bayes_plot` or `ols_plot` methods, though some experiment classes have custom plot methods.

## API changes
The change in API for the user is relatively small. The only change should really be how the experiment classes are imported. For example:

### Before

```python
import causalpy as cp
df = cp.load_data("did")
result = cp.pymc_experiments.DifferenceInDifferences(
    df,
    formula="y ~ 1 + group*post_treatment",
    time_variable_name="t",
    group_variable_name="group",
    model=cp.pymc_models.LinearRegression(sample_kwargs={"random_seed": seed}),
)
```

### After
The import changes from `cp.pymc_experiments.DifferenceInDifferences` to `cp.DifferenceInDifferences`.

```python
import causalpy as cp
df = cp.load_data("did")
result = cp.DifferenceInDifferences(
    df,
    formula="y ~ 1 + group*post_treatment",
    time_variable_name="t",
    group_variable_name="group",
    model=cp.pymc_models.LinearRegression(sample_kwargs={"random_seed": seed}),
)
```

The old API will still work, but will emit a deprecation warning. At some point in the future we may remove the old API, so it is best to make this minor update to existing workflows.

## TODO's
- [x] Fix up the not quite perfect use of `if isinstance` in the experiment classes
- [x] Add missing module level docstrings to improve the auto generated API docs

----
📚 Documentation preview 📚: https://causalpy--381.org.readthedocs.build/en/381/


wd60622 (commented):
Just some initial thoughts. Still making sense of the code base and changes!
Looks of good stuff here 🚀

causalpy/experiments.py
wd60622, line 1: To the point of the submodule that has all the experiment files, this to me seems like a file you'd want to import from. However, this only contains the base class for all Experiements

wd60622 (commented):
Taking a second look. 

I like the separation of the classes a lot more than before. The classes seem much smaller too which is easy to digest as a new reader of the package.

A few more impressions I've gotten:
1. I don't really see the point in having a mixin for data validation of the experiements. Just put them in the class as methods and keep them in the same place rather than jumping around files. 
   - If you would like to provide flexible of data validation long term, inject a class rather than using a Mixin. 
   -  Having two mixins for every experiment type makes it confusing what the `super` call to `__init__` does. 
   - NOTE: I'm not a big fan of mixins 
2. Make use of file structures. All of the `expt_...` files are off the root of package but clearly have some relationship to each other. Creating a submodule might make sense to provide these grouping in the package. Plus, the import will look very similar.
3. Deprecate the changes. Will this be 0.4.0 release (minor) change. You'd probably want to not introduce breaking changes into the API
4. I am not seeing the point of the `PlotComponent` class. It is very large and when using in experiment instance, seems to only one specific methods. This could be done with functions instead of classes. or small classes with two methods per experiment type.

juanitorduz:
@drbenvincent I merged a pre-commit PR and solved the conflict with this one in https://github.com/pymc-labs/CausalPy/pull/381/commits/133ee3bdf4f1786e894d4aeb7795c7e12f5a1f88 so should be all good. Pleas pull from this branch to update your local branch :)

causalpy/exp_inverse_propensity_weighting.py
```diff
@@ -0,0 +1,433 @@
...
+        self._t_design_info = X.design_info
+        self.labels = X.design_info.column_names
+        self.t, self.X = np.asarray(t), np.asarray(X)
+        self.y = self.data[self.outcome_variable]
+
+        COORDS = {"obs_ind": list(range(self.X.shape[0])), "coeffs": self.labels}
+        self.coords = COORDS
+        self.model.fit(X=self.X, t=self.t, coords=COORDS)
+
+    def make_robust_adjustments(self, ps):
```
juanitorduz, line 100: Even though we ar not using MyPy yet. For this refactor I suggest to use return types as well :)

causalpy/exp_inverse_propensity_weighting.py
```diff
@@ -0,0 +1,433 @@
...
+        weighted_outcome0 = (1 - t[t == 0]) * outcome_ntrt * i_propensity0
+        return weighted_outcome0, weighted_outcome1, n_ntrt, n_trt
+
+    def make_doubly_robust_adjustment(self, ps):
+        """The doubly robust weighting scheme is also
+        discussed in Aronow and Miller, but a bit more generally
+        than our implementation here. Here we have specified
+        the outcome model to be a simple OLS model.
+        In this way the compromise between the outcome model and
+        the propensity model is always done with OLS."""
```
juanitorduz, line 165: What about having docstrings as Numpy (pymc-marketing)? The user mainly wants to know what `ps` is and what the function does. The rest is a (useful!) comment which can also be added.

causalpy/exp_inverse_propensity_weighting.py
```diff
@@ -0,0 +1,433 @@
...
+            trt = np.sum(weighted_outcome_trt) / np.sum(n_trt)
+        else:
+            (
+                weighted_outcome_ntrt,
+                weighted_outcome_trt,
+                n_ntrt,
+                n_trt,
+            ) = self.make_doubly_robust_adjustment(ps)
+            trt = np.mean(weighted_outcome_trt)
+            ntrt = np.mean(weighted_outcome_ntrt)
```
juanitorduz, line 217: Each of these conditions should be individual functions and have their own unit-test. Then `get_ate` simply can call them.

causalpy/exp_inverse_propensity_weighting.py
```diff
@@ -0,0 +1,433 @@
...
+        ate = trt - ntrt
+        return [ate, trt, ntrt]
+
+    def plot_ATE(self, idata=None, method=None, prop_draws=100, ate_draws=300):
```
juanitorduz, line 221: not fan of upper-case un function names(very personal opinion)

causalpy/experiments.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+Base class for quasi experimental designs.
+"""
+
+
+class ExperimentalDesign:
+    """Base class for quasi experimental designs."""
+
+    model = None
+    expt_type = None
```
juanitorduz, line 23: why do you need this if they are in the constructor __init__ method?

causalpy/expt_diff_in_diff.py
```diff
@@ -0,0 +1,203 @@
...
+            # just the treatment period(s)
+            .query("post_treatment == True")
+            # drop the outcome variable
+            .drop(self.outcome_variable_name, axis=1)
+            # We may have multiple units per time point, we only want one time point
+            .groupby(self.time_variable_name)
+            .first()
+            .reset_index()
+        )
+        assert not self.x_pred_counterfactual.empty
```
juanitorduz, line 148: As in the pymc-marketing repo, It is better to raise an error than having an assert statement (jus a suggestion)

causalpy/expt_diff_in_diff.py
```diff
@@ -0,0 +1,203 @@
...
+            # This is the coefficient on the interaction term
+            coeff_names = self.model.idata.posterior.coords["coeffs"].data
+            for i, label in enumerate(coeff_names):
+                if "post_treatment" in label and self.group_variable_name in label:
+                    self.causal_impact = self.model.idata.posterior["beta"].isel(
+                        {"coeffs": i}
+                    )
+        elif isinstance(self.model, ScikitLearnModel):
+            # This is the coefficient on the interaction term
+            # TODO: THIS IS NOT YET CORRECT ?????
```
juanitorduz, line 170: Is there an issue for this TODO?

juanitorduz (commented):
Some initial suggestions

causalpy/exp_inverse_propensity_weighting.py
wd60622: +1

causalpy/exp_inverse_propensity_weighting.py
```diff
@@ -0,0 +1,433 @@
...
+import pandas as pd
+from matplotlib.lines import Line2D
+from patsy import dmatrices
+from sklearn.linear_model import LinearRegression as sk_lin_reg
+
+from causalpy.data_validation import PropensityDataValidator
+from causalpy.experiments import ExperimentalDesign
+
+# from causalpy.pymc_models import PyMCModel
+# from causalpy.utils import round_num
```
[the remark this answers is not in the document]
drbenvincent: done

causalpy/__init__.py
```diff
@@ -13,17 +13,31 @@
...
 from causalpy.version import __version__
 
 from .data import load_data
+from .exp_inverse_propensity_weighting import InversePropensityWeighting
+from .expt_diff_in_diff import DifferenceInDifferences
+from .expt_instrumental_variable import InstrumentalVariable
+from .expt_prepostfit import InterruptedTimeSeries, SyntheticControl
+from .expt_prepostnegd import PrePostNEGD
+from .expt_regression_discontinuity import RegressionDiscontinuity
+from .expt_regression_kink import RegressionKink
```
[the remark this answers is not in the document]
drbenvincent: I've now added an `experiments` submodule

causalpy/experiments.py
drbenvincent: So would you recommend putting the base `ExperimentalDesign` class into `causalpy/base_classes.py`, or simply rename `experiments.py` to `expt_base_class.py`?

causalpy/expt_diff_in_diff.py
drbenvincent: Done. I've removed all asserts outside of tests and replaced with checks which raise exceptions

causalpy/experiments.py
drbenvincent: removed in a commit that will soon be pushed

causalpy/exp_inverse_propensity_weighting.py
drbenvincent: global find replace plot_ATE -> plot_ate done in a commit soon to be pushed

causalpy/experiments.py
wd60622: Think the submodule pretty much takes case of it. 
`from causalpy.experiments.base import BaseExperiment` might also make sense as an alternative name in order signify that it doesn't have a purpose on its own

drbenvincent: Done: changed `experiment.py::ExperimentalDesign` to `base.py::BaseExperiment`

causalpy/plotting.py
```diff
@@ -0,0 +1,627 @@
...
+from causalpy.utils import round_num
+
+LEGEND_FONT_SIZE = 12
+az.style.use("arviz-darkgrid")
+
+
+class PlotComponent(ABC):
+    """Abstract Base Class for PlotComponent."""
+
+    pass
```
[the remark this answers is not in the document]
drbenvincent: It's unnecessary. I'm removing it now, though I notice that there are some comments about rethinking the plot component approach - will get to those shortly :)

causalpy/exp_inverse_propensity_weighting.py
drbenvincent: I've now added type hints for returns for all plotting functions

drbenvincent:
# Handling deprecation

My initial PR did not handle deprecation at all, so users using the old API would have received an error. I did update the docs, but even so, it makes a lot of sense to have deprecation warnings.

What I did was to add back the `pymc_experiments.py` and `skl_experiments.py` modules. But these now contain thin wrappers which:
a) Provide a deprecation warning
b) Pass on the arguments and keyword arguments to the replacement experiment class. This means that the old API will work. **We have an option of removing this component in the future so that the old API will be broken, but the user gets a description of how to fix the problem.**

Here's an example:

```python
from .experiments.diff_in_diff import (
    DifferenceInDifferences as NewDifferenceInDifferences,
)


def DifferenceInDifferences(*args, **kwargs):
    warnings.warn(
        """causalpy.pymc_experiments.DifferenceInDifferences is deprecated and will be removed in a future release. Please use causalpy.experiments.DifferenceInDifferences instead.""",
        DeprecationWarning,
        stacklevel=2,
    )
    return NewDifferenceInDifferences(*args, **kwargs)
```

I have kept the warning noncommittal about when the deprecation will happen. We _may_ not decide to deprecate it for some time - partially because the old API makes an appearance in a published book (see https://github.com/pymc-labs/CausalPy/issues/356).

I have also added tests so that these deprecation warnings are delivered and are not broken by any changes in the future. These are in `test_intergration_pymc_examples.py` and `test_intergration_skl_examples.py`, and an example would be:

```python
def test_did_deprecation_warning():
    """Test that the old DifferenceInDifferences class raises a deprecation warning."""

    with pytest.warns(DeprecationWarning):
        df = cp.load_data("did")
        result = cp.pymc_experiments.DifferenceInDifferences(
            df,
            formula="y ~ 1 + group*post_treatment",
            time_variable_name="t",
            group_variable_name="group",
            model=cp.pymc_models.LinearRegression(sample_kwargs=sample_kwargs),
        )
        assert isinstance(result, cp.DifferenceInDifferences)
```

## Limitations?

My only concern with the current way of handling deprecation is that the contents of `pymc_experiments.py` and `skl_experiments.py` modules are identical. Well - `skl_experiments.py` doesn't contain wrappers for all the experiment types because we don't have sci-kit learn implementations for all the experiment types. In theory these may be added in the future.

Is there a way of only having one of these files and have `cp.pymc_experiments.RegressionDiscontinuity` and `cp.skl_experiments.RegressionDiscontinuity` route through the same deprecation wrapper function? Maybe it will be obvious to me tomorrow, but I'd welcome any thoughts @juanitorduz / @wd60622.

causalpy/pymc_experiments.py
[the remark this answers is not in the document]
drbenvincent: I've now added some proper deprecation handling, see comment below https://github.com/pymc-labs/CausalPy/pull/381#issuecomment-2223279679

causalpy/skl_experiments.py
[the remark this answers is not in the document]
drbenvincent: I've now added some proper deprecation handling, see comment below https://github.com/pymc-labs/CausalPy/pull/381#issuecomment-2223279679

drbenvincent:
Excellent comments from @juanitorduz + @wd60622. I'm not done, but this is already feeling more tight.

classes

causalpy/plotting.py
```diff
@@ -0,0 +1,627 @@
...
+
+
+class BayesianPlotComponent(PlotComponent):
+    """Plotting component for Bayesian models."""
+
+    @staticmethod
+    def plot_pre_post(results, round_to=None, counterfactual_label=None):
```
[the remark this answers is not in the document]
drbenvincent: Agreed. In the latest commit https://github.com/pymc-labs/CausalPy/pull/381/commits/fbc4c9443f195d3618d73e76b64943109c94b2e4 I moved all the plot methods into the experiment classes. So now there are `ols_plot` and `bayesian_plot` methods. Though the user still just calls the `plot` method. The `BaseExperiment` now does the dispatching to either `ols_plot` and `bayesian_plot`. There should now be no trace of the `PlotComponent` stuff.

causalpy/exp_inverse_propensity_weighting.py
drbenvincent: I've created a separate issue for this https://github.com/pymc-labs/CausalPy/issues/388

causalpy/exp_inverse_propensity_weighting.py
drbenvincent: I've created a separate issue about this https://github.com/pymc-labs/CausalPy/issues/389

causalpy/expt_diff_in_diff.py
drbenvincent: Issue created https://github.com/pymc-labs/CausalPy/issues/390

causalpy/skl_models.py
```diff
@@ -18,35 +18,56 @@
...
+
+    def get_coeffs(self):
+        """Get the coefficients of the model as a numpy array."""
+        return np.squeeze(self.coef_)
+
+
+class LinearRegression(ScikitLearnModel, LinearRegression):
+    """Linear regression model for causal inference"""
+
+    pass
```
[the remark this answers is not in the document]
drbenvincent: Good idea. https://github.com/pymc-labs/CausalPy/pull/381/commits/fa640b8f90cf5a2fccb3b9d9aa8d139d46b8535d brings in the new `create_causalpy_compatible_class`. This has resulted in a small API change just when we are using scikit-learn linear models. I don't actually think many people are using the sci-kit learn models to be honest, so right now there is no deprecation management.

causalpy/experiments/diff_in_diff.py
```diff
@@ -0,0 +1,440 @@
...
+        return f"Causal impact = {convert_to_string(self.causal_impact, round_to=round_to)}"
+
+    def bayesian_plot(self, round_to=None, **kwargs) -> tuple[plt.Figure, plt.Axes]:
+        """
+        Plot the results
+
+        :param round_to:
+            Number of decimals used to round results. Defaults to 2. Use "None" to return raw numbers.
+        """
+        round_to = kwargs.get("round_to")
```
wd60622, line 236: Remove since explicit

```suggestion
```

causalpy/skl_experiments.py
```diff
@@ -11,721 +11,64 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-"""Quasi-Experiment classes for OLS inference"""
+"""
+This module exists to maintain the old API where experiment classes were defined in the
+`causalpy.skl_experiments` namespace. They have moved to `causalpy.experiments` and
+this module is a thin wrapper around the new classes to maintain backwards
+compatibility. A deprication warning is delivered to the user. This module may be
```
wd60622, line 18: This isn't a british / american difference is it? 😄 
```suggestion
compatibility. A deprecation warning is delivered to the user. This module may be
```

causalpy/experiments/base.py
```diff
@@ -0,0 +1,63 @@
...
+        else:
+            raise ValueError("Unsupported model type")
+
+    @abstractmethod
+    def bayesian_plot(self, *args, **kwargs):
+        """Abstract method for plotting the model."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def ols_plot(self, *args, **kwargs):
```
wd60622, line 61: Will `round_to` be a parameter for all? Is so, I find being explicit is way to go
But not required

wd60622 (commented):
Still mid review

causalpy/experiments/base.py
```diff
@@ -0,0 +1,63 @@
...
+from abc import abstractmethod
+
+from causalpy.pymc_models import PyMCModel
+from causalpy.skl_models import ScikitLearnModel
+
+
+class BaseExperiment:
+    """Base class for quasi experimental designs."""
+
+    def __init__(self, model=None, **kwargs):
```
wd60622, line 27: What do the kwargs do? They are passed through in the child classes. However, they just seems error prone as people will silently be able to pass additional kwargs

causalpy/experiments/base.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @abstractmethod
+    def bayesian_plot(self, *args, **kwargs):
+        """Abstract method for plotting the model."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def ols_plot(self, *args, **kwargs):
+        """Abstract method for plotting the model."""
+        raise NotImplementedError
```
wd60622, line 63: This seems to be hit with a few models. Include a warning message there

docs/source/api/index.md
```diff
@@ -8,8 +8,14 @@
...
-  skl_experiments
   skl_models
+  experiments
+  expt_diff_in_diff
+  expt_prepostnegd
+  expt_prepostfit
+  expt_regression_discontinuity
+  expt_regresion_kink
+  expt_instrumental_variable
+  expt_inverse_propensity_weighting
```
wd60622, line 20: Remove now

causalpy/experiments/base.py
```diff
@@ -0,0 +1,63 @@
...
+"""
+
+from abc import abstractmethod
+
+from causalpy.pymc_models import PyMCModel
+from causalpy.skl_models import ScikitLearnModel
+
+
+class BaseExperiment:
+    """Base class for quasi experimental designs."""
```
wd60622, line 25: It might be either as the reader to have `supports_ols` `supports_bayes` class attributes. Just because those are at the top of the child classes. Or use type hints for the models instead of just None

Otherwise I have to get into the weeds of the code for each.

First might be like this
```python
class BaseExperiment:
    """Base class for quasi experimental designs."""
    supports_bayes: bool
    supports_ols: bool
   
    def __init__(self, model) -> None:
        if model is not None:
            self.model = model

        if isinstance(self.model, PyMCModel) and not self.supports_bayes: raise ...
        if isinstance(self.model, ScikitLearnModel) and not self.supports_ols: raise ...

        if self.model is None:
            raise ValueError("fitting_model not set or passed.")
    ...


class DifferenceInDifference(BaseExperiment): 
    supports_ols = True
    supports_bayes = True


class RegressionKink(BaseExperiment): 
    supports_bayes = True
    supports_ols = False
```

wd60622 (commented):
In terms of a refactor, I think this is fairly good. Have a few more suggestions. Also some more questions on expected user behavior and how backward compatible you'd like this to be. And I have a few more general points but it might go more into (new) functionality / structure. 

1. Would it be convenient to be able to pass scikit-learn models as well and use the `create_causalpy_compatible_class` behind within the class? It seems like that was the behavior before and might be break some previous workflows. I see most of the tests just do it once at the top which could then be avoided.
2. The import of experiments in the warnings do not work. Do you want people to use `cp.DifferenceInDifferences` or `cp.experiments.DifferenceInDifferences` or both?
3. The `fit` method should ideally have the same signature. This leads to the if elif blocks that are all over the place. Though this might be another refactor

causalpy/experiments/base.py
drbenvincent: Good spot. There were some unnecessary karts being passed in for the DID experiments which were just going into a black hole here. I've removed `**kwargs` from being passed to `BaseExperiment.__init__`

docs/source/api/index.md
drbenvincent: Well spotted. Done.

causalpy/experiments/base.py
drbenvincent: This is a great idea. I've implemented this in https://github.com/pymc-labs/CausalPy/pull/381/commits/cc62438e7adbc30e68fbcd25044475d57e7aa758. I've also added tests to check that exceptions are raised when providing old models to experiment classes that only support bayesian models. These can be removed over time if that support is added.

causalpy/experiments/base.py
drbenvincent: I've added a small warning here in https://github.com/pymc-labs/CausalPy/pull/381/commits/dede64a281070d5094126927da279c184e658166 Not sure if you had something specific in mind?

drbenvincent:
Thanks @wd60622. Still working through the comments, but I didn't follow what you meant about the deprecation warnings not working.

When I run with the old API (e.g. `cp.pymc_experiments.DifferenceInDifferences`) then we do get a warning, followed with the rest 

Screenshot 2024-08-07 at 20 52 36

I'm also seeing the deprecation tests pass (at the bottom of `test_integration_pymc_examples.py`)

drbenvincent:
> 1. Would it be convenient to be able to pass scikit-learn models as well and use the `create_causalpy_compatible_class` behind within the class? It seems like that was the behavior before and might be break some previous workflows. I see most of the tests just do it once at the top which could then be avoided.

This is a great suggestion @wd60622, and stopped me from being a bit lazy! I've addressed this in https://github.com/pymc-labs/CausalPy/pull/381/commits/02dacb213ca4d41604aee7f236fda82b4223c04c

The first attempt I tried turned out to be a dead end. It basically came down to the fact that we are passing in an instantiated model object, not a model class. This means that if the model was built with non-trivial kwargs, it was getting highly complex to create another class with the mixin approach. 

So the solution I went with was to simply take the user-provided model instance and to use a helper function to attach the methods required by CausalPy to that model instance. I did use GPT to help with this function :)

```python
def add_mixin_methods(model_instance, mixin_class):
    for attr_name in dir(mixin_class):
        attr = getattr(mixin_class, attr_name)
        if callable(attr) and not attr_name.startswith("__"):
            # Bind the method to the instance
            method = attr.__get__(model_instance, model_instance.__class__)
            setattr(model_instance, attr_name, method)
    return model_instance
```

So now users can just provide unadulterated scikit-learn model instances. The experiment base class then adds the required methods to this behind the scenes. So there is zero API change for the user.

drbenvincent:
> 3. The fit method should ideally have the same signature. This leads to the if elif blocks that are all over the place. Though this might be another refactor.

So this is an example in `DifferencesInDifferences`:
```python
# fit model
if isinstance(self.model, PyMCModel):
    COORDS = {"coeffs": self.labels, "obs_indx": np.arange(self.X.shape[0])}
    self.model.fit(X=self.X, y=self.y, coords=COORDS)
elif isinstance(self.model, RegressorMixin):
    self.model.fit(X=self.X, y=self.y)
else:
    raise ValueError("Model type not recognized")
```

You are right, we only need all this type checking because of the different signatures of the `fit` methods. I can't not provide coords to the pymc fit method. I think the only other alternative is to create and pass in coords to both fit methods - so I'd have to override the default fit method of the scikit learn objects to make it accept `**kwargs` and the coords would basically disappear into a black hole.

It's maybe a bit clunky because you are needlessly defining and passing coords when you have scikit-learn models, but that seems like the easiest way to go.

However, this would only affect internal code and not affect the user experience. So I'm happy if we sit and think about this one and address it at a later date. But open to any other ideas.

causalpy/experiments/__init__.py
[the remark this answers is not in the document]
drbenvincent: I'm unclear on this comment. See a response below https://github.com/pymc-labs/CausalPy/pull/381#issuecomment-2274245007

causalpy/pymc_experiments.py
```diff
@@ -11,1882 +11,108 @@
...
+)
+from .experiments.regression_kink import RegressionKink as NewRegressionKink
+
+# Ensure deprecation warnings are always shown in Jupyter Notebooks
+warnings.simplefilter("always", DeprecationWarning)
+
+
+def PrePostNEGD(*args, **kwargs):
+    warnings.warn(
+        """causalpy.pymc_experiments.PrePostNEGD is deprecated and will be removed in a future release. Please use causalpy.experiments.PrePostNEGD instead.""",
```
wd60622, line 51: I can clarify with this as an example.

The recommendation here is that the user should use `causalpy.experiments.PrePostNEGD`. However, that isn't a real import. Also, most of your examples use `cp.PrePostNEGD` as you make the class available in the `causalpy/__init__.py`

What doesn't work: 
```python
# Recommended and doesn't work
import causalpy

causalpy.experiments.PrePostNEGD

# Also doesn't work
import causalpy.experiments

causalpy.experiments.PrePostNEGD

# Works but not recommended and probably not desired solution
from causalpy.experiments.prepostnegd import PrePostNEGD

PrePostNEGD


# Works and is used in the documentation
import causalpy as cp

cp.PrePostNEGD

# Or also since same but without alias
import causalpy

causalpy.PrePostNEGD
```

So just recommend either of the last two for all of these deprecations. Or import these in the `causalpy/experiments/__init__.py` in order to make this recommendation work

drbenvincent: Thanks, now I get it :)

Fixed in https://github.com/pymc-labs/CausalPy/pull/381/commits/01ce582f61468df45f2cde0479391389b9c28985. I've also updated it so that the depreciation warning text shows up as red in the terminal output, so should be more visible.

causalpy/skl_experiments.py
```diff
@@ -11,721 +11,64 @@
...
+    return NewSyntheticControl(*args, **kwargs)
 
-        .. warning::
 
-            Assumes treatment is given to those ABOVE the treatment threshold.
-        """
-        return np.greater_equal(x, self.treatment_threshold)
+def DifferenceInDifferences(*args, **kwargs):
+    warnings.warn(
+        """causalpy.skl_experiments.DifferenceInDifferences is deprecated and will be removed in a future release. Please use causalpy.experiments.DifferenceInDifferences instead.""",
```
wd60622, line 52: Same deprecation suggestion here

causalpy/pymc_experiments.py
```diff
@@ -11,1882 +11,126 @@
...
-    RDDataValidator,
-    RegressionKinkDataValidator,
-    PrePostNEGDDataValidator,
-    IVDataValidator,
-    PropensityDataValidator,
+"""
+This module exists to maintain the old API where experiment classes were defined in the
+`causalpy.pymc_experiments` namespace. They have moved to `causalpy.experiments` and
+this module is a thin wrapper around the new classes to maintain backwards
+compatibility. A deprication warning is delivered to the user. This module may be
```
wd60622, line 18:
```suggestion
compatibility. A deprecation warning is delivered to the user. This module may be
```

causalpy/skl_experiments.py
drbenvincent: Well spotted - I've updated for the scikit-learn experiments in https://github.com/pymc-labs/CausalPy/pull/381/commits/66f22aaf0a428132b02f298b6d13032f81214413

drbenvincent:
Thanks for the final points @wd60622. Fingers crossed, that should be it now?

causalpy/__init__.py
```diff
@@ -13,18 +13,38 @@
 #   limitations under the License.
 import arviz as az
 
-from causalpy import pymc_experiments, pymc_models, skl_experiments, skl_models
+import causalpy.pymc_experiments as pymc_experiments  # to be depricated
+import causalpy.pymc_models as pymc_models
+import causalpy.skl_experiments as skl_experiments  # to be depricated
```
wd60622, line 18:
```suggestion
import causalpy.pymc_experiments as pymc_experiments  # to be deprecated
import causalpy.pymc_models as pymc_models
import causalpy.skl_experiments as skl_experiments  # to be deprecated
```

causalpy/tests/test_misc.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+Miscellaneous unit tests
+"""
+
+import causalpy as cp
+
+sample_kwargs = {"tune": 20, "draws": 20, "chains": 2, "cores": 2}
+
+
+def test_regression_kink_gradient_change():
```
wd60622, line 23:
```suggestion
import causalpy as cp

sample_kwargs = {"tune": 20, "draws": 20, "chains": 2, "cores": 2}


def test_regression_kink_gradient_change():
```
```suggestion
import causalpy as cp


def test_regression_kink_gradient_change():
```

wd60622 (approved):
Looks good! Just two things I noticed

causalpy/tests/test_misc.py
drbenvincent: Good catch, have removed `sample_kwargs`

causalpy/__init__.py
drbenvincent: Thanks - did a global find/replace

twiecki:
Congrats!

PR #407, opened 2024-09-10
Add test for API stability for a CausalPy example in Causal Inference and Discovery in Python by Aleksander Molak

### LINKED ISSUE #356: Add test for API stability for a CausalPy example in Causal Inference and Discovery in Python by Aleksander Molak

arthurmello:
Closes #356 

- Saves the original dataset in causalpy/data/gt_social_media_data.csv
- Processes data in the same way as [the original example](https://github.com/PacktPublishing/Causal-Inference-and-Discovery-in-Python/blob/27ebda4090c54676cc4d99323d972b41ef134bb2/Chapter_11.2.ipynb#L18)
- Asserts that `WeightedSumFitter`, `SyntheticControl` and the `results` plot all run without errors

----
📚 Documentation preview 📚: https://causalpy--407.org.readthedocs.build/en/407/


drbenvincent (changes requested):
Thanks @arthurmello ! The only change that's needed is to speed up the tests by asking for fewer samples. We do this in some of the other tests, such as `test_integration_pymc_example.py`. So we just need to define custom sample kwargs `sample_kwargs = {"tune": 20, "draws": 20, "chains": 2, "cores": 2}` and pass those in when the model object is first created.

While that is a slight deviation from the original code, it doesn't represent a meaningful change, and it speeds up the tests which is important both locally and remotely.

arthurmello:
@drbenvincent yes, it makes sense! I have included those, let me know if anything else is needed.

drbenvincent (approved):
Looks good to me :)

Thanks for the PR @arthurmello

PR #403, opened 2024-09-02
Implement the codespell pre-commit hook

westford14:
Closes #401 -- I ignored all the jupyter notebooks as there's a lot of custom variable definition in there that is getting flagged as incorrect words.  I felt like trying to maintain a list of all these words, long term, as a dictionary would be a bit too tricky / annoying. 

----
📚 Documentation preview 📚: https://causalpy--403.org.readthedocs.build/en/403/

PR #411, opened 2024-09-16
Codespell CI Job

westford14:
From conversation here: https://github.com/pymc-labs/CausalPy/pull/403#discussion_r1752040083 -- we are now separating the `codespell` check for the jupyter notebooks into a separate CI job since they _most likely_ don't change as often and running it within a `pre-commit` check is overkill 

----
📚 Documentation preview 📚: https://causalpy--411.org.readthedocs.build/en/411/


westford14:
@drbenvincent -- here's the related PR to use the codespell check on the jupyter notebooks within CI instead of the pre-commit check

drbenvincent:
Looks good @westford14. Let's see if @maresb has any specific comments...

maresb (approved):
This looks great! Thanks a lot @westford14 for splitting it up like this.

Issue #422, opened 2024-10-11
Tidy up docs home page


drbenvincent:
There was a recent refresh of the [`pymc-marketing`](https://www.pymc-marketing.io/) docs homepage, and it's looking quite nice.

Let's make the following changes:
- [ ] Add download stats badges (@juanitorduz might be interested in doing this?)
- [ ] Change the **Features** section
  - [ ] Remove all the images
  - [ ] Add a table to outline all the experiment types. Perhaps having ticks/crosses for Bayes and OLS versions
  - [ ] Add a table to list out all the custom PyMC model types and what they are used for

PR #423, opened 2024-10-11
Update README and home page of the docs

drbenvincent:
Partially addresses #422


juanitorduz (approved):
Regarding the PR itself it looks good. 

Still, personally, I think I like the plots more, but this is just my personal opinion. The table is also good :)

drbenvincent:
I know what you mean, but it started to feel cluttered.

PR #429, opened 2024-11-19
Calculate goodness of fit (`r2_score`) based on posterior expectation rather than posterior prediction

### LINKED ISSUE #428: Fix failing doctest
With `pymc 5.18.2`, and `pytensor 2.26.3` the doctest for `PyMCModel` is failing

https://github.com/pymc-labs/CausalPy/blob/142694fb7e16b5f965a1cc3ed7ec28f3484a1c3f/causalpy/pymc_models.py#L29-L68

drbenvincent:
* Previously, the `r2_score` was calculated by comparing `y_true` and `y_pred`, where `y_pred` was the likelihood-influenced predictions i.e. with the observation noise added. Now we set `y_pred` to be based on the model's expected outcomes.
* Closes #428

----
📚 Documentation preview 📚: https://causalpy--429.org.readthedocs.build/en/429/

### DIFF
--- a/causalpy/pymc_models.py
+++ b/causalpy/pymc_models.py
@@ -59,13 +59,13 @@ class PyMCModel(pm.Model):
     ... )
     >>> model.fit(X, y)
     Inference data...
-    >>> X_new = rng.normal(loc=0, scale=1, size=(20,2))
-    >>> model.predict(X_new)
-    Inference data...
     >>> model.score(X, y)
     r2        0.19157
     r2_std    0.11238
     dtype: float64
+    >>> X_new = rng.normal(loc=0, scale=1, size=(20,2))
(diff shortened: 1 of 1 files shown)


drbenvincent:
Oh, I think this is because of `rng` being used multiple times (in the docstring there was a call to `rng.normal` before the `model.score`) which changed its state. Simple fix will be to rearrange order of calculations in the docstring. Though I'm not totally on top of why doctests were passing locally and not remotely.

drbenvincent:
Why are these f*king doctests passing locally but not remotely. Does `rng` have different behaviour on different systems or numpy versions?

cetagostini-wise (approved):

cetagostini (approved):

Issue #456, opened 2025-04-19
Upgrade synthetic control to model multiple treated units


drbenvincent:
## What
Currently, the synthetic control functionality is constrained to a single treatment unit. Clearly, having one treated unit is the minimum you could have for a working synthetic control solution. This still offers non-trivial functionality, and we have docs with a generic example with simulated data, and also for the effects of Brexit (the UK is the only treated unit).

However, there are many situations where you will have more than one treated unit. This could happen in many different domains, but it will be notable in marketing with geolift situations. We also have a docs page on geolift with a single treated geo. We also have a docs page on multi-cell geolift analysis where we have multiple treated geos. That docs page currently walks through an example of a pooled analysis approach where we simply take the average of the outcome variable across the treated geos and then proceed to model it as a single treated unit case of synthetic control. The alternative was to treat the geos as unpooled - in that case we simply run _multiple_ independent single treated unit synthetic control analyses.

## Why
This issue proposes that we add the ability to model multiple treated units (or geos). This is has a number of motivations:
* it is a more general solution
* it would allow a single modeling approach to geo testing (or any other multiple treatment unit situation)
* it would allow the full flexibility from pooled and unpooled analysis approaches, but also newly, partially pooled analysis where there could be information sharing across weights.
* it will lay the foundation for implementing synthetic differences in differences #47

### Changes to the `WeightedSumFitter` class

This pymc model class would need to be changed so that we have a weight _matrix_, rather than a weight _vector_.

https://github.com/pymc-labs/CausalPy/blob/4227edf88fbc861181b9adfa8b0f949fc306f2be/causalpy/pymc_models.py#L254-L271

So rather than `dims="coeffs"` (where `coeffs` correspond to control units), it would be `dims=("control_units", "treated_units")`. This would give us an unpooled set of weights of each of the control units for each of the treated units. A later step could them implement partial pooling over these weights (across the `treated_unit`) dimension.

The `WeightedSumFitter.build_model` method would also change to update the fact that the raw data would no longer be long form, so the incoming data (currently a design matrix `X` would now be a 2D matrix, probably shape `("time", "unit")`.

### Changes to the `SyntheticControl` class

[...]

PR #460, opened 2025-04-21
API change for the `SyntheticControl` experiment class

### LINKED ISSUE #459: Convert all instances of `obs_indx` to `obs_ind`
This is not giving rise to any known bugs, but I've seen that the coord for rows (observations) is almost always called `obs_ind`, there are some cases where it's labelled as `obs_indx`. So we need to convert all to `obs_ind` to be consistent.

drbenvincent:
* Towards #456
* This does _not_ yet enable multiple treated units in synthetic control experiments. But it implements important prep work which will enable it. The changes focus on:
* API changes
* Bit of a spaghetti situation, but I had to switch from storing dataframes to `xarray.DataArrays`. This helps with the broadcasting. The model functions get varied input depending on the situation (e.g. experiment), so it was getting complicated. Xarray simplifies broadcasting in functions like `PyMCModel.calculate_impact`. (A bunch of time was used trying different solutions before it became clear that the xarray approach was the easiest). 
* Because the API is changing, I elected to get rid of some legacy backward compatibility/ depreciation handling stuff. So when we do the next release, this will include breaking API changes _and_ abandoning backward compatibility with an old API. I'm not fussed about this, we are currently at version `0.x`, so people should expect the API to change until we reach `1.x`.

### Remaining taks:
- [x] Check the multi cell geolift notebook is working as expected
- [x] Fix failing doctest
- [ ] ~~Resolve error below~~ Using a linear regression model with synthetic control is not a planned use case. If it becomes important then we can deal with that at the time.

### Remaining bug, not captured by tests
In the scikit-learn synthetic control notebook, we are getting an error in the second part where we call

```python
result = cp.SyntheticControl(
    df,
    treatment_time,
    control_units=["a", "b", "c", "d", "e", "f", "g"],
    treated_units=["actual"],
    model=LinearRegression(positive=True),
)
```

There is a broadcasting issue resulting in this:
Screenshot 2025-04-23 at 10 34 27

So I need to think about if doing this (linear model with synthetic control experiment) even make sense. If it is, then I need to add a test to catch this error because it's not covered by tests currently, and fix it.

----
📚 Documentation preview 📚: https://causalpy--460.org.readthedocs.build/en/460/


causalpy/experiments/diff_in_diff.py
```diff
@@ -102,6 +104,21 @@ def __init__(
         self.y, self.X = np.asarray(y), np.asarray(X)
         self.outcome_variable_name = y.design_info.column_names[0]
 
+        # turn into xarray.DataArray's
+        self.X = xr.DataArray(
+            self.X,
+            dims=["obs_ind", "coeffs"],
```
NathanielF, line 110: dims of X as coeffs? Should surely be covariates... coeffs is the output, and while they should be 1:1 naming is a little misleading no?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -125,10 +155,8 @@ def __init__(
 
         # calculate the counterfactual
         self.post_pred = self.model.predict(X=self.post_X)
-        self.pre_impact = self.model.calculate_impact(self.pre_y[:, 0], self.pre_pred)
-        self.post_impact = self.model.calculate_impact(
-            self.post_y[:, 0], self.post_pred
-        )
+        self.pre_impact = self.model.calculate_impact(self.pre_y, self.pre_pred)
```
NathanielF, line 158: definitely nicer than the [:, 0] syntax

causalpy/experiments/synthetic_control.py
```diff
@@ -41,8 +41,10 @@ class SyntheticControl(BaseExperiment):
         A pandas dataframe
     :param treatment_time:
         The time when treatment occurred, should be in reference to the data index
-    :param formula:
-        A statistical model formula
+    :param control_units:
```
NathanielF, line 44: This is the major API change. Ok cool.

causalpy/experiments/synthetic_control.py
```diff
@@ -66,64 +69,110 @@ class SyntheticControl(BaseExperiment):
...
-        self.post_X = np.asarray(new_x)
-        self.post_y = np.asarray(new_y)
 
         # fit the model to the observed (pre-intervention) data
         if isinstance(self.model, PyMCModel):
-            COORDS = {"coeffs": self.labels, "obs_ind": np.arange(self.pre_X.shape[0])}
-            self.model.fit(X=self.pre_X, y=self.pre_y, coords=COORDS)
+            COORDS = {
+                # key must stay as "coeffs" unless we can find a way to auto identify
+                # the predictor dimension name. "coeffs" is assumed by
```
NathanielF, line 138: Ah ok fair enough. I guess easier to be consistent across code base

causalpy/experiments/synthetic_control.py
```diff
@@ -41,8 +41,10 @@ class SyntheticControl(BaseExperiment):
         A pandas dataframe
     :param treatment_time:
         The time when treatment occurred, should be in reference to the data index
-    :param formula:
-        A statistical model formula
+    :param control_units:
+        A list of control units to be used in the experiment
+    :param treated_units:
+        A list of treated units to be used in the experiment
```
NathanielF, line 47: Ok, sorry. i forgot this was synthetic control. So treated as target makes sense. The idea would be to enable multiple treatment groups? Do you want to enable different "equations" for the different treatment units? Or would it be just the same fixed controls for each potential treatment unit?

causalpy/pymc_models.py
```diff
@@ -270,12 +285,7 @@ def build_model(self, X, y, coords):
             n_predictors = X.shape[1]
             X = pm.Data("X", X, dims=["obs_ind", "coeffs"])
             y = pm.Data("y", y[:, 0], dims="obs_ind")
-            # TODO: There we should allow user-specified priors here
             beta = pm.Dirichlet("beta", a=np.ones(n_predictors), dims="coeffs")
```
NathanielF, line 288: So this is the component that would be updated with multiple treatment units? Allowing a matrix of n_predictors x n_targets?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -107,6 +110,33 @@ def __init__(
         )
         self.post_X = np.asarray(new_x)
         self.post_y = np.asarray(new_y)
+        # turn into xarray.DataArray's
+        self.pre_X = xr.DataArray(
```
NathanielF, line 114: Seems like a pretty extensive API change for multiple different methods. Should we plan for removing patsy altogether? I do like formulas... but consistency is kind of important overall....

NathanielF (commented):
This was a quick pass but looks broadly sensible. Just trying to understand how extensive you want to make the API change? Asked some questions but mostly clarifying for my own understanding. 

I can see how the API change makes it easier to allow for a matrix of weights in the WeightedFitter. 

My one questions was whether you wanted to allow multiple "equations" for the different treatment groups i.e. generate different synthetic controls based on a varying set of inputs. Because If so maybe patsy could still work.... but you'd have to more flexible in the PyMC model than in the API ....

causalpy/pymc_models.py
drbenvincent: Yes. Plan was to do things stepwise, so this PR is about API which allows for the next PR which allows multiple treated units. Just to keep things bite-sized etc.

drbenvincent:
Thanks for the review.

> I can see how the API change makes it easier to allow for a matrix of weights in the WeightedFitter.
> 
> My one questions was whether you wanted to allow multiple "equations" for the different treatment groups i.e. generate different synthetic controls based on a varying set of inputs. Because If so maybe patsy could still work.... but you'd have to more flexible in the PyMC model than in the API ....

My original idea was to see if patsy allowed something like `treated1, treated 2, treated 3 ~ control1 + ... + controlN`, but it doesn't do that.

The rationale for going for a list of treated and a list of control units was to a) allow multiple treated units, b) allow the weights in the pymc model to change from a 1D vector to a 2D array. Crucially, if all treated units are predicted by the _same_ set of control units then this would be a relatively trivial change to the shape of the Dirichlet distributed weights.

I have to admit, I didn't think of your idea of providing a list of formulas. This is interesting:
* ✅ Retains the same basic formula interface
* ✅ The user gets extra ability to add domain knowledge into the modelling process. They could have prior knowledge that some control outlets make sense to go into the pool for some treated units but not others. You could just let the data decide, but having that ability (to effectively set some coefficients at zero by not including a control unit as a predictor) to provide domain knowledge into the modelling is appealing.
* ❌ It would be slightly harder to implement in the `WeightedSumFitter` class. Probably not that hard though - you could probably just iterate over the treated units, building up the model with the right length 1-dimensional Dirichlet distributions as you go.

causalpy/experiments/interrupted_time_series.py
drbenvincent: I think keeping formulas as the main API is a good idea. See longer [reply](https://github.com/pymc-labs/CausalPy/pull/460#issuecomment-2865983360) below.

NathanielF:
Yes, i think you have the suggestion right. I think it potentially gives you more flexibility. I'm trying to do the same here: https://github.com/pymc-labs/pymc-marketing/pull/1654

But it does kind of move the complexity into the model class as you have to parse and construct the relevant components.

drbenvincent:
Hi @NathanielF. I'm just thinking more about this. I'm not necessarily averse to diving in with the list of formula approach, and there is something to be said for "just deliver the feature" and don't worry if the implementation gets a bit messy. But for whatever reason I seem to be taking the approach of [measure twice and cut once](https://en.wiktionary.org/wiki/measure_twice_and_cut_once) before I pull the trigger and implement. The code structure is semi close to elegant (or at least simple) at this point, so I'm just thinking it through.

## If we change the API to allow lists of formulas

### API
The `formula` kwarg is now a list of strings (formulas) rather than a single formula string. This allows users control over what control groups are treated as 'available' in the synthetic control donor pool.

### `SyntheticControl` class

* Still uses `patsy` to generate $X$ and $y$
* Though now we will have pairs of $X, y$ for each treated unit
* Because users can provide different numbers and/or identities of control units, we can't just put this into a 3D matrix shape `observations, control_units, treatment_units`.

### `WeightedSumFitter` class
* This poses a slight challenge. At first glance you can think to simply iterate over each treated unit, pulling the $X, y$ data for each treated unit.
* However, the whole `PyMCModel` class is build around a scikit-learn style passing around of predictor matrices $X$ and outcomes $y$. So this presents a challenge. Though it might not be that problematic - for example the `InstrumentalVariableRegression` class has additional inputs to the `build_model` method, and this seems to be dealt with by simply overriding the `fit` method. I'm not 100% about this yet, but I think it will likely necessitate overriding more methods of the `PyMCModel` class.
* Potentially the simplest way to do it would be to pass in _lists_ of $X$ and $y$. The method signatures could remain the same, but the methods would have to be specialised (or dispatch) to deal with either ndarray or lists of ndarrays.

## If we stick with list of treated and control units

As the PR currently stands we have lists of treated and control units. This assumes that the same set of control units are used in the synthetic control donor pool for all treated units.

❌ Less flexibility for the user
❌ A deviation in the API away from the formula approach
✅ Could allow for $X$ and $y$ to simply gain another dimension (i.e. `treated_units`) which _will hopefully_ allow for the method signatures to stay the same, not have to override methods of `PyMCModel`, assuming that xarray can handle all the broadcasting properly.

drbenvincent:
Another potential advantage of having the same set of control units for all target units is that we can have a 2D matrix of weights (the number of controls by the number of treated units). I'm not that experienced with the Dirichlet distribution, but perhaps this allows for partial pooling of weights across treated units. That might be a nice selling point?

drbenvincent:
Just to make sure connections are made - as part of #463 we are talking about potentially swapping `patsy` for `formulaic`. As far as I can tell this allows multiple terms on the left hand side of the `~` which _could_ be useful if we go down the formula route here.

drbenvincent:
I'm thinking that we can end up with both API's. Though I propose moving forward with the list of treated and control unit API's just to keep the momentum going, and slightly because the implementation will be easier. Then we can follow up in the future with the list of formulas API. What do you think @NathanielF?

NathanielF:
That makes sense @drbenvincent i definitely didn't want suggest this was blocking. More just food for thought. I agree with the current proposal to move forward

causalpy/experiments/diff_in_diff.py
drbenvincent: My motivation for calling it `coeffs` was the the use case of extracting the inferred coefficient values. I agree it does make more sense to label these as covariates when you are building the model and its dims. I might put together an issue to describe this - it could be a good refactor, but I'd say it's best dealt with separately.

NathanielF (approved):
Looks great! Congrats.

PR #462, opened 2025-04-21
fix bug: obs_indx -> obs_ind and improve `PyMCModel._data_setter`

### LINKED ISSUE #459: Convert all instances of `obs_indx` to `obs_ind`

drbenvincent:
* Closes #459
* global find and replace obs_indx -> obs_ind
* Also had to update `PyMCModel._data_setter` to make it work
* All tests + doctests work locally, so the remote tests should also work 🤞🏻

----
📚 Documentation preview 📚: https://causalpy--462.org.readthedocs.build/en/462/

### DIFF
--- a/causalpy/pymc_models.py
+++ b/causalpy/pymc_models.py
@@ -87,9 +87,20 @@ class PyMCModel(pm.Model):
 
         This method is used internally to register new data for the model for
         prediction.
+
+        NOTE: We are actively changing the `X`. Often, this matrix will have a different
+        number of rows than the original data. So to make the shapes work, we need to
+        update all data nodes in the model to have the correct shape. The values are not
+        used, so we set them to 0. In our case, we just have data nodes X and y, but if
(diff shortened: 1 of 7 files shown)


causalpy/pymc_models.py
```diff
@@ -111,7 +122,7 @@ def fit(self, X, y, coords: Optional[Dict[str, Any]] = None) -> None:
             )
         return self.idata
 
-    def predict(self, X):
+    def predict(self, X: np.ndarray):
```
NathanielF, line 125: Is the type hint needed or always necessary?

NathanielF (approved):
Generally looks good. Left one question

PR #473, opened 2025-05-24
Implement Bayesian Structural Time Series (BSTS)

cetagostini:
**PR Description:**

This PR introduces Bayesian Structural Time Series (BSTS) modeling to CausalPy, enabling advanced time series analysis, forecasting, and causal impact estimation. Inspired by principles from structural time series modeling basic with only trend/seasonal components (e.g., [TensorFlow Probability's STS first example](https://blog.tensorflow.org/2019/03/structural-time-series-modeling-in.html)), this work decomposes time series into trend, seasonality, and optional regressor components within a Bayesian framework.

**Key Changes:**

1.  **`BayesianStructuralTimeSeries` Model (`causalpy/pymc_models.py`):**
    *   New PyMC model for BSTS. Defaults to `pymc-marketing` components (`LinearTrend`, `YearlyFourier`), supports custom components, and optional exogenous regressors (`X`).
    *   Adapted `_data_setter`, `predict`, `score`, and `fit` for time-dependent forecasting and ensuring `mu` (sum of components) is sampled.

2.  **`StructuredTimeSeries` Experiment (`causalpy/experiments/structured_time_series.py`):**
    *   New experiment class for causal inference with BSTS.
    *   Handles `DatetimeIndex` data, `treatment_time`, and `patsy` formulas for exogenous regressors.
        *   **Intercept Logic:** Correctly manages formulas like `y ~ 0` or `y ~ 1` by ensuring the BSTS model's trend component provides the baseline, avoiding redundant patsy-generated intercepts for exogenous regressors.
    *   Provides `summary()`, `plot()` (3-panel: fit/counterfactual, pointwise impact, cumulative impact), and `get_plot_data()`.

3.  **Testing & Integration:**
    *   Added comprehensive integration tests (`test_bayesian_structural_time_series`) covering various scenarios, including custom components and error handling.
    *   Fixed a plotting issue in `plot_xY` (`causalpy/plot_utils.py`).

4.  **API & Docs:**
    *   `StructuredTimeSeries` available as `causalpy.StructuredTimeSeries`.
    *   Added backward-compatible wrapper in `causalpy.pymc_experiments.py`.
    *   Updated relevant docstrings.


cetagostini:
Thinking to [bring HSGP to here](https://www.pymc-marketing.io/en/stable/api/generated/pymc_marketing.mmm.hsgp.HSGP.html#pymc_marketing.mmm.hsgp.HSGP), and be able to replicate the second example.

cetagostini:
Hey @drbenvincent 

Still BSTS, the Structural time series (STS) models are a family of probability models for time series that includes and generalizes many standard time-series modeling ideas, including:
- autoregressive processes,
- moving averages,
- local linear trends,
- seasonality, and regression and variable selection on external covariates (other time series potentially related to the series of interest).

Definitely, it's handy and popular to talk about state-space, but my understanding is that BSTS in the broad sense does not require a true state‐space; it merely requires a Bayesian, additive decomposition into trend, seasonality, and optional regressors. Unless you insist on the classic DLM/state‐space definition from a Kalman‐Filter sense.

In the Google blog the first model its very similar to this, the only difference is that the slope from the linear trend is "_evolving slowly over time following a random walk_". 

Thanks to the HSGP class, we can add something similar if we want to.

cetagostini:
> Not able to take a deep dive at this point, but my hope in terms of the API would be that we could use the existing InterruptedTimeSeries experiment class and simply inject a new time series based pymc model class. Will aim to look into this in more detail - if there was a reason for a new experiment class then we can look into whether we can adapt it to make it more general.

Happy to do it, my only opinion is that the API for the new class is quite different and allow for arbitrary components for trend and seasonality. IF they follow pymc-marketing protocols.

I'm doing that because will help me to easily input HSGP and replace this fourier bases approach (just as example), for the same structural time series. Could probably help to just replace by the state-space.

I'll be working on that as well :) I feel should be easy to do.

cetagostini:
> The new notebook is a great start. Could be good to compare the previous approach "y ~ 1 + t + C(month)" in its_pymc.ipynb to see how things differ.

Almost the same. The bsts fits more during training (high R2) and has less variance (lower sd).

cetagostini:
> Nice. So I am a beginner in time series modelling - which is why I've not given this a go yet. But that's a good clarification BSTS != state space. 
> 
> Do you think your implementation leaves the door open for extension into autoregressive components etc?

Indeed, I'll try to bring an example tomorrow!

drbenvincent (changes requested):
Firstly, this is very cool! Will probably need a few rounds of review to get this merged, but it _will_ happen :)

Let's try to make it fit into the current docs and codebase as seamlessly as possible. So far we've got a nice distinction between "experiments" and "models". 

The new _model_ class `BayesianStructuralTimeSeries` fits in with the current structure very nicely. Given the range of experiments we've got so far the obvious candidate is to work with the interrupted time series _experiment_, and _maybe_ multi-observation differences in differences. So the natural way to make this work would be to make the `InterruptedTimeSeries` class accept your new `BayesianStructuralTimeSeries` class. That would be in an ideal world, but let's see what we can do quickly to move in that direction.

The API differences between `InterruptedTimeSeries` and `StructuredTimeSeries` are minimal. Although obviously what is done with the provided data diverges a lot. Trying to resolve this in one go might be a bit much, but at the moment I've got these concrete suggestions:
* Move the new `StructuredTimeSeries` class into `interrupted_time_series.py` and delete `structured_time_series.py`.
* I relatively strongly think we should avoid the duplication of the plotting code. Like I say, we're trying to make the interrupted time series _experiment_ work with a BSTS _model_, I don't think there is any reason why any custom plot logic is needed. We could extract the plot code into a mixin class (called something like `ITSPlotter`) which can be injected into both `StructuredTimeSeries` and `InterruptedTimeSeries`.

I think this already goes a long way towards keeping the nice distinction between experiment classes and model classes. The PR is already not far off that - the main issue I see is that `StructuredTimeSeries` is currently treated like a new experiment when in fact it's exactly the same as interrupted time series, but just separate in order to handle the different model input.

The new notebook is clearly applied to the interrupted time series experiment, so I think the notebook contents can be appended to the existing `its_pymc.ipynb`. That could be quite nice - the notebook would then be a kind of "how to do ITS from simple linear model to proper time series modeling".

AlexAndorra:
This looks really cool and useful!!
Just dropping in as I see you guys are thinking of making this broader, with HSGP and broader TS decomposition (which I love!): any reason you're not plugging into `pymc-extras.statespace`? It should already have everything you need, and you can offload the model dev to it

drbenvincent:
> any reason you're not plugging into `pymc-extras.statespace`

Thanks @AlexAndorra - this is the direction I was hoping we'd go in. I've done some basic experimentation with it, but I'm a time series newbie and am not confident enough to know when I need to use what component, how to deal with non-stationarity etc. There's a lot more lower hanging fruit for me to work on in CausalPy, so I'm happy that this has been dropped on my lap by @cetagostini :) 

I have a _very_ crude understanding of how BSTS and state space approaches relate to each other (though @cetagostini explained it a bit [above](https://github.com/pymc-labs/CausalPy/pull/473#issuecomment-2906970474)).

Would it make sense to end up having _both_ this kind of BSTS model class and (eventually) the `pymc-extras` state space stuff? If that is silly then I guess the proposed implementation could be over-written by a state space implementation? Very happy to take guidance on this.

drbenvincent:
@cetagostini The `BayesianStructuralTimeSeries.build_model` method seems like a _reasonably_ simple bit of pymc code. Though as far as I can tell, this seems to be build as an out of the box solution. Nothing wrong with that, but the [TFP blog post](https://blog.tensorflow.org/2019/03/structural-time-series-modeling-in.html) you linked has a _very_ modular API where users can build their own model from components.

I don't think having a fixed out of the box BSTS model is bad, and in some ways it is good because it is what it is and doesn't require a user to meddle much. That said, it could be pretty cool to think about an API where users can modularly build up a pymc model to pass into the experiment class. Though I imagine this would probably be a considerable bit more work?

drbenvincent:
Sure @cetagostini I'm happy with an iterative approach. Happy to proceed along the lines in my first review.

cetagostini:
@drbenvincent 

Apply the changes:
- Under experiements, change Interrupted Time Series name to Structural Time Series, more generic and adhoc to get Interrupted time and basis expansion time series. This will help to hold the same structural time series class able to get StateSpaceTimeSeries, BasisExpansionTimeSeries or InterruptedTimeSeries. All under same family.
- Keep interrupted time series signatures to backward compatibility.
- Move the example under the same ITS notebook.
- Avoid duplication of plotting and stuff because we rely on the same class.

cetagostini:
@drbenvincent Implementing the StateSpace it's taking more effort than expected, will continue during week... But I think, it requires  probably a following PR to not make this massive, instead of all in one.. Not sure, take a look and give me feedback. I already add a class draft in the notebook but need more work to be able to be used by the new experiment STS class.

drbenvincent:
I'll take a look at the conflicts when I'm back from a work trip. 

Just wanted to keep up the momentum but I know you've got other demands on your time.

drbenvincent:
@cetagostini I've not resolved all the conflicts. There were some major changes that got merged into main.

Note, the way how we build docs has now slightly changed. It should be identical to how you do it in `pymc-marketing` now, but see [here](https://github.com/pymc-labs/CausalPy/blob/main/CONTRIBUTING.md#building-the-documentation-locally) if in doubt

jessegrabowski:
> The issue with state space it's observe even in the examples in BSTS from pycm-extras.

I talked to Carlos about this in private, but just so that it's repeated here: there's no bug in extras, only in that plotting code.

drbenvincent:
Anything needed from me at this point? Even if it's just to coordinate and provide extra motivational pizazz?

cetagostini:
Would be to much to ask to keep by now? `cp.StructuralTimeSeries` and `cp.InterruptedTimeSeries` for a bit? I made both to be equivalent, after a few PRs, interrupted change quite a bit and they are not similar any more, and looks like a bit of work to go over all changes to release.

I feel better to keep this separate, merge, and make another just to kill StructuralTimeSeries (Try to tackle all in this will make it as stuck as power analysis at some point). Probably, would be nice, if I can get a grasp of the recent changes.

I applied, @jessegrabowski explanation and everything looks nicer on the plot now :)

drbenvincent:
Doc test failing for `StructuralTimeSeries`. Currently that docstring uses the `LinearRegression` model, which I guess is not a doable experiment/model combination? Do just need to replace with the right model I assume?

causalpy/pymc_models.py
```diff
@@ -749,3 +759,779 @@ class initialisation.
...
+
+    This model allows for the inclusion of trend, seasonality (via Fourier series),
+    and optional exogenous regressors.
+
+    .. math::
+        \text{trend} &\sim \text{LinearTrend}(...) \\
+        \text{seasonality} &\sim \text{YearlyFourier}(...) \\
+        \beta &\sim \mathrm{Normal}(0, \sigma_{\beta}) \quad \text{(if X is provided)} \\
+        \sigma &\sim \mathrm{HalfNormal}(\sigma_{err}) \\
+        \mu &= \text{trend_component} + \text{seasonality_component} [+ X \cdot \beta] \\
```
drbenvincent: This line in the docstring is confusing. That `+` in the square brackets.

drbenvincent (changes requested):
* Could add `pymc-extras` to the `autodoc_mock_imports` in `conf.py`. Should help with autodoc when we refer to the time series stuff in `pymc-extras`. But consider as optional - can come back to polish this in another small and focussed PR.
* Mentioned elsewhere that it would be super cool to get more intro information in the docs about time series modelling. Maybe it would be a lower bar to just expand on this in the docstrings of the new classes?

NathanielF:
Don't have complete familiarity with this PR, but just on the dependency question. For the IV experiment class, we recommended posterior predictive sampling with numpyro rather than the baseline pymc because MvNormal ppc was so slow. 

I think recommending a sampler as an optional dependency seems more reasonable than a whole package.... 

Not saying we shouldn't but pymc-marketing is huge. Would avoid inflating dependencies if we could avoid it...

drbenvincent:
## On the desired merge `StructuralTimeSeries` into `InterruptedTimeSeries`

Bear in mind that since this PR was started all experiment classes (including `InterruptedTimeSeries`) has embraced using xarray data structures, rather than numpy. So at the very least, we need to embrace that change. Not looked in detail, but that might also eliminate some of the difference between the two classes? Could be much easier with some LLM help on that front. I can give it a go if you're not keen @cetagostini?

Looking at the diff between the two, I think it's doable. Most of the differences related to the numpy -> xarray change.

cetagostini:
@drbenvincent and @NathanielF I did `pymc-extras` and `pymc-marketing` optional dependencies with lazy imports. On top I merge `ITS` with the old `STS` and then delete the `STS`.

I move the notebook to a new draft folder, only to keep it, but we need to work in the notebooks in the future. Maybe another PR?

cetagostini-wise:
@drbenvincent @NathanielF any feedback?

NathanielF:
Sorry @cetagostini will try to review this evening. Been a crazy few weeks.

causalpy/tests/conftest.py
```diff
@@ -30,14 +39,23 @@ def rng() -> np.random.Generator:
...
+        # No-op fixture to satisfy tests when PyMC testing helpers are unavailable
+        yield
 
 
 @pytest.fixture(autouse=True)
 def mock_sample_for_doctest(request):
     if not request.config.getoption("--doctest-modules", default=False):
         return
 
+    if not _HAVE_PYMC_TESTING or mock_sample is None:
```
NathanielF, line 57: This is good to have.

docs/source/notebooks/drafts/its_pymc copy.ipynb
NathanielF, line 1: rename the file i guess. 

Notebook looks good, re: three examples. 

Maybe do some more explanatory writing around this comment 
# Exogenous regressors are optional

If that's the framing, isn't it cool that we can absorb some of this variation into really compelling latent components? Be nice to motivate it that way...maybe?

docs/source/notebooks/its_pymc.ipynb
NathanielF, line 1: Nice., Yeah, move the notebooks out of draft. Add to the How to sections and maybe add a little story telling of how adding structure is a good example of bayesian workflow. Y

causalpy/pymc_models.py
```diff
@@ -749,3 +770,773 @@ class initialisation.
...
+        \sigma &\sim \mathrm{HalfNormal}(\sigma_{err}) \\
+        \mu &= \text{trend_component} + \text{seasonality_component} + X \cdot \beta \quad \text{(if X is provided)} \\
+        y &\sim \mathrm{Normal}(\mu, \sigma)
+
+    Parameters
+    ----------
+    n_order : int, optional
+        The number of Fourier components for the yearly seasonality. Defaults to 3.
+        Only used if seasonality_component is None.
+    n_changepoints_trend : int, optional
```
NathanielF, line 1080: Non-blocking but i think after we merge this and @JeanVanDyk 's intervention detection module we should think about change points more holistically... to see if there are any options for consolidation/streamlining

causalpy/pymc_models.py
```diff
@@ -749,3 +770,773 @@ class initialisation.
...
+        if self._custom_trend_component is not None:
+            return self._custom_trend_component
+
+        # Create default trend component (lazy import of pymc-marketing)
+        if self._trend_component is None:
+            try:
+                from pymc_marketing.mmm import LinearTrend
+            except ImportError as err:
+                raise ImportError(
+                    "BayesianBasisExpansionTimeSeries requires pymc-marketing when default trend "
```
NathanielF, line 1163: Cool. This is a nice way of doing it.

causalpy/pymc_models.py
```diff
@@ -749,3 +770,773 @@ class initialisation.
...
+                time_for_trend,
+                dims="obs_ind",
+            )
+            t_season_data = pm.Data(
+                "t_season_data",
+                time_for_seasonality,
+                dims="obs_ind",
+            )
+
+            # Get validated components (no more ugly imports in build_model!)
```
NathanielF, line 1373: Nice.

causalpy/pymc_models.py
```diff
@@ -749,3 +770,773 @@ class initialisation.
...
+                    raise ValueError(
+                        f"Shape mismatch: X_values_for_pymc has {X_values_for_pymc.shape[1]} columns, but "
+                        f"{len(self._exog_var_names)} names in self._exog_var_names ({self._exog_var_names})."
+                    )
+                X_data = pm.Data("X", X_values_for_pymc, dims=["obs_ind", "coeffs"])
+                beta = pm.Normal("beta", mu=0, sigma=10, dims="coeffs")
+                mu_ = mu_ + pm.math.dot(X_data, beta)
+
+            # Make mu_ an explicit deterministic variable named "mu"
+            mu = pm.Deterministic("mu", mu_, dims="obs_ind")
```
NathanielF, line 1422: Cool. This is neat.

causalpy/pymc_models.py
```diff
@@ -749,3 +770,773 @@ class initialisation.
...
+        initial_trend_dims, sigma_trend_dims, annual_dims, P0_dims = (
+            self.ss_mod.param_dims.values()
+        )
+        coordinates = {**coords, **self.ss_mod.coords}
+
+        # Build model
+        with pm.Model(coords=coordinates) as self.second_model:
+            # Add coords for statespace (includes 'time' and 'state' dims)
+            P0_diag = pm.Gamma("P0_diag", alpha=2, beta=1, dims=P0_dims[0])
+            _P0 = pm.Deterministic("P0", pt.diag(P0_diag), dims=P0_dims)
```
NathanielF, line 1741: Just out of curiosity, can these priors be changed easily, like would the use the prior class allow us to swap them in and out e.g. sigma = 80 in ZeroSumNormal... is there a reason we want fixed at that magnitude?

causalpy/pymc_models.py
```diff
@@ -749,3 +770,773 @@ class initialisation.
...
+        mu_pred = az.extract(
+            pred_output, group="posterior_predictive", var_names="mu"
+        ).T.values
+        # Note: First argument must be a 1D array
+        return r2_score(y.flatten(), mu_pred)
+
+
+class StateSpaceTimeSeries(PyMCModel):
+    """
+    State-space time series model using pymc_extras.statespace.structural.
```
NathanielF: Maybe link to some @jessegrabowski 's more developed notebooks or docs for the interested reader to find out more about statespace generally

causalpy/pymc_models.py
cetagostini: No, I'll make an issue and get familiar how we are allowing Prior class to allow to modify!

causalpy/pymc_models.py
cetagostini: Done!

causalpy/experiments/interrupted_time_series.py
```diff
@@ -273,30 +367,65 @@ def _bayesian_plot(
...
+        r2_std_val = None
+        try:
+            if isinstance(self.score, pd.Series):
+                if "unit_0_r2" in self.score.index:
+                    r2_val = self.score["unit_0_r2"]
+                    r2_std_val = self.score.get("unit_0_r2_std", None)
+                elif "r2" in self.score.index:
+                    r2_val = self.score["r2"]
+                    r2_std_val = self.score.get("r2_std", None)
+        except Exception:
```
drbenvincent, line 395: Can you be more specific about the exception type

causalpy/tests/conftest.py
```diff
@@ -20,7 +20,16 @@
...
 import numpy as np
 import pytest
-from pymc.testing import mock_sample, mock_sample_setup_and_teardown
+
+# Try to use PyMC's testing helpers if available; otherwise, fall back to no-op fixtures
+try:  # pragma: no cover - conditional import for compatibility across PyMC versions
+    from pymc.testing import mock_sample, mock_sample_setup_and_teardown  # type: ignore
+
+    _HAVE_PYMC_TESTING = True
+except Exception:  # pragma: no cover
```
drbenvincent, line 29: Can you be more specific about the exception type

PR #480, opened 2025-05-28
Draft (new feature) : Model to estimate when a intervention had effect

JeanVanDyk:
## New Feature: `InterventionTimeEstimator` for Unknown Treatment Timing

This PR introduces a new model, `InterventionTimeEstimator`, designed to estimate when an intervention has an effect in a time series — especially in cases where the exact treatment time is unknown.

---

### Use Case

- Enhances the **Interrupted Time Series (ITS)** feature by providing a way to _infer the likely time of intervention_
- Supports scenarios where the _treatment onset is uncertain or delayed_
- Helps identify _lagged effects_ between intervention and observable outcomes

This addition gives users a flexible, **Bayesian approach** to model treatment timing uncertainty directly within the **CausalPy** framework.

---

### Notes / Open Questions

- **Where should this model fit into the CausalPy workflow?**  
  I’m unsure whether `InterventionTimeEstimator` should be integrated **within the `InterruptedTimeSeries` (ITS) feature**, or used as a **standalone tool**.  
  This affects how a **user-defined model** could be supported.  
  Depending on the intended usage, I can propose a solution to allow users to inject their own custom models.

- **Custom model usage — base vs. intervention**  
  Should users be able to:
  - Provide a custom model to represent the **base time series** (e.g. intercept, trend, seasonality)?
  - Provide a custom model to capture the **intervention effect** (e.g. shape or dynamics of the post-switch impact)?
  - Or support both?
 
- **Covariates**  
  I considered adding time-varying covariates to improve the fit. Would that be useful or out of scope?

- **Multivariate Time Series**  
  It's relatively easy to extend the model for multivariate input. Let me know if this is something you'd like to see.

---

### Model Summary

- **Inputs**:
  - `t`: 1D array of time points
  - `y`: 1D array of observed values
  - Optional `span`: restricts the window for switchpoint detection
  - Optional `coords`: can include `seasons` for modeling periodic effects
  - `effect`: list of components, e.g. `"trend"`, `"level"`, `"impulse"`
  - `grain_season`: number of time steps per season

- **Model Components**:
  - Time series is modeled as:  
    _intercept + trend + seasonal_
  - A **Uniform prior** is used to place a switchpoint
  - A **sigmoid curve** models the onset of the effect after the switchpoint, applied to the selected `effect` components

---

Feel free to share any feedback or suggestions! I'm happy to refine the model or explore extensions based on your input.

----
📚 Documentation preview 📚: https://causalpy--480.org.readthedocs.build/en/480/


drbenvincent (changes requested):
Hi @JeanVanDyk. I'm very excited about the new addition, thanks for putting it together. Bear with me if I'm sometimes slow to reply - busy dad life!

At this point, could I request that you put together an `ipynb` file that showcases the functionality. I'm assuming this PR will evolve and go through a couple of iterations, so this notebook will end up being a new docs page to help users understand the new functionality. If you edit the `./docs/source/notebooks/index.md` file, you can add the notebook name under the interrupted time series section. That way the notebook will render if you either build the docs locally, but it should also render in the remote docs preview build.

Did you still want feedback on the GitHub Discussion at this point?

drbenvincent:
@JeanVanDyk I just did a quick update from `main`. Please make sure you pull the latest to avoid any conflicts :)

Hoping to start taking a decent look at this now. Will try to provide some comments even though I know this is not 100% finished.

docs/source/notebooks/index.md
```diff
@@ -40,6 +40,7 @@ did_pymc_banks.ipynb
 its_skl.ipynb
 its_pymc.ipynb
 its_covid.ipynb
+its_no_treatment_time.ipynb
```
drbenvincent, line 43: This is good. But in the notebook can you ensure you only have ONE top level markdown header. Otherwise it adds these as additional entries into the How To index page.

drbenvincent (changes requested):
* Can you run `make uml`? This will generate updated uml plots (e.g. `classes.png`)
* Looking at that generated class diagram (and not yet having looked at the `InterventionTimeEstimator` the hope would be to avoid overwriting the methods `calculate_impact`, `predict`, `score` and `fit` methods. These are hopefully very generic and should hopefully be able to be dealt with in the `PyMCModel` base class. Like I say, I've not yet taken a look at the code itself, but this would be the hope.
* I cannot currently run the new notebook. When the first model fit is attempted I get this

<details>
<summary>Traceback</summary>
TypingError                               Traceback (most recent call last)
Cell In[4], [line 10](vscode-notebook-cell:?execution_count=4&line=10)
      2 from causalpy.pymc_models import InterventionTimeEstimator as ITE
      4 model = ITE(
      5     time_variable_name="t",
      6     treatment_type_effect={"level": []},
      7     sample_kwargs={"sample_seed": seed},
      8 )
---> [10](vscode-notebook-cell:?execution_count=4&line=10) result = ITS(
     11     data=df,
     12     treatment_time=None,
     13     formula="y ~ 1 + t",
     14     model=model,
     15 )

File ~/git/CausalPy/causalpy/experiments/interrupted_time_series.py:284, in InterruptedTimeSeries.__init__(self, data, treatment_time, formula, model, **kwargs)
    281     raise ValueError("Model type not recognized")
    283 # score the goodness of fit to the pre-intervention data
--> [284](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/git/CausalPy/causalpy/experiments/interrupted_time_series.py:284) self.score = self.model.score(X=self.pre_X, y=self.pre_y)
    286 # Postprocessing with handler
    287 self.datapre, self.datapost, self.pre_y, self.pre_X, self.treatment_time = (
    288     self.handler.data_postprocessing(
    289         self.model, data, idata, treatment_time, self.pre_y, self.pre_X
    290     )
    291 )

File ~/git/CausalPy/causalpy/pymc_models.py:773, in InterventionTimeEstimator.score(self, X, y)
    771 mu_ts = az.extract(mu_ts, group="posterior_predictive", var_names="mu_ts").T
    772 # Note: First argument must be a 1D array
--> [773](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/git/CausalPy/causalpy/pymc_models.py:773) return r2_score(y.data, mu_ts)

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats.py:1167, in r2_score(y_true, y_pred)
   1134 def r2_score(y_true, y_pred):
   1135     """R² for Bayesian regression models. Only valid for linear models.
   1136 
   1137     Parameters
   (...)   1165 
   1166     """
-> [1167](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats.py:1167)     r_squared = r2_samples(y_true=y_true, y_pred=y_pred)
   1168     return pd.Series([np.mean(r_squared), np.std(r_squared)], index=["r2", "r2_std"])

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats.py:1127, in r2_samples(y_true, y_pred)
   1125     var_e = _numba_var(svar, np.var, (y_true - y_pred))
   1126 else:
-> [1127](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats.py:1127)     var_y_est = _numba_var(svar, np.var, y_pred, axis=1)
   1128     var_e = _numba_var(svar, np.var, (y_true - y_pred), axis=1)
   1129 r_squared = var_y_est / (var_y_est + var_e)

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/utils.py:372, in _numba_var(numba_function, standard_numpy_func, data, axis, ddof)
    350 """Replace the numpy methods used to calculate variance.
    351 
    352 Parameters
   (...)    369 
    370 """
    371 if Numba.numba_flag:
--> [372](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/utils.py:372)     return numba_function(data, axis=axis, ddof=ddof)
    373 else:
    374     return standard_numpy_func(data, axis=axis, ddof=ddof)

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats_utils.py:535, in stats_variance_2d(data, ddof, axis)
    533     var = np.zeros(a_a)
    534     for i in range(a_a):
--> [535](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats_utils.py:535)         var[i] = stats_variance_1d(data[i], ddof=ddof)
    536 else:
    537     var = np.zeros(b_b)

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/utils.py:205, in maybe_numba_fn.__call__(self, *args, **kwargs)
    203 """Call the jitted function or normal, depending on flag."""
    204 if Numba.numba_flag:
--> [205](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/utils.py:205)     return self.numba_fn(*args, **kwargs)
    206 else:
    207     return self.function(*args, **kwargs)

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/numba/core/dispatcher.py:424, in _DispatcherBase._compile_for_args(self, *args, **kws)
    420         msg = (f"{str(e).rstrip()} \n\nThis error may have been caused "
    421                f"by the following argument(s):\n{args_str}\n")
    422         e.patch_message(msg)
--> [424](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/numba/core/dispatcher.py:424)     error_rewrite(e, 'typing')
    425 except errors.UnsupportedError as e:
    426     # Something unsupported is present in the user code, add help info
    427     error_rewrite(e, 'unsupported_error')

File ~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/numba/core/dispatcher.py:365, in _DispatcherBase._compile_for_args.<locals>.error_rewrite(e, issue_type)
    363     raise e
    364 else:
--> [365](https://file+.vscode-resource.vscode-cdn.net/Users/benjamv/git/CausalPy/docs/source/notebooks/~/mambaforge/envs/CausalPy/lib/python3.13/site-packages/numba/core/dispatcher.py:365)     raise e.with_traceback(None)

TypingError: Failed in nopython mode pipeline (step: nopython frontend)
non-precise type pyobject
During: typing of argument at /Users/benjamv/mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats_utils.py (517)

File "../../../../../mambaforge/envs/CausalPy/lib/python3.13/site-packages/arviz/stats/stats_utils.py", line 517:
    def copy(self, deep=True):  # pylint:disable=overridden-final-method
        <source elided>

@conditional_jit(nopython=True)
^

During: Pass nopython_type_inference 

This error may have been caused by the following argument(s):
- argument 0: Cannot determine Numba type of <class 'xarray.core.dataarray.DataArray'>    
</details>

* I think there should be no changes to `its_skl.ipynb` right? Can we revert those changes back to the state on main, so there's no change in that file in this PR.
* Same for `its_covid.ipynb`?

docs/source/notebooks/index.md
JeanVanDyk: Okay, sure. I'll fix that!

As for the two notebooks you mentioned, I don't recall making any changes to them. I only ran them to check whether the modifications contained any small mistakes. I'll try reverting those changes.

drbenvincent:
Could you send me a screenshot of the graphviz? At the moment I can't get the `result` because of the error mentioned above. But if it's working for you, then you can do this. It would be useful for me to get the birds-eye-view of the model as I'm reading the code.

Screenshot 2025-06-20 at 14 27 34

drbenvincent (changes requested):
## Handlers
* Can we rename to make more clear. E.g. `UnknownTreatmentTimeHandler` and `KnownTreatmentTimeHandler`
* We discussed a range of options to do this, and I still think the handler pattern is quite a good approach. If we did conditional logic in the main `InterruptedTimeSeries` class then it would be disconnected and unclear. Though I think we should formalise the interface with an abstract base class. Any major objections?

```python
from abc import ABC, abstractmethod

class TreatmentTimeHandler(ABC):
    @abstractmethod
    def data_preprocessing(self, data, treatment_time, model): pass
    
    @abstractmethod
    def data_postprocessing(self, ...): pass
    
    @abstractmethod
    def plot_intervention_line(self, ax, ...): pass
    
    @abstractmethod
    def plot_impact_cumulative(self, ax, ...): pass
    
    def plot_treated_counterfactual(self, ax, ...):
        """Optional: override if needed"""
        pass
```

## `treatment_type_effect` input

Minor gripes about this. The data structure is potentially a bit complex and allows user to add multiple keys (i.e. effects). What do you think about changing the API to something like this:

```python
model = ITE(
    time_variable_name="t",
    treatment_effect_type="level", # Required: one of "level", "trend", "impulse"
    treatment_effect_params=None, # Optional: custom parameters
    sample_kwargs={"sample_seed": seed},
)
```

## Notebook

* Rather than using the abbreviation `ITE`, can you just import as `InterventionTimeEstimator`? We want to maximise clarity for the reader and not require that they remember acronyms :)
* Sorry to drag it out, but I will have to leave a proper review of the notebook for another day. But we are definitely getting there. Maybe a couple more small iterations.

drbenvincent (changes requested):
## Architecture / API though

At the moment there is an API change to the norm in that for the first time we would be directly telling the model about the data when we generate it, by giving it `time_variable_name="t"`, like this:

```python
model = ITE(
    time_variable_name="t",
    treatment_type_effect={"impulse": []},
    sample_kwargs={"random_seed": seed, "target_accept": 0.95},
)
```

All the other models just consume `sample_kwargs` and don't override the `PyMCModel.__init__` method

https://github.com/pymc-labs/CausalPy/blob/8b26d427194081709b021d949f8af72a9394d202/causalpy/pymc_models.py#L71-L78

Your approach is to call the `PyMCModel.__init__` with `super().__init__(sample_kwargs)` and then add in the other arguments to the model self.

I don't yet know if this is deeply problematic. But it is a change in terms of information flow in that it's the first time we are giving the models information about the data while they are created.

My temptation would be to get rid of this and stick with the current pattern. 

An important point to think about here is that the `build_model` method is not actually called until we call `PyMCModel.fit`, so perhaps we can instead provide the `time_variable_name` and `treatment_type_effect` related stuff to the experiment and pass it to the model via `build_model`.

I'm tempted to try to revert to the current architecture. Something along the lines of

```python
class InterruptedTimeSeries:
    def fit(self):
        time_var = self._extract_time_variable_from_formula(self.formula)
        
        if isinstance(self.model, ITE):
            idata = self.model.fit(X, y, coords, time_variable=time_var)
        else:
            idata = self.model.fit(X, y, coords)
```

JeanVanDyk:
Hi @drbenvincent,

## Handlers

I made the changes you required. I also think it is a good thing to add an abstract class as it makes the code more scalable, more robust and easier to maintain. I thought about making it from the beginning, but talking with Andy, we figured that it might be an overkill and to just keep with two simpler class until we get your opinion on the matter. So no problem i've just committed that :)

I made the changes you requested. I also believe adding an abstract class is a good idea, as it makes the code more scalable, robust, and easier to maintain. I had considered implementing it from the start, but after discussing with Andy, we thought it might be overkill and decided to keep two simpler classes until we had your opinion on the matter. So no problem, I’ve just committed the update :)

## `treatment_type_effect` input

I agree that the current data structure may be complicated for users. If I understand correctly, you want to restrict users to specifying only one effect gievn as a string, with an optional parameter to indicate priors. If that is the case, we would lose the flexibility to support multiple simultaneous effects. However, perhaps this is intentional and you do not want that functionality? If not, I was thinking to let the users have these two options :
 
```
model = ITE(
    time_variable_name="t",
    treatment_effect_type="level", # Required: one of "level", "trend", "impulse"
    treatment_effect_params=None, # Optional: custom parameters
    sample_kwargs={"sample_seed": seed},
)
```

And having the possibility to do it also like this : 

```
model = ITE(
    time_variable_name="t",
    treatment_effect_type=["level", "impulse"], # Required: at least one of "level", "trend", "impulse"
    treatment_effect_params={"level":[50,10]}, # Optional: custom parameters
    sample_kwargs={"sample_seed": seed},
)
```

What do you think about this ?

## Architecture / API

I am not entirely sure how this line: `time_var = self._extract_time_variable_from_formula(self.formula)` would work in practice, given that we don’t want to impose constraints on how the user names the time variable or where it appears in the dataset. One possible solution might be to require the user to place the time variable as the last term in the formula, ensuring it is preceded by a space. Another option would be to add the time column explicitly as a parameter of InterruptedTimeSeries, but that would imply a significant API change.

The reason I initially needed the time variable was so the model could know at which point in time predictions are made. **Before implementing handlers**, I used to run predictions separately for pre- and post-intervention data. I couldn’t rely on the index as a time indicator because users might use datetime indices, and generating a new index starting at zero for both parts would cause incorrect sigmoid calculations.

**However, with the handlers now in place**, *we predict on the entire dataset at once*, then slice the results into pre- and post-intervention periods. So **I no longer think the model truly needs the actual time variable**. Since the data length is consistent between fit and predict, we could simply use something like: ``n = np.arange(len(data))`` and it should work.

The downside is that this approach would prevent, for example, fitting the model on one time period and predicting on a different one, since it would assume both correspond to the same timeline. But since the main intended use case is within InterruptedTimeSeries, I’m not sure this limitation matters much.

If it is enough that the model works well for this purpose, I think removing the time variable parameter would simplify things considerably. What do you think?

NathanielF:
Sorry folks. Been swamped and to be honest, completely forgot about this. Will review this week. Probably tonight. 

@JeanVanDyk , @drbenvincent

causalpy/custom_exceptions.py
```diff
@@ -37,3 +37,10 @@ class DataException(Exception):
 
     def __init__(self, message: str):
         self.message = message
+
+
+class ModelException(Exception):
+    """Exception raised given when there is some error in user-provided model"""
```
NathanielF, line 43: Some error? What's the value of the specific error class if it's quite unspecific?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -25,16 +26,279 @@
...
 from causalpy.plot_utils import get_hdi_to_df, plot_xY
 from causalpy.pymc_models import PyMCModel
 from causalpy.utils import round_num
 
-from .base import BaseExperiment
-
 LEGEND_FONT_SIZE = 12
 
 
+class TreatmentTimeHandler(ABC):
```
NathanielF: Should abstract classes sit in the ITS class i.e. will it be used for any other time-based causal inference methods?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -25,16 +26,279 @@
...
+    and visualization.
+    """
+
+    def data_preprocessing(self, data, treatment_time, model):
+        """
+        Preprocesses the input data by constraining the model's
+        treatment time inference window.
+        """
+        # Restrict model's treatment time inference to given range
+        model.set_time_range(treatment_time, data)
```
NathanielF: is treatment time not unknown in preprocessing stage? Does the input here warrant a better name?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -25,16 +26,279 @@
...
+        ]
+        timeline_broadcast = np.array(timeline)
+        tt_broadcast = tt_samples[:, :, None].astype(int)
+        mask = (timeline_broadcast >= tt_broadcast).astype(int)
+
+        # --- Compute cumulative post-treatment impact ---
+        post_impact = impact * mask
+        res["post_impact_cumulative"] = model.calculate_cumulative_impact(post_impact)
+
+        return res
```
NathanielF: This seems like a neat class. Nice work

causalpy/experiments/interrupted_time_series.py
```diff
@@ -79,37 +343,41 @@ class InterruptedTimeSeries(BaseExperiment):
...
-        # split data in to pre and post intervention
-        self.datapre = data[data.index < self.treatment_time]
-        self.datapost = data[data.index >= self.treatment_time]
 
+        self.treatment_time = treatment_time
         self.formula = formula
 
-        # set things up with pre-intervention data
+        # Getting the right handler
+        if treatment_time is None or isinstance(treatment_time, tuple):
```
NathanielF: Ok, nice. So the class is as before, but now we handle whether treatment is known or unknown?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -79,37 +343,41 @@ class InterruptedTimeSeries(BaseExperiment):
     def __init__(
         self,
         data: pd.DataFrame,
-        treatment_time: Union[int, float, pd.Timestamp],
+        treatment_time: Union[int, float, pd.Timestamp, tuple, None],
         formula: str,
         model=None,
```
NathanielF, line 84: Ok, so model is None, but we can pass in Intervention estimate pymc models...?

causalpy/experiments/interrupted_time_series.py
```diff
@@ -124,53 +392,47 @@ def __init__(
...
-        self.pre_pred = self.model.predict(X=self.pre_X)
-
-        # calculate the counterfactual
-        self.post_pred = self.model.predict(X=self.post_X)
-        self.pre_impact = self.model.calculate_impact(self.pre_y, self.pre_pred)
-        self.post_impact = self.model.calculate_impact(self.post_y, self.post_pred)
-        self.post_impact_cumulative = self.model.calculate_cumulative_impact(
-            self.post_impact
+        # Postprocessing with handler
+        results = self.handler.data_postprocessing(
```
NathanielF: handler assumes a model, but what happens if none is initialised?

causalpy/pymc_models.py
```diff
@@ -507,3 +507,295 @@ def fit(self, X, t, coords):
...
+        self,
+        treatment_effect_type: str | list[str],
+        treatment_effect_param=None,
+        sample_kwargs=None,
+    ):
+        """
+        Initializes the InterventionTimeEstimator model.
+
+        :param treatment_effect_type: Optional dictionary that specifies prior parameters for the
+            intervention effects. Expected keys are:
```
NathanielF, line 991: This is interesting, so similar to change-point detection, but instead of a mean difference in the level you also look for impulses and trend shifts...but why is the trend and impulse parameterised with sigma and mu? That seems like it's mean differences again?

causalpy/pymc_models.py
```diff
@@ -507,3 +507,295 @@ def fit(self, X, t, coords):
...
+
+            t = pm.Data("t", np.arange(len(X)), dims="obs_ind")
+            X = pm.Data("X", X, dims=["obs_ind", "coeffs"])
+            y = pm.Data("y", y, dims="obs_ind")
+
+            lower_bound = pm.Data("lower_bound", self.time_range[0])
+            upper_bound = pm.Data("upper_bound", self.time_range[1])
+
+            # --- Priors ---
+            treatment_time = pm.Uniform(
```
NathanielF: Do we want to allow for Discrete intervals, or does that make sampling harder?

causalpy/pymc_models.py
```diff
@@ -507,3 +507,295 @@ def fit(self, X, t, coords):
...
+                "treatment_time", lower=lower_bound, upper=upper_bound
+            )
+            beta = pm.Normal(
```
NathanielF: Ah, ok,... so you are using a slope parameter just allow configuration in terms of mu, sigma... get you

causalpy/pymc_models.py
```diff
@@ -507,3 +507,295 @@ def fit(self, X, t, coords):
...
+            if "impulse" in self.treatment_effect_param:
+                impulse_amplitude = pm.Normal(
+                    "impulse_amplitude",
+                    mu=self.treatment_effect_param["impulse"][0],
+                    sigma=self.treatment_effect_param["impulse"][1],
+                )
+                decay_rate = pm.HalfNormal(
+                    "decay_rate", sigma=self.treatment_effect_param["impulse"][2]
+                )
+                impulse = pm.Deterministic(
```
NathanielF, line 1065: Nice.

NathanielF: Might be worth unpacking the math in words a bit  in  doc-string , for each of the three changepoint models

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -402,6 +402,77 @@ def test_its(mock_pymc_sample):
...
+            treatment_time,
+            formula="standardize(deaths) ~ 0 + t + C(month) + standardize(temp)",  # noqa E501
+            model=cp.pymc_models.LinearRegression(sample_kwargs=sample_kwargs),
+        )
+    assert (
+        "If treatment_time is a tuple, provided model must have a 'set_time_range' method"
+        in str(exc_info.value)
+    )
+
+    result = cp.InterruptedTimeSeries(
```
NathanielF: Ok, nice API

docs/source/notebooks/its_no_treatment_time.ipynb
NathanielF, line 1: This graph doesn't look right. Sampler is struggling. Can you comment on it. 

image

If you're comparing these two models. Use a forest plot or similar to show how one is better at identifying the period of treatment: 

image

Notebook generally looks good, API is compelling. Can you maybe speak to how this links to changepoint detection and supply a few references?

Also try and mute the errors in the notebook and add the mathjax formula for each intervention detection type to the notebook too. I know they're in the docstrings but it might be worth discussing how each detection method works in the notebook alongside the demonstration of the API calls.

NathanielF (changes requested):
Looks very powerful. Asked for some tidying and clarifications.

causalpy/experiments/interrupted_time_series.py
JeanVanDyk: That's true, it might get confusing. I've changed `treatment_time` to `time_range`.

causalpy/experiments/interrupted_time_series.py
JeanVanDyk: Yes, that's it. At first, I went with simple `if ... else ...` statements, but together with Ben we realized it would be cleaner and easier to understand using abstract classes.

causalpy/experiments/interrupted_time_series.py
JeanVanDyk: I’m not sure yet to what extent this can be generalized to all time-based experiments. For this project, I’ve mainly focused on ITS, without really considering whether the model could be extended to other types of experiments.

That said, it might indeed be a good way to structure things and could also make the project more scalable. I had also thought at some point about creating an abstract class to handle scikit-learn models. Maybe it could be worth opening an issue to explore this further, and in the meantime we can keep it as is, as an example.

causalpy/experiments/interrupted_time_series.py
JeanVanDyk: Yes, that’s correct. Since the experiment inherits from BaseExperiment and has `self.support_bayes = True`, it should work with any PyMC model.

causalpy/experiments/interrupted_time_series.py
JeanVanDyk: If `None` is provided, an error should be raised before the model is passed through the handlers. Specifically, `super().__init__(model=model)`, leading to `BaseExperiment`, should raise an error if the model is `None`.

causalpy/pymc_models.py
JeanVanDyk: I recall trying to use discrete intervals at some point, but ran into issues because the model would then need to rely on multiple samplers. We therefore decided to stick with continuous variables so that we could rely solely on NUTS.

causalpy/custom_exceptions.py
JeanVanDyk: I tried to follow the approach used in the input_validation method. I noticed they relied on custom exceptions, so I figured that since none of the existing exceptions fit the issue I wanted to raise, it made sense to create my own, following the same pattern.

The idea is that someone might want to provide their own PyMC model for changepoint detection. Since the handler calls set_time_range during data preprocessing, it seems reasonable to require that user-provided models implement a set_time_range method.

Do you think it would be in fact just cleaner to remove it ?

docs/source/notebooks/its_no_treatment_time.ipynb
JeanVanDyk: I’ve added the forest plot, which makes it much easier to see how providing a time range improves the precision of the sampling.

I’ve also removed all the errors and added the references.

Additionally, I included the mathematical formulas for each modeled effect.

causalpy/custom_exceptions.py
NathanielF: No, it's fine. I guess the docstring just threw me a bit, seemed redundant, but i get the idea of it.

docs/source/notebooks/its_no_treatment_time.ipynb
NathanielF: Thank you!

It's a nit but you can add ";" after plots to prevent the axes being printed. 

Optional - but it might be nice to see (or just discuss) a "failure mode" of the model e.g. when applied to a timeseries with no change point.... i'm guessing it'll show uncertainty? That might be as instructive as highlighting the usage.

NathanielF (approved):
This is good work! Thanks so much for adding it. 

I've added a comment or two but consider them optional. Approving now. cc @drbenvincent

docs/source/notebooks/its_no_treatment_time.ipynb
JeanVanDyk: Thanks ! I've added the ";" and I'll check with ben whether I should invest time on this now or later.

causalpy/experiments/interrupted_time_series.py
drbenvincent: Happy any scikit-learn support to be a new issue. And it can be a low priority one because it does make sense to focus our finite time on the Bayesian side of things.

causalpy/pymc_models.py
```diff
@@ -749,3 +749,349 @@ class initialisation.
...
         return idata_outcome, model_outcome
+
+
+class InterventionTimeEstimator(PyMCModel):
+    r"""
+    Custom PyMC model to estimate the time an intervention took place.
+
+    This model implements three types of changepoints: level shift, trend change, and impulse response.
+
+    In words:
```
drbenvincent, line 900: We need a new line before the bullet points in order for the rendered docs to show properly

drbenvincent (changes requested):
So a few high level things. Sorry if this comes across as incredibly annoying, but I want to make sure we are adding new features so they are set up well for future maintainability.

## Re-packaging this as interrupted time series and change point detection

At the moment we're packaging this as two flavours of interrupted time series, and that is how we've been thinking about it. But I wonder if it might be simpler, more understandable, and potentially more accurate if we package up as a) traditional ITS that we have before, b) new change point detection functionality. What do you think?

(Tagging @NathanielF for thoughts)

## Perfecting the handler architecture 

**If we stick with the current approach**, I think there are some improvements to the handler architecture.

Current Flow:
* Handler preprocessing → splits data
* Main init → builds matrices from split data, fits model
* Handler postprocessing → splits data again + builds post-treatment matrices

Better Flow Would Be:
* Handler preprocessing → prepares training data appropriately
* Main init → builds matrices, fits model
* Handler postprocessing → uses the already established data splits

So we could do a refactor... the handlers should probably own the entire data management lifecycle:

## [Optional] Support for `Prior` class 
Now that we've merged support for the custom [`Prior` class](https://github.com/pymc-labs/CausalPy/pull/488), it would be really cool to use that here. That would also need some updates to the docs. _However_ I'm also fine for that to be a separate follow  up PR that we do in the near future.

causalpy/pymc_models.py
NathanielF: I could see the division of ITS and a more general changepoint detection  module being conceptually quite appealing. 

If it's an easy lift to do, maybe do so now. If it's a big push, I'd say let's merge this one and do a follow on PR. 

If i understand you correctly it's just restructuring Jean's existing code so the ITS class calls an independent CP clasd. Not adding new functionality, right?

drbenvincent: As far as I could tell it would involve:
* Leaving the `InterruptedTimeSeries` experiment as it was. 
* Creating a new `ChangePointDetection` class. This would basically be a near duplication of the `InterruptedTimeSeries` class but have your new functionality in it
* Keeping the `InterventionTimeEstimator` model class, but perhaps renaming it to focus more on change point detection.
* The `TreatmentTimeHandler` class hierarchy would disappear. For example, the `KnownTreatmentTimeHandler` would be irrelevant because it's all handled by the `InterruptedTimeSeries` experiment class. And the `UnknownTreatmentTimeHandler` would be merged into the new `ChangePointDetection` class.
* So this would in a way be winding the clock back to when we had two different interrupted time series classes _but_ the difference is that the new one is 'sold' as change point detection. I do feel a bit bad for going back and forth with these suggestions, but my feeling is:
  * This increased separation will be better for maintainability 
  * It would be clearer to end-users
  * The term 'change point detection' is appealing and people just know what it means.
* ~~Extracting out the new content in the interrupted time series example notebook into a new change point detection notebook~~. Slight updates to the new notebook to push the 'change point' aspect. Fine to mention the connections to interrupted time series. And adding a new Change Point sub-section of the "how to" docs page.
* Updating the homepage to highlight the new feature - but I can do that in a different PR.

I'm aware that's not a trivial amount of work. But conceptually it's easy in that it doesn't involve solving any new problems. It can probably be facilitated by use of an LLM. 

Does that seems reasonable @JeanVanDyk, or onerous. If you don't have capacity then I could give it a stab. I'm aware that I've been a bit demanding.

JeanVanDyk: Hi @drbenvincent @NathanielF,

Thanks for your reviews! I also think it would be a nice experiment to include in CausalPy’s package. I’ll start working on it this week, though I can’t promise I’ll have enough time to finish before moving on to another project. I’ll keep you posted on my progress.

Also, if I do get enough time, it might be worth looking into more sophisticated models. I could start digging into the literature to get a sense of the state of the art in Bayesian changepoint modeling.

drbenvincent: If you are on board and happy with the change in 'packaging', then I'm happy to give it a go if you are capacity limited. Let me know @JeanVanDyk

PR #488, opened 2025-06-16
Support for custom priors via Prior class

### LINKED ISSUE #387: Allow users to define custom priors
## High level goals

* The API should include default priors and not _require_ the user to define their own priors.
* But the user should be able to customise their own priors.
* Users should be able to ask the experiments what the default priors are, and use that as part of a workflow in going from default priors to customising priors for specific situations.
* Use of custom priors should be documented not just in the API docs, but we should also have a number of examples sprinkled in to multiple (existing) example notebooks.

## Implementation
We can perhaps learn from `pymc-marketing` which started off allowing users to specify priors by providing `dict`s but which has moved to having a prior class (see https://github.com/pymc-labs/pymc-marketing/pull/759). We may find that `dict`s are fine for our purposes, but there could be advantages of going down the path of using classes. (Tagging @wd60622)

williambdean:
This using `pymc-extras` to allow for custom priors. Might need to adjust the remaining a bit to work. 

Closes #387

---

@drbenvincent did some work to:

* get remote tests passing
* add docstrings
* add tests


causalpy/pymc_models.py
```diff
@@ -245,10 +259,9 @@ def build_model(self, X, y, coords):
             self.add_coords(coords)
             X = pm.Data("X", X, dims=["obs_ind", "coeffs"])
             y = pm.Data("y", y, dims="obs_ind")
-            beta = pm.Normal("beta", 0, 50, dims="coeffs")
-            sigma = pm.HalfNormal("sigma", 1)
```
williambdean: sigma will not be in the model anymore but rather, y_hat_sigma based on the default name generation. Is that breaking change an issue? There is a workaround for this if needed

causalpy/pymc_models.py
```diff
@@ -286,9 +303,8 @@ def build_model(self, X, y, coords):
             X = pm.Data("X", X, dims=["obs_ind", "coeffs"])
             y = pm.Data("y", y[:, 0], dims="obs_ind")
             beta = pm.Dirichlet("beta", a=np.ones(n_predictors), dims="coeffs")
-            sigma = pm.HalfNormal("sigma", 1)
```
williambdean: same breaking change concern

causalpy/pymc_models.py
```diff
@@ -237,6 +248,11 @@ class LinearRegression(PyMCModel):
     Inference data...
     """  # noqa: W605
 
+    default_priors = {
```
drbenvincent, line 403: need to add `@property` decorator here? Or is that remembered from it being done in the `PyMCModel` base class?

Getting an Pylance warning: `Type "dict[str, Prior]" is not assignable to declared type "property"`

causalpy/pymc_models.py
drbenvincent: I don't think this is a big deal. I pushed a fix which make tests pass https://github.com/pymc-labs/CausalPy/pull/488/commits/a9f821c8e84a5edaea2615a0c60582a205056318 (well looks like there is one failing doctest)

causalpy/pymc_models.py
drbenvincent: Replied above

causalpy/pymc_models.py
```diff
@@ -286,9 +305,8 @@ def build_model(self, X, y, coords):
             X = pm.Data("X", X, dims=["obs_ind", "coeffs"])
             y = pm.Data("y", y[:, 0], dims="obs_ind")
             beta = pm.Dirichlet("beta", a=np.ones(n_predictors), dims="coeffs")
```
drbenvincent: We definitely want a custom prior for the Dirichlet. I think the Dirichlet would be used always (or nearly always), but there are plenty of real world use cases where the user might want to change the hyper parameters (currently `a=np.ones(n_predictors)`).

williambdean: Alright. since it is function of data, we will have to handle differently

causalpy/pymc_models.py
williambdean: What line of code bring that on? Maybe having a setter will help?

causalpy/pymc_models.py
williambdean: I've address this with dc20e3e. If user wants to change, they can add to the "priors" at initialization.
Take a look and let me know what you think.

causalpy/pymc_models.py
```diff
@@ -295,16 +300,22 @@ class WeightedSumFitter(PyMCModel):
         "y_hat": Prior("Normal", sigma=Prior("HalfNormal", sigma=1), dims="obs_ind"),
     }
 
+    def priors_from_data(self, X, y) -> Dict[str, Any]:
+        n_predictors = X.shape[1]
+
+        return {
+            "beta": Prior("Dirichlet", a=np.ones(n_predictors), dims="coeffs"),
```
drbenvincent: Just realised that `n_predictors` will equal length of the "coeffs" dim. Does that have to be the case in fact? If so, do we need `priors_from_data`?

Not saying I don't want it, it could be really cool. But just wondering more generally if it's needed or not. Will try to think more with a fresh head in the morning, but does this spark off any thoughts?

causalpy/pymc_models.py
juanitorduz: Agree, if you want a property, maybe we can have a setter method? (not a blocker for now and maybe create an issue?)

juanitorduz (approved):
Great! There is a comment by @williambdean  which we could tackle later (please create an issue :) )

drbenvincent:
Thanks for the review @juanitorduz. My feeling at the moment is...

The current workflow is that computation is kicked off automatically when creating an experiment object instance. As in, you don't first create the experiment then manually call a `fit` method, for example. Because we have that workflow, then having a setter doesn't makes sense. 

Changed `PyMCModel.default_priors` from a `@property` method to a class attribute to match how it's implemented in all subclasses (`LinearRegression`, `WeightedSumFitter`, `PropensityScore`). The property decorator was causing Pylance type warnings since subclasses were defining `default_priors` as dictionaries rather than properties. Since priors are static configuration data that don't depend on instance state, using a class attribute is more appropriate and provides better performance (no method call overhead). This change maintains backward compatibility while making the codebase consistent and eliminating type checker warnings.

PR #489, opened 2025-06-16
Use mock sample in tests

williambdean:
Saving about 50s in the test suite by mocking `pm.sample`

----
📚 Documentation preview 📚: https://causalpy--489.org.readthedocs.build/en/489/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup environment
         run: pip install -e .[test]
       - name: Run doctests
-        run: pytest --doctest-modules --ignore=causalpy/tests/ causalpy/
+        run: pytest --doctest-modules --ignore=causalpy/tests/ causalpy/ --config-file=causalpy/tests/conftest.py
       - name: Run extra tests
         run: pytest docs/source/.codespell/test_notebook_to_markdown.py
       - name: Run tests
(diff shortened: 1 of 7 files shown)


williambdean:
66s -> 16s test suite runs. nice

williambdean:
Doing for doctest would be greater reduction of time

drbenvincent (commented):
This is pretty cool - thanks for the contribution. 

On my local machine...

* `make test` runs in 24 seconds (down from 50 seconds on main) ✅
* `make doctest` speed is unchanged❓

I guess I just have some questions to make sure I understand this

* `mock_sample_for_doctest` kind of looks transparent in that we're overriding `pm.sample` with `pymc.testing.mock_sample`. Though that's not working or doing much?
* `mock_pymc_sample` is provided as an input to tests (like `test_did` for example) but it's never used anywhere in those test functions. So this is a bit confusing. Does this make the use of `sample_kwargs` irrelevant now?

williambdean:
So the pm.sample *is* being mocked in the CI/CD tests. Need to figure out why calling `pytest --doctest-modules causalpy/` doesn't work when `pytest --doctest-modules` does. Is this a blocker for merging this PR?

drbenvincent:
I'll have a look and run locally again, but are you seeing this as finished @williambdean ?

drbenvincent:
Running locally on `main` vs this branch:
* `make test`: 47s -> 11s
* `make doctest 50s -> 50s

Shall we merge anyway and figure out the doctest in another PR?

drbenvincent (approved):

williambdean:
amazing 🚀

PR #494, opened 2025-06-28
Enable multiple treated units in synthetic control quasi experiments

### LINKED ISSUE #456: Upgrade synthetic control to model multiple treated units

drbenvincent:
Closes #456

### DIFF
--- a/causalpy/experiments/synthetic_control.py
+++ b/causalpy/experiments/synthetic_control.py
@@ -155,10 +155,16 @@ class SyntheticControl(BaseExperiment):
             raise ValueError("Model type not recognized")
 
         # score the goodness of fit to the pre-intervention data
-        self.score = self.model.score(
-            X=self.datapre_control.to_numpy(),
-            y=self.datapre_treated.isel(treated_units=0).to_numpy(),
-        )
+        if isinstance(self.model, PyMCModel):
+            self.score = self.model.score(
+                X=self.datapre_control.to_numpy(),
+                y=self.datapre_treated.to_numpy(),
+            )
+        else:
+            self.score = self.model.score(
+                X=self.datapre_control.to_numpy(),
+                y=self.datapre_treated.isel(treated_units=0).to_numpy(),
+            )
 
         # get the model predictions of the observed (pre-intervention) data
         self.pre_pred = self.model.predict(X=self.datapre_control)
@@ -207,75 +213,95 @@ class SyntheticControl(BaseExperiment):
         self.print_coefficients(round_to)
 
     def _bayesian_plot(
-        self, round_to=None, **kwargs
+        self, round_to=None, treated_unit=None, **kwargs
     ) -> tuple[plt.Figure, List[plt.Axes]]:
         """
-        Plot the results
+        Plot the results for a specific treated unit
 
         :param round_to:
             Number of decimals used to round results. Defaults to 2. Use "None" to return raw numbers.
+        :param treated_unit:
+            Which treated unit to plot. Can be an integer index or string name.
+            If None, plots the first treated unit.
         """
         counterfactual_label = "Counterfactual"
 
         fig, ax = plt.subplots(3, 1, sharex=True, figsize=(7, 8))
         # TOP PLOT --------------------------------------------------
         # pre-intervention period
+        primary_unit_idx = self._get_primary_treated_unit_index(treated_unit)
+        primary_unit_name = self.treated_units[primary_unit_idx]
+
+        # For multi-unit, select primary unit for main plot
+        if len(self.treated_units) > 1:
+            pre_pred_plot = self.pre_pred["posterior_predictive"].mu.isel(
+                treated_units=primary_unit_idx
+            )
+            post_pred_plot = self.post_pred["posterior_predictive"].mu.isel(
+                treated_units=primary_unit_idx
+            )
+        else:
+            pre_pred_plot = self.pre_pred["posterior_predictive"].mu
+            post_pred_plot = self.post_pred["posterior_predictive"].mu
+
         h_line, h_patch = plot_xY(
             self.datapre.index,
-            self.pre_pred["posterior_predictive"].mu,
+            pre_pred_plot,
             ax=ax[0],
             plot_hdi_kwargs={"color": "C0"},
         )
         handles = [(h_line, h_patch)]
         labels = ["Pre-intervention period"]
 
+        # Plot observations for primary treated unit
         (h,) = ax[0].plot(
-            self.datapre.index, self.datapre_treated, "k.", label="Observations"
+            self.datapre.index,
+            self.datapre_treated.isel(treated_units=primary_unit_idx),
+            "k.",
+            label=f"Observations ({self.treated_units[primary_unit_idx]})",
         )
         handles.append(h)
-        labels.append("Observations")
+        labels.append(f"Observations ({self.treated_units[primary_unit_idx]})")
 
         # post intervention period
         h_line, h_patch = plot_xY(
             self.datapost.index,
-            self.post_pred["posterior_predictive"].mu,
+            post_pred_plot,
             ax=ax[0],
             plot_hdi_kwargs={"color": "C1"},
         )
         handles.append((h_line, h_patch))
         labels.append(counterfactual_label)
 
-        ax[0].plot(self.datapost.index, self.datapost_treated, "k.")
-        # Shaded causal effect
+        ax[0].plot(
+            self.datapost.index,
+            self.datapost_treated.isel(treated_units=primary_unit_idx),
+            "k.",
+        )
+        # Shaded causal effect for primary treated unit
         h = ax[0].fill_between(
             self.datapost.index,
-            y1=az.extract(
-                self.post_pred, group="posterior_predictive", var_names="mu"
-            ).mean("sample"),
-            y2=np.squeeze(self.datapost_treated),
-            color="C0",
+            y1=post_pred_plot.mean(dim=["chain", "draw"]).values,
+            y2=self.datapost_treated.isel(treated_units=primary_unit_idx).values,
+            color="C2",
             alpha=0.25,
+            label="Causal impact",
         )
         handles.append(h)
         labels.append("Causal impact")
 
-        ax[0].set(
-            title=f"""
-            Pre-intervention Bayesian $R^2$: {round_num(self.score.r2, round_to)}
-            (std = {round_num(self.score.r2_std, round_to)})
-            """
-        )
+        ax[0].set(title=f"{self._get_score_title(round_to)}\nUnit: {primary_unit_name}")
 
         # MIDDLE PLOT -----------------------------------------------
         plot_xY(
             self.datapre.index,
-            self.pre_impact.sel(treated_units=self.treated_units[0]),
+            self.pre_impact.sel(treated_units=self.treated_units[primary_unit_idx]),
             ax=ax[1],
             plot_hdi_kwargs={"color": "C0"},
         )
         plot_xY(
             self.datapost.index,
-            self.post_impact.sel(treated_units=self.treated_units[0]),
+            self.post_impact.sel(treated_units=self.treated_units[primary_unit_idx]),
             ax=ax[1],
             plot_hdi_kwargs={"color": "C1"},
         )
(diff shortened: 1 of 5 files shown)


causalpy/pymc_models.py
```diff
@@ -184,20 +243,43 @@ def print_row(
...
+        # Check if we have multiple treated units
+        if "treated_units" in coeffs.dims:
+            # Multiple treated units case - print coefficients for each unit
+            treated_units = coeffs.coords["treated_units"].values
+            for unit in treated_units:
+                print(f"\nTreated unit: {unit}")
+                unit_coeffs = coeffs.sel(treated_units=unit)
+                unit_sigma = az.extract(self.idata.posterior, var_names="sigma").sel(
+                    treated_units=unit
+                )
```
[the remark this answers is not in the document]
drbenvincent: I don't think this is correct. `labels` is an input to the `print_coefficients` method. Are you maybe getting confused by the two local functions defined above in the `print_coefficients` method?

causalpy/pymc_models.py
```diff
@@ -96,11 +104,38 @@ def _data_setter(self, X) -> None:
...
+        else:
+            has_treated_units = False
+
         with self:
-            pm.set_data(
-                {"X": X, "y": np.zeros(new_no_of_observations)},
-                coords={"obs_ind": np.arange(new_no_of_observations)},
-            )
+            # Get the number of treated units from the model coordinates
+            treated_units_coord = getattr(self, "coords", {}).get("treated_units", [])
```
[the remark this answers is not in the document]
drbenvincent: I don't think that's necessary because `obs_coords` is being constructed in the `_data_setter` method and that's the only place it's being used. Or am I missing something?

causalpy/pymc_models.py
```diff
@@ -163,8 +201,29 @@ def score(self, X, y) -> pd.Series:
...
+        # Handle both single and multiple treated units
+        if "treated_units" in mu_data.dims:
+            # Multiple treated units - we need to score each unit separately
+            treated_units = mu_data.coords["treated_units"].values
+            scores = {}
+
+            for unit in treated_units:
+                unit_mu = mu_data.sel(treated_units=unit).T  # (sample, obs_ind)
+                unit_y = y.sel(treated_units=unit).data
+                unit_score = r2_score(unit_y, unit_mu.data)
```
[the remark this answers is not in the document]
drbenvincent: Good observation, but the line above `mu_data = az.extract(mu, group="posterior_predictive", var_names="mu")` collapses the `chain` and `draw` dimensions into a single `samples` dimension. So we are good here I believe

drbenvincent:
That last commit was quite a big one. Model likelihoods are now 2-dimensional. This means we can avoid a lot of conditional branching based on single vs multiple treated unit situations.

causalpy/pymc_models.py
```diff
@@ -379,9 +468,9 @@ def build_model(self, X, Z, y, t, coords, priors):
             pm.Deterministic(name="cov", var=pt.dot(l=chol, r=chol.T))
 
             # --- Parameterization ---
-            mu_y = pm.Deterministic(name="mu_y", var=pm.math.dot(X, beta_z))
+            mu_y = pm.Deterministic(name="mu_y", var=pt.dot(X, beta_z))
```
NathanielF, line 471: Curious what the difference is here?

causalpy/experiments/synthetic_control.py
```diff
@@ -207,84 +207,105 @@ def summary(self, round_to=None) -> None:
...
+            treated_units=treated_unit
+        )
+        post_pred_plot = self.post_pred["posterior_predictive"].mu.sel(
+            treated_units=treated_unit
+        )
+
         h_line, h_patch = plot_xY(
             self.datapre.index,
-            self.pre_pred["posterior_predictive"].mu,
+            pre_pred_plot,
```
NathanielF: I guess the naming here is a bit strange to my eyes... `pre_pred_plot` seems like it should be plot, but it's a data frame passed to a plot? More a nit than anything...

causalpy/experiments/synthetic_control.py
```diff
@@ -207,84 +207,105 @@ def summary(self, round_to=None) -> None:
...
-            (std = {round_num(self.score.r2_std, round_to)})
-            """
-        )
+        ax[0].set(title=f"{self._get_score_title(treated_unit, round_to)}")
 
         # MIDDLE PLOT -----------------------------------------------
         plot_xY(
             self.datapre.index,
-            self.pre_impact.sel(treated_units=self.treated_units[0]),
+            self.pre_impact.sel(treated_units=treated_unit),
```
NathanielF, line 295: This is definitely better. I like how you flagged the first will be selected as default in the docstrings

causalpy/experiments/diff_in_diff.py
```diff
@@ -114,14 +114,18 @@ def __init__(
             },
         )
         self.y = xr.DataArray(
-            self.y[:, 0],
-            dims=["obs_ind"],
-            coords={"obs_ind": np.arange(self.y.shape[0])},
+            self.y,
+            dims=["obs_ind", "treated_units"],
```
NathanielF, line 118: This is nicer too. Always wary of hardcoded indexs

causalpy/experiments/synthetic_control.py
```diff
@@ -156,8 +156,8 @@ def __init__(
 
         # score the goodness of fit to the pre-intervention data
         self.score = self.model.score(
-            X=self.datapre_control.to_numpy(),
-            y=self.datapre_treated.isel(treated_units=0).to_numpy(),
+            X=self.datapre_control,
```
NathanielF, line 159: cleaner

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -725,3 +725,283 @@ def test_inverse_prop(mock_pymc_sample):
     assert all(isinstance(ax, plt.Axes) for ax in axs)
     with pytest.raises(NotImplementedError):
         result.get_plot_data()
+
+
+@pytest.fixture(scope="module")
+def multi_unit_sc_data(rng):
```
NathanielF, line 731: Nice DGP clear

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -725,3 +725,283 @@ def test_inverse_prop(mock_pymc_sample):
...
+        sc = cp.SyntheticControl(
+            data=df,
+            treatment_time=treatment_time,
+            control_units=control_units,
+            treated_units=treated_units,
+            model=model,
+        )
+
+        # Check basic attributes
+        assert sc.treated_units == treated_units
```
NathanielF, line 838: Looks good

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -725,3 +725,283 @@ def test_inverse_prop(mock_pymc_sample):
...
+        captured = capsys.readouterr()
+        output = captured.out
+
+        # Check that output contains information for multiple treated units
+        assert "Treated units:" in output
+        for unit in treated_units:
+            assert unit in output
+
+    @pytest.mark.integration
+    def test_single_unit_backward_compatibility(self, single_unit_sc_data):
```
NathanielF, line 898: Nice. Good discipline.

causalpy/tests/test_integration_pymc_examples.py
```diff
@@ -725,3 +725,283 @@ def test_inverse_prop(mock_pymc_sample):
...
+            treatment_time=treatment_time,
+            control_units=control_units,
+            treated_units=treated_units,
+            model=model,
+        )
+
+        # Test that invalid treated unit name is handled gracefully
+        # Note: Current implementation may not raise ValueError, so we test default behavior
+        try:
+            sc.plot(treated_unit="invalid_unit")
```
NathanielF, line 1000: I noticed when adding a test earlier that it asked for closing plots after opening
image

causalpy/experiments/synthetic_control.py
```diff
@@ -451,36 +492,83 @@ def get_plot_data_bayesian(self, hdi_prob: float = 0.94) -> pd.DataFrame:
...
+        post_impact_lower_upper = post_impact_hdi.iloc[:, [0, -1]].values
+
+        pre_data[[impact_lower_col, impact_upper_col]] = pre_impact_lower_upper
+        post_data[[impact_lower_col, impact_upper_col]] = post_impact_lower_upper
 
         self.plot_data = pd.concat([pre_data, post_data])
 
         return self.plot_data
+
+    def _get_score_title(self, treated_unit: str, round_to=None):
```
juanitorduz: better type hints rount_to: Optional[fleat] = None and return str?

juanitorduz (approved):
Besides the remaining @NathanielF comments this looks cleaner :)

causalpy/pymc_models.py
drbenvincent: I believe that `pm.math.dot` just aliases `pt.dot`. So this shouldn't have any effect and is purely optional.

causalpy/experiments/synthetic_control.py
drbenvincent: Agreed. Fixed in https://github.com/pymc-labs/CausalPy/pull/494/commits/8579a14312d7e8d29012179d2cbbda9deab11ad0

causalpy/tests/test_integration_pymc_examples.py
drbenvincent: When I run `causalpy/tests/test_integration_skl_examples.py::test_sc` I'm not getting any warnings. Any more info on how to replicate this @NathanielF ?

causalpy/experiments/synthetic_control.py
drbenvincent: Nice. Fixed in https://github.com/pymc-labs/CausalPy/pull/494/commits/3101cdc491622aa45f97013d597c9935e28f2f64 though I make it an Optional[int]

drbenvincent:
<p>Good idea. Implemented in <a href="https://github.com/pymc-labs/CausalPy/pull/494/commits/b213b0f37aec89872b0853bbbff2c67735846462" target="_blank">b213b0f</a></p>

---

 <a href='https://app.reviewnb.com/pymc-labs/CausalPy/pull/494/discussion/#comment-3064949463'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

PR #505, opened 2025-07-19
Fix to minor bug in simulated data generation function

### LINKED ISSUE #433: The argument `N` is partially ignored in `causalpy.data.simulate_data.generate_time_series_data`
Hi there! 

I am not sure if I understand the code exactly right, but i believe that the `N` argment in `causalpy.data.simulate_data.generate_time_series_data` is being ignored. 

From the docstrings I think that the `N` is not propagated through the function (everywhere) and in many cases `100` is hardcoded. 

If I am right, I would be happy to submit a PR for this. If it is more involved, please advise. 

Cheers! 
(in general quite happy with this project!)

drbenvincent:
Closes #433

> P.S.: In the same function, should the input intercept (i.e. `beta_intercept`) in the output dataframe instead of the constant `1`

I believe the answer to this is "no". The `beta_intercept` input is being used in the data generation process, but the purpose of `df["intercept"] = np.ones(df.shape[0])` is to create input data for modelling where it's common to create a column of zeros. Though practically I think this is not needed or is even being used because it's being taken care of by `patsy` based on the user-provided model formula. But for the moment I've just added an explanatory comment.

----
📚 Documentation preview 📚: https://causalpy--505.org.readthedocs.build/en/505/

### DIFF
--- a/causalpy/data/simulate_data.py
+++ b/causalpy/data/simulate_data.py
@@ -125,12 +125,12 @@ def generate_time_series_data(
         The intercept
 
     """
-    x = np.arange(0, 100, 1)
+    x = np.arange(0, N, 1)
     df = pd.DataFrame(
         {
             "temperature": np.sin(x * 0.5) + 1,
-            "linear": np.linspace(0, 1, 100),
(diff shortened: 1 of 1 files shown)


ErikRingen (approved):
LGTM

Issue #390, opened 2024-07-12
Double check correctness of causal impact calculation in scikit-learn flavour of `DifferenceInDifferences`


drbenvincent:
At the moment we have a TODO comment which needs some attention

```python
# This is the coefficient on the interaction term
# TODO: THIS IS NOT YET CORRECT ?????
self.causal_impact = (
    self.y_pred_treatment[1] - self.y_pred_counterfactual[0]
)[0]
```

From memory, this is not wrong, but achieved the right effect in a different manner to the pymc flavour of difference in differences.

Either way, we need to double check and resolve.

roesta07:
Hi @drbenvincent I recently worked on a DID project using Pymc. I would be happy to contribute. Can you elaborate bit more on the issue on bug?
Did you mean to directly reference the interaction parameter(causal_estimate) from the sklearn model's estimate as we did it for pymc model?

drbenvincent:
Hi @roesta07. Sorry for the very delayed reply on this! The notification slipped through the net.

This whole section feels like it could be tightened up to be honest:

https://github.com/pymc-labs/CausalPy/blob/fdce5b0d435e25fbec70337bdb929a519bc0764f/causalpy/experiments/diff_in_diff.py#L193-L206

**For the PyMC models...**

It's clear what the code is doing, but there is no comment about why it's looking for a coefficient containing `"post_treatment"`. There could be some explanation there. And the way it's implemented with a loop and if (with not break) is a bit lazy.

Actually, looking a bit deeper I'm not 100% sure it's robust. When we look at the result summary in the notebook there is the main effect _and_ the interaction term which contains `post_treatment`. It is clear from looking at the figure that the right term (the interaction) is being taken as the causal impact, but it's not clear that the code will always robustly get that coefficient.

```
Model coefficients:
    Intercept                     1.1, 94% HDI [1, 1.1]
    **post_treatment**[T.True]        0.99, 94% HDI [0.92, 1.1]
    group                         0.16, 94% HDI [0.094, 0.23]
    group:**post_treatment**[T.True]  0.5, 94% HDI [0.4, 0.6]
    sigma                         0.082, 94% HDI [0.066, 0.1]
```

[...]

Issue #514, opened 2025-07-29
Allow custom column names for `post_treatment` variable in DiD for clarity


roesta07:
Allow the user to use whatever column name they want for `'post_treatment'` variable while constructing DifferenceInDifferences object with new parameter `post_treatment_variable_name` . Also setting its default value to `'post_treatment' `so that it does not break previously written codes.
@drbenvincent  As discussed more in issue no #390

PR #515, opened 2025-07-30
Refactored scikit-learn flavour of DifferenceInDifferences and allowed custom column names for post_treatment variable.

roesta07:
closes issues #390 and #514

- causal impact calculation in scikit-learn flavour of DifferenceInDifferences
- Allow the user to use whatever column name they want for 'post_treatment' variable while constructing DifferenceInDifferences object with new parameter post_treatment_variable_name . Also setting its default value to 'post_treatment' so that it does not break previously written codes.

### DIFF
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -26,7 +26,6 @@ from sklearn.base import RegressorMixin
 
 from causalpy.custom_exceptions import (
     DataException,
-    FormulaException,
 )
 from causalpy.plot_utils import plot_xY
 from causalpy.pymc_models import PyMCModel
@@ -84,6 +83,7 @@ class DifferenceInDifferences(BaseExperiment):
         formula: str,
         time_variable_name: str,
         group_variable_name: str,
+        post_treatment_variable_name: str = "post_treatment",
         model=None,
         **kwargs,
     ) -> None:
@@ -95,6 +95,7 @@ class DifferenceInDifferences(BaseExperiment):
         self.formula = formula
         self.time_variable_name = time_variable_name
         self.group_variable_name = group_variable_name
+        self.post_treatment_variable_name = post_treatment_variable_name
         self.input_validation()
 
         y, X = dmatrices(formula, self.data)
@@ -128,6 +129,12 @@ class DifferenceInDifferences(BaseExperiment):
             }
             self.model.fit(X=self.X, y=self.y, coords=COORDS)
         elif isinstance(self.model, RegressorMixin):
+            # For scikit-learn models, automatically set fit_intercept=False
+            # This ensures the intercept is included in the coefficients array rather than being a separate intercept_ attribute
+            # without this, the intercept is not included in the coefficients array hence would be displayed as 0 in the model summary
+            # TODO: later, this should be handled in ScikitLearnAdaptor itself
+            if hasattr(self.model, "fit_intercept"):
+                self.model.fit_intercept = False
             self.model.fit(X=self.X, y=self.y)
         else:
             raise ValueError("Model type not recognized")
@@ -173,7 +180,7 @@ class DifferenceInDifferences(BaseExperiment):
             # just the treated group
             .query(f"{self.group_variable_name} == 1")
             # just the treatment period(s)
-            .query("post_treatment == True")
+            .query(f"{self.post_treatment_variable_name} == True")
             # drop the outcome variable
             .drop(self.outcome_variable_name, axis=1)
             # We may have multiple units per time point, we only want one time point
@@ -189,7 +196,10 @@ class DifferenceInDifferences(BaseExperiment):
         # INTERVENTION: set the interaction term between the group and the
         # post_treatment variable to zero. This is the counterfactual.
         for i, label in enumerate(self.labels):
-            if "post_treatment" in label and self.group_variable_name in label:
+            if (
+                self.post_treatment_variable_name in label
+                and self.group_variable_name in label
+            ):
                 new_x.iloc[:, i] = 0
         self.y_pred_counterfactual = self.model.predict(np.asarray(new_x))
 
@@ -198,16 +208,24 @@ class DifferenceInDifferences(BaseExperiment):
             # This is the coefficient on the interaction term
             coeff_names = self.model.idata.posterior.coords["coeffs"].data
             for i, label in enumerate(coeff_names):
-                if "post_treatment" in label and self.group_variable_name in label:
+                if (
+                    self.post_treatment_variable_name in label
+                    and self.group_variable_name in label
+                ):
                     self.causal_impact = self.model.idata.posterior["beta"].isel(
                         {"coeffs": i}
                     )
         elif isinstance(self.model, RegressorMixin):
             # This is the coefficient on the interaction term
-            # TODO: CHECK FOR CORRECTNESS
-            self.causal_impact = (
-                self.y_pred_treatment[1] - self.y_pred_counterfactual[0]
-            ).item()
+            # Store the coefficient into dictionary {intercept:value}
+            coef_map = dict(zip(self.labels, self.model.get_coeffs()))
+            # Create and find the interaction term based on the values user provided
+            interaction_term = (
+                f"{self.group_variable_name}:{self.post_treatment_variable_name}"
+            )
+            matched_key = next((k for k in coef_map if interaction_term in k), None)
+            att = coef_map.get(matched_key)
+            self.causal_impact = att
         else:
             raise ValueError("Model type not recognized")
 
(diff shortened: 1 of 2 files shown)


causalpy/experiments/diff_in_diff.py
```diff
@@ -198,32 +209,66 @@ def __init__(
...
-                "A predictor called `post_treatment` should be in the formula"
-            )
-
-        if "post_treatment" not in self.data.columns:
-            raise DataException(
-                "Require a boolean column labelling observations which are `treated`"
-            )
+        # Check if post_treatment_variable_name is in formula
+        if self.post_treatment_variable_name not in self.formula:
+            if self.post_treatment_variable_name == "post_treatment":
```
drbenvincent: I've got a minor preference to just give one generic exception message, rather than a custom one dependent on `self.post_treatment_variable_name`. That will also cut down on the number of tests required to achieve high test coverage.

causalpy/experiments/diff_in_diff.py
```diff
@@ -198,32 +209,66 @@ def __init__(
...
+                # Custom case - user specified custom name, so remind them what they specified
+                raise FormulaException(
+                    f"Missing required variable '{self.post_treatment_variable_name}' in formula.\n\n"
+                    f"Since you specified post_treatment_variable_name='{self.post_treatment_variable_name}', "
+                    f"please ensure formula includes '{self.post_treatment_variable_name}'"
+                )
+
+        # Check if post_treatment_variable_name is in data columns
+        if self.post_treatment_variable_name not in self.data.columns:
+            if self.post_treatment_variable_name == "post_treatment":
```
drbenvincent: Same comment as above. Just give one more generic exception message, regardless of what `self.post_treatment_variable_name` is.

causalpy/experiments/diff_in_diff.py
```diff
@@ -198,32 +209,66 @@ def __init__(
...
         elif isinstance(self.model, RegressorMixin):
             # This is the coefficient on the interaction term
-            # TODO: CHECK FOR CORRECTNESS
-            self.causal_impact = (
-                self.y_pred_treatment[1] - self.y_pred_counterfactual[0]
-            ).item()
+            # Store the coefficient into dictionary {intercept:value}
+            coef_map = dict(zip(self.labels, self.model.get_coeffs()))
+            # Create and find the interaction term based on the values user provided
+            interaction_term = (
```
drbenvincent, line 231: Nice. We'll need more tests anyway to ensure test coverage, so when you do that can you add cases for when people specify formulas like `post_treatment:a` and `post_treatment*b`. It _should_ work because we'll always get a coefficient for `post_treatment:a`, but it is worth adding the test

causalpy/experiments/diff_in_diff.py
```diff
@@ -128,6 +130,12 @@ def __init__(
             }
             self.model.fit(X=self.X, y=self.y, coords=COORDS)
         elif isinstance(self.model, RegressorMixin):
+            # For scikit-learn models, automatically set fit_intercept=False
```
drbenvincent, line 140: Nice

drbenvincent (changes requested):
* Looks like the remote checks are failing. Sometimes you need to run the pre-commit checks locally twice - the interrogate thing is a bit fiddly.
* And looks like we'll need to increase test coverage. So obvious ones would be to include tests where we use the default, or a user-provided post treatment variable name.

Overall, this is looking good. Thanks for the PR :)

Oh, remember to update from main regularly :)

causalpy/experiments/diff_in_diff.py
roesta07: Yeah absolutely!! More generic ones like "Missing required variable '{self.post_treatment_variable_name}' in formula" can be used

causalpy/experiments/diff_in_diff.py
roesta07: Yeah, will add some tests for a cases where a user provides post treatment variable name and check for `FormulaExeption` and `DataException`

 but @drbenvincent  can you elaborate on this specific test. Are we also checking the coefficient value where two interaction terms are used?

drbenvincent: I'd not thought of that. I guess it's easy to find and interaction term of the post treatment variable and something else. But if there are two interaction terms, both including the post treatment variable, then that might get messy. Can we think of any situations where that be a good idea? If not, then maybe that could throw and exception and we just say we can't deal with a formula like that?

roesta07: Since our users can write any formula freely—unlike other libraries that rely on closed systems—they could specify any formula like `post_treatment * group + post_treatment * group * male` which might be uncommon but it’s entirely possible in our setup. 

The users can obtain estimates for exactly what they define in the formula. However, we’ve built this did object specifically for two-way Diff-in-diff with a single interaction term ?-- thus the other features might get messed up as you said.

So yeah @drbenvincent  I agree that we could throw exception if we encounter any two interaction term with `post_treatment` to move forward

roesta07:
Hi @drbenvincent, here is a draft with following changes:
- setup validation for cases to catch invalid interaction term specifications (e.g., three-way interactions, multiple interaction terms, repeated interaction terms)
- expanded test coverages to test those invalid interaction terms and custom post_treatment_variable_name
- made error messages more generic as stated above
Also not sure how the pre-commits are failing

causalpy/experiments/diff_in_diff.py
```diff
@@ -84,6 +86,7 @@ def __init__(
         formula: str,
         time_variable_name: str,
         group_variable_name: str,
+        post_treatment_variable_name: str = "post_treatment",
```
drbenvincent, line 94: Can you add `post_treatment_variable_name` into the docstring to make it ultra clear what it does

causalpy/experiments/diff_in_diff.py
```diff
@@ -236,6 +262,61 @@ def input_validation(self):
                 coded. Consisting of 0's and 1's only."""
             )
 
+    def _get_interaction_terms(self):
```
drbenvincent: Suggestion (not requirement). This could be made a static method and you could just pass in the formula string as an argument. Or it could be a simple utility function that could go in `utils.py`. It might help making testing the function marginally simpler.

Ideally we'd add some test cases for _get_interaction_terms. As in, come up with a set of example formulas and the expected outputs of the function.

causalpy/experiments/diff_in_diff.py
drbenvincent: I did this

causalpy/experiments/diff_in_diff.py
NathanielF: This is smarter

causalpy/experiments/diff_in_diff.py
```diff
@@ -173,7 +188,7 @@ def __init__(
             # just the treated group
             .query(f"{self.group_variable_name} == 1")
             # just the treatment period(s)
-            .query("post_treatment == True")
+            .query(f"{self.post_treatment_variable_name} == True")
```
NathanielF, line 191: nice

causalpy/experiments/diff_in_diff.py
```diff
@@ -198,31 +216,44 @@ def __init__(
...
+        # Validate formula structure and interaction interaction terms
+        self._validate_formula_interaction_terms()
+
         """Validate the input data and model formula for correctness"""
-        if "post_treatment" not in self.formula:
+        # Check if post_treatment_variable_name is in formula
+        if self.post_treatment_variable_name not in self.formula:
             raise FormulaException(
-                "A predictor called `post_treatment` should be in the formula"
+                f"Missing required variable '{self.post_treatment_variable_name}' in formula"
```
NathanielF, line 250: good

causalpy/experiments/diff_in_diff.py
```diff
@@ -236,6 +267,34 @@ def input_validation(self):
...
+        # Check for interaction terms with more than 2 variables (more than one '*' or ':')
+        for term in interaction_terms:
+            total_indicators = sum(
+                term.count(indicator) for indicator in INTERACTION_INDICATORS
+            )
+            if (
+                total_indicators >= 2
+            ):  # 3 or more variables (e.g., a*b*c or a:b:c has 2 symbols)
+                raise FormulaException(
+                    f"Formula contains interaction term with more than 2 variables: {term}. Only two-way interactions are allowed."
```
NathanielF: Do we want to say that they will be supported or why we have limited it? Not needed.

causalpy/tests/test_input_validation.py
```diff
@@ -60,6 +72,88 @@ def test_did_validation_post_treatment_formula():
...
+        _ = cp.DifferenceInDifferences(
+            df,
+            formula="y ~ 1 + group + group:post_treatment + group:male",
+            time_variable_name="t",
+            group_variable_name="group",
+            model=cp.pymc_models.LinearRegression(sample_kwargs=sample_kwargs),
+        )
+
+    # Test 10: Mixed issues - multiple terms + three-way interaction (should be invalid)
+    with pytest.raises(FormulaException):
```
NathanielF, line 148: Should we say anything about how multiple interactions of this type make the interpretation of the DiD coefficients harder?

causalpy/tests/test_utils.py
```diff
@@ -57,3 +62,43 @@ def test_round_num():
     assert round_num(123.456, 5) == "123.46"
     assert round_num(123.456, 6) == "123.456"
     assert round_num(123.456, 7) == "123.456"
+
+
+def test_get_interaction_terms():
```
NathanielF, line 67: Very cool

NathanielF (approved):
Seems nice to me. Would maybe add a comment to the interaction functionality on how and where interactions are useful in DiD settings.

causalpy/tests/test_input_validation.py
drbenvincent: Should be resolved by https://github.com/pymc-labs/CausalPy/pull/515/commits/ea2ed69dbe2ae1b6fa743c306f066beec896e436

causalpy/experiments/diff_in_diff.py
drbenvincent: Should be resolved by https://github.com/pymc-labs/CausalPy/pull/515/commits/ea2ed69dbe2ae1b6fa743c306f066beec896e436

drbenvincent (approved):

roesta07:
Ohh Nice, thanks @NathanielF and @drbenvincent for moving this

PR #537, opened 2025-10-23
adding to knowledgebase on SCM

NathanielF:
Adding a notebook to the knowledge base on structural causal modelling with Bayesian models.

I think it will be useful to highlight what is unique and special about bayesian causal modelling and lean into the structural causal modelling perspective. This would distinguish CausalPy more as a Bayesian package with a unique selling point. In the notebook i demonstrate joint causal models of outcome and treatment and will discuss links and contrast with the potential outcome framework. 

image

Along the way i demonstrate the usefulness of spike and slab priors for improving identification with a species of joint causal model for both continuous and binary treatments. These could be added as distinct models or just as options for priors in existing models. I'm not entirely sure where to take this modelling and would like to discuss it when the draft is more complete and ready for review. 

Questions for discussion. 

- Structural causal Modelling is present in `CausalPy` via the IV modelling implementation. But i wonder if we want to make the implementation more robust following some the details in this notebook e.g. variable selection priors or the explicit modelling of rho
- CATE inference with BART treatment models are shown here
- General thoughts about SCM and the link between potential outcomes and the way we gloss the relationship between these topics in `CausalPy`. 

----
📚 Documentation preview 📚: https://causalpy--537.org.readthedocs.build/en/537/


drbenvincent (commented):
First-pass review. I like it. Here's a mix of high and low level thoughts at this point...

I'm wondering if we want to use the language of 'model selection' a bit more, especially in the introduction. Would it also be useful and instructive to discuss Savage-Dickey where we have a 'super model' and certain parameters (e.g. being zero) structurally alter the model so that we are effectively exploring different models via the parameter values.

I'd maybe add a legend under the diggity image and take some time to relate that back to the data simulation code. Basically use it as a breather/pause to allow the reader to keep up with the details. [Not sure if we can do a legend, so feel free to ignore the formatting point, but I think some more explanation would help readers.]

We can maybe add the `hide-input` cell tag on cells that are exclusively for plot generation? That's just a personal preference for docs.

Pedantic and python stylistic point, but `fit_models` could be made cleaner with a dictionary comprehension.

Can we get references in there for each of the prior types. Academic papers and/or useful blogposts. Maybe a brief description of each for those not familiar.

Not sure "reduced-form" is defined anywhere?

Can you be more specific with "schizphrenic posterior distribution"

Sorry if this is annoying, but can you add a glossary term (in `glossary.rst`) for "probabilistic programming"

Something iffy going on with formatting here: `"Y_bin ~ T_bin + feature_0 +	feature_1 +	feature_2 +	feature_3 +	feature_4 + feature_5 + feature_6 + feature_7 +feature_8",`

If there are any book chapters in Bayesian stats books which touch on this topic, it would be good to add them in as references.

Link or reference for the credibility revolution?

I think there's scope for a summary directed at practitioners. Are you advocating for the general practice of using these kinds of priors for sparsifying model structure? If so, are there concrete situations where that can be used, and how could a skeptic be convinced of it's utility above and beyond this being a form of basic feature selection when you throw all your variables into the causal soup?

## Typos
* empircal

NathanielF:
On the BART stuff, yeah if the PR is just this knowledge base article. Maybe not needed in environment... but if we wanted to add CATE functionality it might be a good default for flexible approximation with a bayesian flavour

NathanielF:
> I think there's scope for a summary directed at practitioners. Are you advocating for the general practice of using these kinds of priors for sparsifying model structure? If so, are there concrete situations where that can be used, and how could a skeptic be convinced of it's utility above and beyond this being a form of basic feature selection when you throw all your variables into the causal soup?

I was primarily thinking of these as a kind of sensitivity analysis tool for cases with lots of covariates.  Not as a magical salve to replace theory driven instrument selection. Re: summary directed at practitioners i can add one at the end before the conclusion, but i'll keep it brief.

docs/source/_static/forwards_backwards.png
drbenvincent, line 1: Being totally honest I'm not finding this figure clarifies anything. Maybe we could go for a data -> hypothesis (forward model) and hypothesis -> data (backwards inference) kind of flow diagram?

Though I guess that could be adapted for Worlds rather than Hypotheses.

docs/source/knowledgebase/structural_causal_models.ipynb
drbenvincent, line 1: `PH balances` -> `pH balances`

NathanielF:
Following up on point 1, i've demonstrated an example here on a branch off this one: https://github.com/NathanielF/CausalPy/pull/1

I'm using the Prior class but as this is a general functionality think it would be worth discussing more broadly. Opened an issue here: https://github.com/pymc-labs/CausalPy/issues/550

docs/source/_static/forwards_backwards.png
NathanielF: added slightly more exposition making clearer that a world-state is the parameter setting.

NathanielF:
I think i did action @juanitorduz 's comments. I changed the position of the title and broke up the variable selection parameterisations into separate functions.

PR #538, opened 2025-10-27
Don't change matplotlib style on import of causalpy + misc dev fixes

### LINKED ISSUE #519: Importing causalpy changes matplotlib style
This seems a bit opinionated. Would it makes sense to remove?

https://github.com/pymc-labs/causalpy/blob/2016ff86869be16127260702f559b96875c4aed9/causalpy/__init__.py?plain=1#L31

drbenvincent:
* Closes #519 so will not longer change the matplotlib style when causalpy is imported.
* Made a few other small fixes to the `Makefile` to resolve some issues I was having.
* Also had to pin Python version to `>=3.11` to avoid environment problems.

----
📚 Documentation preview 📚: https://causalpy--538.org.readthedocs.build/en/538/

### DIFF
--- a/causalpy/experiments/base.py
+++ b/causalpy/experiments/base.py
@@ -17,6 +17,8 @@ Base class for quasi experimental designs.
 
 from abc import abstractmethod
 
+import arviz as az
+import matplotlib.pyplot as plt
 import pandas as pd
 from sklearn.base import RegressorMixin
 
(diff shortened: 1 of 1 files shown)


causalpy/experiments/base.py
```diff
@@ -63,12 +63,17 @@ def plot(self, *args, **kwargs) -> tuple:
...
         depending on the model type.
         """
-        if isinstance(self.model, PyMCModel):
-            return self._bayesian_plot(*args, **kwargs)
-        elif isinstance(self.model, RegressorMixin):
-            return self._ols_plot(*args, **kwargs)
-        else:
-            raise ValueError("Unsupported model type")
+        import arviz as az
+        import matplotlib.pyplot as plt
```
williambdean: These can potentially be top level imports

drbenvincent: Agree

Issue #542, opened 2025-10-29
Add Self-Documenting Command to Makefile and Tidy It A Bit #909


louismagowan:
Same thing as [what I did in pymc-marketing](https://github.com/pymc-labs/pymc-marketing/pull/910)

drbenvincent:
Closed by #543

PR #543, opened 2025-10-29
feat(Makefile): Add self-documenting commands and re-organise

louismagowan:
Tiny PR - just adding some comments to the Makefile and a self-documenting command that then returns those comments when run. Also, making the default Make command to be the self-documenting command (help) so that users can easily see what Makefile commands are available and what they do, without having to open the makefile.

So users can run make help or just make and see the options easily
Solves: #542 

----
📚 Documentation preview 📚: https://causalpy--543.org.readthedocs.build/en/543/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,31 +1,50 @@
-.PHONY: init lint check_lint test uml html cleandocs doctest
+#################################################################################
+# GLOBALS                                                                       #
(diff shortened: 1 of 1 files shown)


drbenvincent (approved):
Very cool @louismagowan

PR #582, opened 2025-12-05
Add environment activation instructions to `AGENTS.md`

drbenvincent:
----
📚 Documentation preview 📚: https://causalpy--582.org.readthedocs.build/en/582/

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,9 @@
 # AGENTS
 
+## Environment
+
+- **Activation required**: Always run `conda activate CausalPy` before executing any Python commands, tests, or scripts
+
 ## Testing preferences
 
 - Write all Python tests as `pytest` style functions, not unittest classes


AGENTS.md
```diff
@@ -1,5 +1,9 @@
 # AGENTS
 
+## Environment
+
+- **Activation required**: Always run `conda activate CausalPy` before executing any Python commands, tests, or scripts
```
juanitorduz, line 5: I use micromamba XD as conda is siper slow ... and maybe others develop with pixi ... 🤔

drbenvincent: That is a good point. I wonder if there's a way to give instructions which can work out what a particular developer is doing. One solution would be a local file which has that info, but that's a bit of a burden on the developer and might start to put barriers up for new contributors. Maybe we could give it a list of options to try? Open to ideas.

drbenvincent: @copilot any thoughts?

PR #614, opened 2025-12-24
Add `PiecewiseITS` experiment for known interruption dates

### LINKED ISSUE #613: Add `PiecewiseITS` experiment for known interruption dates
## Summary

Add a new experiment class, **PiecewiseITS**, that implements *piecewise / segmented-regression* Interrupted Time Series (ITS) with **known interruption dates**.

This is **distinct** from the existing `InterruptedTimeSeries` experiment, which fits a model to the **pre-intervention** period and forecasts a post-intervention counterfactual. `PiecewiseITS` instead fits **one model to the full time series** and estimates **explicit level and/or slope changes** at each interruption.

This issue intentionally **excludes** inference of unknown change-points; that should be tracked separately.

## Motivation

Many ITS analyses in applied fields are implemented as **segmented regression** / **piecewise linear** models, estimating:

- baseline trend,
- an immediate **level** change at an interruption,
- and/or a **slope** change after an interruption.

This approach:

- naturally supports **one or multiple** interruption dates,
- provides interpretable parameters per interruption,
- enables counterfactual predictions by “turning off” interruption terms,
- can be implemented cleanly as a new experiment without changing semantics of the existing `InterruptedTimeSeries` API.

### In scope (v1)

- `PiecewiseITS` experiment class.
- **Known** interruption dates provided by the user.
- Piecewise/segmented-regression parameterizations with configurable:
  - level change only,
  - slope change only,
  - level + slope change.
- Support for **multiple** interruption dates.
- Posterior summaries and plots for:
  - observed vs fitted,
  - counterfactual (interruption terms removed),
  - estimated effects and cumulative effects.

### Out of scope (track separately)

- Inferring unknown interruption dates / change-points / structural breaks.
- Smooth onset functions / latent switchpoints.

### Class name

- `PiecewiseITS`

### Constructor sketch

```python
PiecewiseITS(
    data: pd.DataFrame,
    outcome: str,
    time: str,
    interruption_times: list["datetime|int|float"],
    # modelling options
    include_level_change: bool = True,
    include_slope_change: bool = True,
    # optional additions
    controls: list[str] | None = None,
    seasonal: dict | None = None,
    model: "PyMC model spec or factory" | None = None,
)
```

Notes:

- `interruption_times` are **known** and define the breakpoints.
- `controls` and `seasonal` are optional; if existing ITS supports these patterns, mirror them for consistency.

### Single interruption (canonical)

Let interruption time be `T0`.

Define:

- `t`: time index (numeric).
- `I0(t) = 1[t >= T0]` (step function).
- `R0(t) = max(0, t - T0)` (ramp / hinge).

Model (level + slope change):

$$
  y_t = \beta_0 + \beta_1 t + \beta_2 I_0(t) + \beta_3 R_0(t) + \epsilon_t
$$

Where:

- `β0`: baseline intercept,
- `β1`: baseline slope,
- `β2`: immediate level change at `T0`,
- `β3`: slope change after `T0`.

### Multiple interruptions

For interruption times `T_k`, k=1..K:

- `I_k(t) = 1[t >= T_k]`
- `R_k(t) = max(0, t - T_k)`

Model:

$$
 y_t = \beta_0 + \beta_1 t + \sum_{k=1}^K \big(\beta_{2k} I_k(t) + \beta_{2k+1} R_k(t)\big) + \gamma^\top X_t + \epsilon_t
$$

Where `X_t` are optional control covariates.

### Effect definitions

Define the “no-interruption” counterfactual by setting interruption contributions to zero:

$$
 \hat{y}^{(cf)}_t = \beta_0 + \beta_1 t + \gamma^\top X_t
$$

The estimated effect at time `t`:

$$
 \Delta_t = \hat{y}_t - \hat{y}^{(cf)}_t = \sum_{k=1}^K \big(\beta_{2k} I_k(t) + \beta_{2k+1} R_k(t)\big)
$$

Cumulative effect over a window `W` is the sum (or integral) of `Δ_t` over `t in W`.

### Noise / likelihood

Follow existing CausalPy patterns and defaults where possible:

- start with a Gaussian likelihood with `sigma` prior,
- optionally allow AR(1) errors or other time-series likelihoods later.

The v1 implementation should keep likelihood options modest unless the repo already provides standardized components.

### Design matrix / features

- Convert `time` into a numeric index `t` (monotonic). For datetime, map to sequential integers or a scaled numeric representation.
- For each interruption `T_k`, generate:
  - `I_k(t)` step column,
  - `R_k(t)` ramp column.
- Guardrails:
  - ensure `interruption_times` are sorted and within data range,
  - handle repeated or very-close times (raise or warn),
  - ensure `t` is strictly increasing.

### Identifiability and priors

- Priors should be weakly informative and consistent with existing defaults.
- Consider standardizing/scaling `t` (e.g., to mean 0, sd 1) to improve sampling; if scaling is used, ensure reported parameters/effects are back-transformed or clearly documented.
- For multiple interruptions, watch for collinearity between ramps/steps; include tests and document recommended practices.

### Relationship to existing experiments

- Keep as a **new experiment class** to avoid semantic confusion with `InterruptedTimeSeries`.
- Reuse shared utilities where appropriate:
  - plotting primitives,
  - posterior summarization,
  - effect/cumulative effect computations,
  - input validation.

### Core results

- Posterior summaries for:
  - baseline intercept and slope,
  - per-interruption level change(s),
  - per-interruption slope change(s).
- Posterior for:
  - fitted `y_hat`,
  - counterfactual `y_cf` (interruption terms off),
  - pointwise effect `Δ_t`,
  - cumulative effect over user-selected windows.

### Plots

Minimum set:

1. Observed `y_t` with posterior mean/HDI for fitted `y_hat`.
2. Counterfactual vs observed (and/or vs fitted) with credible intervals.
3. Effect over time `Δ_t` and cumulative effect.

Optional:

- Per-interruption contribution plots.

## Acceptance criteria

- New `PiecewiseITS` class added with docstring and user-facing documentation.
- Unit tests cover:
  - feature generation (`I_k`, `R_k`),
  - single interruption (level-only, slope-only, both),
  - multiple interruptions,
  - counterfactual consistency (`y_hat - y_cf == Δ_t` within numerical tolerance).
- Example notebook or docs page demonstrating:
  - one interruption,
  - multiple interruptions,
  - interpretation of parameters.
- No changes to behavior of existing `InterruptedTimeSeries`.

## Suggested follow-up issues (separate)

- Change-point / unknown interruption dates (discrete or continuous time; step/ramp with inferred `T_k`).
- Smooth onset functions.
- More flexible piecewise basis functions (e.g., splines) if desired.
- Autocorrelated errors (ARMA, GP) as standardized options if not already present.

### Core methodological / tutorial references (single interruption; extensible to multiple)

- **Wagner AK, Soumerai SB, Zhang F, Ross-Degnan D. (2002).** *Segmented regression analysis of interrupted time series studies in medication use research.* **Journal of Clinical Pharmacy and Therapeutics**, 27(4):299–309.

- **Lopez Bernal J, Cummins S, Gasparrini A. (2017).** *Interrupted time series regression for the evaluation of public health interventions: a tutorial.* **International Journal of Epidemiology**, 46(1):348–355.

- **Taljaard M, McKenzie JE, Ramsay CR, Grimshaw JM. (2014).** *The use of segmented regression in analysing interrupted time series studies: an example in pre-hospital ambulance care.* **Implementation Science**, 9:77.

- **Kontopantelis E, Doran T, Springate DA, Buchan I, Reeves D. (2015).** *Regression based quasi-experimental approach when randomisation is not an option: interrupted time series analysis.* **BMJ**, 350:h2750.

### References explicitly covering multiple interventions / multiple breaks

- **Box GEP, Tiao GC. (1975).** *Intervention analysis with applications to economic and environmental problems.* **Journal of the American Statistical Association**, 70(349):70–79.
  - Classic “intervention analysis” paper; supports multiple interventions and different intervention shapes.

- **Linden A. (2015).** *Conducting interrupted time-series analysis for single- and multiple-group comparisons.* **The Stata Journal**, 15(2):480–500.
  - Practical segmented-regression ITS implementation; includes guidance/examples for specifying **multiple interventions**.

- **Bai J, Perron P. (2003).** *Computation and analysis of multiple structural change models.* **Journal of Applied Econometrics**, 18(1):1–22.
  - Not “ITS” per se, but a classic reference for regression models with **multiple breakpoints** (useful background for multi-break / multi-phase extensions).

drbenvincent:
Closes #613

This pull request introduces support for Piecewise Interrupted Time Series (ITS) analysis in the codebase. The main changes include adding a new experiment class, stateful patsy transforms for specifying level and slope changes at multiple intervention points, a simulation utility for generating piecewise ITS data, and updates to the package API and documentation to expose these new features.

**Piecewise ITS support and API exposure:**
- Added the `PiecewiseITS` experiment class to the codebase and included it in the main package API (`__init__.py`) and experiments API (`causalpy/experiments/__init__.py`). This enables users to import and use `PiecewiseITS` directly.

**Patsy transforms for segmented regression:**
- Introduced a new module `causalpy/transforms.py` providing stateful patsy transforms: `step` for level changes and `ramp` for slope changes at arbitrary intervention points. These can be used in regression formulas for flexible piecewise ITS modeling, supporting both numeric and datetime time variables.
- Exposed `step` and `ramp` transforms in the main package API (`__init__.py`) to allow easy access.

**Data simulation utilities:**
- Added the `generate_piecewise_its_data` function to `causalpy/data/simulate_data.py` for simulating time series data with multiple interventions, customizable level and slope changes, and ground truth counterfactuals for testing and demonstration.

**Documentation and notebook updates:**
- Added a reference to a new notebook `piecewise_its_pymc.ipynb` in the documentation index to demonstrate piecewise ITS analysis.

**Pre-commit configuration:**
- Updated the pre-commit configuration to exclude the new notebook `piecewise_its_pymc.ipynb` from large file checks, ensuring smoother development workflow.

----
📚 Documentation preview 📚: https://causalpy--614.org.readthedocs.build/en/614/


drbenvincent:
bugbot run

drbenvincent:
https://github.com/pymc-labs/CausalPy/pull/614/commits/bca3699b1a58d9f56ae4d78dc6b8d910406e5a06 adds the most amazing patsy-based API for segmented/piecewise regression!

causalpy/experiments/piecewise_its.py
```diff
@@ -0,0 +1,723 @@
...
+from causalpy.plot_utils import plot_xY
+from causalpy.pymc_models import PyMCModel
+from causalpy.utils import round_num
+
+from .base import BaseExperiment
+
+LEGEND_FONT_SIZE = 12
+
+
+class PiecewiseITS(BaseExperiment):
```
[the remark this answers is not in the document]
drbenvincent: resolved by https://github.com/pymc-labs/CausalPy/pull/614/commits/c4f0568303011edfc7a276ba51a867fbe8f8d821

causalpy/experiments/piecewise_its.py
```diff
@@ -0,0 +1,721 @@
...
+
+    def _extract_time_column(self) -> str:
+        """Extract the time column name from step/ramp calls in formula."""
+        # Match step(var, ...) or ramp(var, ...)
+        pattern = r"(?:step|ramp)\s*\(\s*(\w+)\s*,"
+        matches = re.findall(pattern, self.formula)
+        if matches:
+            return matches[0]
+        # Fallback: try to find a time-like column
+        return "t"
```
JeanVanDyk: I noticed that the current logic merges all thresholds into a single list, regardless of the variable name (for example, step(t, 10) and step(month, 5) would result in thresholds = [10, 5]). This loses the context of which limit applies to which variable.

Is it intended to support multiple tracking variables within a single formula?

If yes: We should consider storing these in a dictionary (e.g., {"t": [10], "month": [5]}) to ensure the thresholds are applied to the correct variables later in the execution.

If no: It might be safer to check the number of unique variables found and raise a ValueError if more than one is detected. This would prevent unexpected behavior if a user provides a complex formula.

causalpy/experiments/piecewise_its.py
```diff
@@ -0,0 +1,721 @@
...
+            try:
+                threshold = pd.Timestamp(first_interruption)
+                post_mask = self.data[time_col] >= threshold
+            except Exception:
+                # Fallback: try direct comparison
+                post_mask = self.data[time_col] >= first_interruption
+        else:
+            # Numeric threshold
+            post_mask = self.data[time_col] >= first_interruption
+
```
JeanVanDyk, line 396: I see we handle str to Timestamp conversion with a fallback to direct comparison. Is there a specific case where we need to compare raw strings that aren't timestamps?

If we are primarily dealing with dates or numbers, I wonder if it wouldn't be safer to convert everything to the proper type (using pd.to_datetime) right at the beginning of the pipeline?

My thinking is that it might allow us to "fail fast" if a user provides an invalid date, and it would simplify the final comparison to a single line: self.data[time_col] >= first_interruption.

I might be overlooking a specific scenario where this late-stage conversion is necessary, so I'd love to hear your thoughts on the intent here!

causalpy/experiments/piecewise_its.py
```diff
@@ -0,0 +1,721 @@
...
+        self, threshold: int | float | str
+    ) -> int | float | pd.Timestamp:
+        """Convert threshold to appropriate type for matplotlib plotting."""
+        if isinstance(threshold, str):
+            # Try to parse as datetime
+            try:
+                return pd.Timestamp(threshold)
+            except Exception:
+                return threshold  # type: ignore[return-value]
+        return threshold
```
JeanVanDyk: If we standardize the threshold types to pd.Timestamp or numeric values immediately upon extraction, this method likely becomes redundant as the data would already be in its final, usable form. I might be overlooking a specific edge case, but it seems that ensuring clean types at the entry point would allow us to simplify the class by removing this defensive logic and the repetitive try/except checks.

JeanVanDyk (commented):
Hi! I’ve done a brief review of the changes, and I must say the notebook is extensive and really interesting—the variety of examples makes the functionality very clear. I’ve also run the notebook and the tests locally: everything works and passes as expected.

However, I noticed a few points that might need addressing before we merge:

Data Structure & Types: There are a few places where date/threshold handling is quite defensive (using multiple try/except and isinstance checks). I suspect we could simplify the entire class by standardizing these to pd.Timestamp or numeric types at the initial extraction point. This would also allow us to remove the redundant _convert_threshold_for_plotting helper.

Missing Method: It looks like the effect_summary method is currently missing. Since the global refactor, this seems to have been dropped or overlooked, but it's quite central to the experiment's output.

Overall, great work on the documentation and examples! Let me know what you think about streamlining the type handling and re-adding the summary method.

drbenvincent:
Thanks for this! I think you are right, there's no need for this to be a stateful transform. It can just be a regular function. 

I think this specific example 

```
f = "y ~ x + x:I(x >= 10)" # ramp
```

might be problematic. Because `x` is not zero at 10 it creates a step and a ramp. I have memories of seeing a paper which flagged that this interaction approach could be problematic. And I guess the size of the step will vary with `x` which could make the interpretation of the coefficient a bit tricky. Would it have to be this?

```python
f = "y ~ x + I(x-10):I(x >= 10)" # ramp
```

(Typing this at 3am after a toddler wake-up, so we'll see if this makes sense in the morning 🤣)

causalpy/experiments/piecewise_its.py
drbenvincent: Evaluation: Good catch — current threshold extraction loses variable-specific context, which can lead to ambiguous behavior if formulas mix time variables.

Plan (todo, deferred):
- Decide contract first: either (a) support exactly one time variable in step/ramp, or (b) fully support multiple variables.
- If we keep single-variable scope, add explicit validation for >1 unique variable and raise a clear ValueError.
- If we support multi-variable scope, refactor to structured storage (e.g. per-variable thresholds) and update downstream plotting/effect logic accordingly.
- Add targeted tests for mixed-variable formulas so behavior is explicit and stable.

Note to self: implement this in the next pass when we address review comments.

causalpy/experiments/piecewise_its.py
drbenvincent: Evaluation: Agree with the fail-fast direction. The current late conversion + fallback is defensive, but it obscures expected input types and can hide invalid threshold values until later stages.

Plan (todo, deferred):
- Move threshold normalization/validation earlier in the pipeline (during extraction/validation).
- Enforce a single canonical type path (numeric or datetime) for each run.\n- Surface invalid threshold/time-column combinations with explicit, actionable errors.
- Keep comparison logic simple once validated (avoid broad try/except in post-intervention mask construction).
- Add tests for valid datetime strings, invalid datetime strings, and numeric thresholds.

Note to self: handle this together with the formula/threshold-contract cleanup above.

causalpy/experiments/piecewise_its.py
drbenvincent: Evaluation: This is consistent with the previous point — if threshold typing is normalized up front, _convert_threshold_for_plotting and similar defensive conversions can likely be simplified or removed.

Plan (todo, deferred):
- After introducing early canonicalization, audit all threshold-conversion call sites.
- Remove redundant conversion helpers where possible, or narrow them to a single well-documented edge case if needed.
- Ensure plotting and post-period masking consume the same validated threshold representation.
- Add regression tests to confirm parity for both numeric and datetime interruption examples.

Note to self: do this cleanup only after the validation refactor lands so we avoid churn.

causalpy/experiments/piecewise_its.py
drbenvincent: @JeanVanDyk Implemented in `8931a805`.

Evaluation / what changed now:
- Added fail-fast validation so all `step()/ramp()` terms must use exactly one time variable in `PiecewiseITS`.
- Added early threshold canonicalization at initialization (numeric for numeric time columns, `Timestamp` for datetime columns) with explicit `FormulaException` on invalid thresholds.
- Simplified downstream logic to use canonical thresholds directly in post-period masking and plotting (removed broad fallback behavior).

Plan status:
- This addresses the current review points for the fail-fast path.
- Multi-variable step/ramp support remains intentionally deferred to a future PR.

Validation:
- Added tests for mixed-variable rejection + invalid numeric/datetime thresholds.
- Ran targeted PiecewiseITS tests and pre-commit checks.

causalpy/experiments/piecewise_its.py
drbenvincent: @JeanVanDyk Correction: resolved in commit 8931a805 (threshold normalization/validation now happens up front, with fail-fast FormulaException for invalid datetime thresholds).

causalpy/experiments/piecewise_its.py
drbenvincent: @JeanVanDyk Correction: resolved in commit 8931a805 (downstream threshold conversion/fallback logic simplified; canonicalized thresholds are used directly).

drbenvincent:
Implemented follow-up fix for the two failing PiecewiseITS tests.

What changed:
- Added `effect_summary(...)` implementation to `PiecewiseITS` for both OLS and PyMC paths.
- Wired it to shared reporting helpers (`_extract_window`, `_extract_counterfactual`, `_compute_statistics*`, `_generate_table*`, `_generate_prose*`).
- Added explicit guard: `period` is not supported for `PiecewiseITS` (clear `ValueError` message).

Why:
- `PiecewiseITS` was falling back to `BaseExperiment.effect_summary`, which raises `NotImplementedError`.

Validation:
- `causalpy/tests/test_piecewise_its.py::test_piecewise_its_effect_summary_ols` ✅
- `causalpy/tests/test_piecewise_its.py::test_piecewise_its_effect_summary_pymc` ✅

Commit: `bcd83273`

Issue #634, opened 2026-01-07
iv_vs_priors.ipynb fails: az.plot_energy() incompatible with numpyro sampler


drbenvincent:
## Summary

The `iv_vs_priors.ipynb` notebook fails during CI execution due to `az.plot_energy()` attempting to access an `energy` attribute that doesn't exist in the sample stats when using the numpyro sampler.

## Blocker

This is blocking PR #633 (Add pre-commit hook to validate BibTeX files) - the notebook test failure is unrelated to that PR's changes.

**Note:** This may also be blocking PR #602 if that PR depends on passing notebook tests.

## Error Details

```
papermill.exceptions.PapermillExecutionError: 
---------------------------------------------------------------------------
Exception encountered at "In [9]":
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[9], line 1
----> 1 az.plot_energy(result_spike_slab.idata, figsize=(20, 6));

File /opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/arviz/plots/energyplot.py:110, in plot_energy(data, kind, bfmi, figsize, legend, fill_alpha, fill_color, bw, textsize, fill_kwargs, plot_kwargs, ax, backend, backend_kwargs, show)
--> 110     energy = convert_to_dataset(data, group="sample_stats").energy.transpose("chain", "draw").values

File /opt/hostedtoolcache/Python/3.12.12/x64/lib/python3.12/site-packages/xarray/core/common.py:306, in AttrAccessMixin.__getattr__(self, name)
--> 306 raise AttributeError(
    307     f"{type(self).__name__!r} object has no attribute {name!r}"
    308 )

AttributeError: 'Dataset' object has no attribute 'energy'
```

## Root Cause

The notebook uses `nuts_sampler="numpyro"` in its `sample_kwargs`:

```python
sample_kwargs = {
    "draws": 1000,
    "tune": 2000,
    "chains": 4,
    "cores": 4,
    "target_accept": 0.95,
    "progressbar": True,
    "random_seed": 42,
    "nuts_sampler": "numpyro",
}
```

When using the numpyro sampler, the `energy` variable is not included in `sample_stats` (or is named differently), causing `az.plot_energy()` to fail.

## Location

- **File:** `docs/source/notebooks/iv_vs_priors.ipynb`
- **Cell:** 14 (In[9] during execution)
[...]

PR #635, opened 2026-01-07
remove energy plot with numpyro sampling

NathanielF:
Just removed the energy plot call as per bug here: 

https://github.com/pymc-labs/CausalPy/issues/634

----
📚 Documentation preview 📚: https://causalpy--635.org.readthedocs.build/en/635/


drbenvincent:
I think https://github.com/pymc-labs/CausalPy/issues/634 is happening because the notebook tests use the trick of sampling from the prior (fast) not the posterior.

PR #679, opened 2026-01-20
Replace ZeroSumNormal with Normal for _annual_seasonal

maresb:
I'm pretty sure you want to use Normal here instead of ZeroSumNormal. I can explain more later.

----
📚 Documentation preview 📚: https://causalpy--679.org.readthedocs.build/en/679/


drbenvincent (changes requested):
Thanks for this. Would you be able to put in a very brief explanation in the docstring (or code comment) so that this doesn't end up getting reverted in the future?

drbenvincent (approved):

Issue #681, opened 2026-01-21
Enhance Skills/Rules with Estimand-Aware Guidance


ErikRingen:
I started thinking about this when trying to understand how different experiment classes compute their causal estimates. In particular I was thinking about the differences between those that have a parametric model of the treatment period (`IV`, `ANCOVA`) vs those that do not (`InterruptedTimesSeries` `SyntheticControl`). The former is estimating differences in expectations while the latter is estimating something like the individual treatment effect, which includes the observation noise.

## Problem
The `.claude/skills` and `.cursor/commands` guidance focuses on how/when to run methods
but not *what* each method estimates. This could lead to incorrect guidance, especially when the
estimand or computation differs from the plot narrative (e.g., DiD vs ITS/SC).

## Why This Matters for Skills/Rules
These files are the first stop for AI behavior and user guidance. Without
estimand context, the assistant can:
- Recommend the wrong method for a target effect (ATT vs LATE vs time-varying)
- Misinterpret time-varying causal impact as population-level ATE/ATT
- Confuse counterfactual plots with reported estimands

## Proposed Enhancement (Skills/Rules First)
Add a consistent "Estimand Guidance" block to each method entry in:
- `.claude/skills/performing-causal-analysis/reference/*.md`
- `.cursor/commands/causalpy_methods.md`

Required fields:
- Estimand (ATT / LATE / discontinuity at cutoff / time-varying impact)
- Computation (coefficient vs observed-minus-counterfactual vs weighted mean)
- Key assumptions
- Short population note only when critical
  (e.g., IV -> compliers; overlap weights -> overlap population)

### Example (DiD)
Estimand: ATT (difference-in-differences interaction effect)  
Computation: coefficient on group x post interaction  
Assumptions: parallel trends, no anticipation, SUTVA  
Note: Plots show counterfactual trajectories, but the reported effect is the
interaction coefficient.

### Example (ITS / Synthetic Control)
Estimand: time-varying causal impact for a treated unit  
Computation: Y_observed(t) - E[Y_counterfactual(t)]  
Assumptions: stable pre-trend, no concurrent shocks, correct counterfactual model  
Note: This is not a population ATE/ATT; it is unit-specific and time-indexed.

### Example (IV)
Estimand: LATE  
Computation: 2SLS coefficient  
Assumptions: relevance, exclusion restriction, monotonicity  
Note: Effect is for compliers (important for interpretation).

### Example (IPW)
[...]

PR #683, opened 2026-01-23
Add estimands knowledgebase page for core methods

ErikRingen:
Documents the causal estimands, computation approaches, and key assumptions for DiD, ITS, Synthetic Control, and Regression Discontinuity. Clarifies the distinction between parametric interaction models (ATT via coefficients) and pre-post counterfactual models (time-varying unit-specific impacts).

Towards #681 

----
📚 Documentation preview 📚: https://causalpy--683.org.readthedocs.build/en/683/

### DIFF
--- a/docs/source/knowledgebase/estimands.md
+++ b/docs/source/knowledgebase/estimands.md
@@ -1,182 +1,240 @@
 # Estimands in CausalPy
 
-Understanding **what** a method estimates is just as important as knowing **how** to use it. This page documents the causal estimands, computation approaches, and key assumptions for CausalPy's core methods.
+Understanding **what** a method estimates is just as important as knowing
+**how** to use it. This page introduces a framework for thinking about causal
+estimands and connects CausalPy's methods to this framework.
 
-## Why Estimands Matter
+## The Estimand Framework
 
-Different causal inference methods target different causal quantities. Misunderstanding what a method estimates can lead to:
+Following {cite:t}`lundberg2021estimand`, we distinguish three interconnected
+concepts in causal inference:
 
-- Choosing the wrong method for your research question
-- Misinterpreting the magnitude or meaning of effects
-- Drawing incorrect conclusions about generalizability
+```
+Theoretical Estimand     -->  Empirical Estimand      -->  Estimator
+```
 
-CausalPy methods fall into two broad categories:
+- **Theoretical Estimand**: The causal quantity of interest, defined by the
+  research question. This specifies *what* causal effect we want to know, for
+  *whom*, and over *what time period*. Examples: "the average effect of a job
+  training program on weekly earnings for program participants" or "the
+  cumulative impact of a marketing campaign on sales for the treated region."
 
-- **Parametric interaction models** (DiD, ANCOVA): Estimate treatment effects via model coefficients representing population-level averages
-- **Pre-post counterfactual models** (ITS, SC): Estimate time-varying, unit-specific impacts by comparing observed outcomes to counterfactual predictions
+- **Empirical Estimand**: A specific, data-linked quantity that can be
+  identified under a set of assumptions. Examples: {term}`ATT`, {term}`ATE`,
+  {term}`LATE`, local effect at a cutoff, time-varying unit-specific impact.
 
-This distinction affects how you should interpret results and plots.
+- **Estimator**: The combination of a statistical model and a computational
+  procedure used to produce an estimate from data. In CausalPy, this typically
+  means choosing a model type (Bayesian via PyMC or OLS via scikit-learn)
+  together with a computation to extract the causal quantity (e.g., coefficient
+  extraction, g-computation). The same design can be implemented with different
+  estimators---for example, a DiD design can use either a Bayesian model with
+  g-computation or an OLS model with coefficient extraction.
+
+The connection between theoretical and empirical estimands requires
+**identification assumptions**---claims about the data-generating process that
+cannot be tested from data alone. These assumptions are often formalized using
+Directed Acyclic Graphs (DAGs). See {doc}`quasi_dags` for DAG-based
+identification strategies for each quasi-experimental method.
+
+The **structure of available data** also constrains which empirical estimands
+are even *candidates*. Panel data with treated and control groups makes the
+ATT a candidate; a single time series restricts you to unit-specific impacts;
+data with a threshold-based assignment mechanism makes local effects at the
+cutoff a possibility. But data structure alone does not guarantee credibility---
+the identification assumptions must also be defensible. The choice of empirical
+estimand is thus jointly determined by the research question, the available
+data, and the assumptions one is willing to defend.
+
+The estimator is the machinery that transforms data into an estimate. Different
+estimators make different bias-variance trade-offs and have different data
+requirements. See {doc}`structural_causal_models` for a deeper treatment of
+structural versus reduced-form approaches.
 
 :::{note}
-The estimand is not always uniquely determined by the experiment class. Some methods have explicit parameters that change the target (e.g., IPW's `weighting_scheme` selects between ATE and ATO). Others produce "local" effects whose scope depends on design choices (e.g., RD bandwidth, IV instrument). The descriptions below assume standard usage.
+This is an iterative process: estimates inform new theoretical questions,
+refining our understanding over time.
 :::
+
+---
+
+## CausalPy Methods: From Questions to Estimates
+
+Each CausalPy experiment class targets a specific empirical estimand and
+supports different estimators (Bayesian or OLS). Understanding which estimand
+your method targets---and under what assumptions---is essential for valid
+causal interpretation.
+
+### Difference-in-Differences
+
+**Typical research questions**: What is the effect of a policy that was
+implemented in some regions/groups but not others? Did the intervention cause a
+change in outcomes for the treated group?
+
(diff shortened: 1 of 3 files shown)


drbenvincent (changes requested):
I think this will be an excellent addition. My own personal request would be to not hard-wrap lines. 

I got a review from Claude - which picked up a few things.

I'm not 100% sure of its point 2.3. And I don't care if it thinks g-computation might surprise Epi readers 🤣 We could maybe add a short description in the glossary, but don't feel you have to do that in this PR. We are likely going to increase the mention of g-computation in the repo in the near future anyway.

---

## Overall Assessment

This is a valuable addition to CausalPy's knowledgebase. The Lundberg et al. (2021) framework is well-chosen as an organizing principle, and the per-method sections provide a consistent structure (research question → estimand → assumptions → estimator → interpretation). The writing is clear and accessible.

However, the document contains **several factual errors** about how CausalPy actually computes treatment effects, particularly for Regression Discontinuity. These must be corrected before merging. Below, findings are organized by severity.

---

## 1. Errors (Must Fix)

### 1.1 RD estimator is mislabeled as "Coefficient-based"

**Lines 190–193 and Quick Reference table (line 228)**

The document states:

> **Estimator**: Coefficient-based. The treatment effect is estimated as the discontinuity in predicted outcomes at the cutoff, typically using local polynomials around the running variable threshold.

This is self-contradictory and factually wrong. The actual implementation in `regression_discontinuity.py` (lines 175–203) computes the treatment effect by:

1. Creating two prediction points: `threshold - epsilon` (untreated) and `threshold + epsilon` (treated).
2. Predicting outcomes at both points using the fitted model.
3. Taking the difference: `pred[above] - pred[below]`.

This is **prediction-based** (closer to G-computation), not coefficient extraction. The effect is stored in `self.discontinuity_at_threshold`, not extracted from a coefficient vector.

**Suggested fix**: Change "Coefficient-based" to "Prediction-based" (or "G-computation") and update the Quick Reference table accordingly.

### 1.2 RD does NOT use "local polynomials"

**Line 191**

CausalPy does **not** implement local polynomial regression. The implementation uses standard regression formulas via `patsy.dmatrices()` (e.g., `y ~ 1 + x + treated + x:treated`). There is an optional `bandwidth` parameter that filters data to a window around the cutoff, but the model fitted within that window is still a standard parametric regression — not a local polynomial estimator (as used by e.g. `rdrobust`).

**Suggested fix**: Remove the phrase "typically using local polynomials" and replace with something like: "using a parametric regression model evaluated on both sides of the cutoff. An optional `bandwidth` parameter restricts the data to a window around the threshold."

### 1.3 RD estimand terminology conflicts with IV

**Line 175**

The document calls the RD estimand:

> Local average treatment effect at the cutoff

In the causal inference literature, "LATE" (Local Average Treatment Effect) is specifically associated with Instrumental Variable designs and refers to the complier average causal effect. CausalPy's own glossary defines LATE as:

> *"Also known as the complier average causal effect (CACE), is the effect of a treatment for subjects who comply with the experimental treatment assigned to their sample group. It is the quantity we're estimating in IV designs."*

Using the same term for RD creates confusion. The standard RD terminology is **"local treatment effect at the cutoff"** or simply **"the sharp RD estimand"**.

**Suggested fix**: Change "Local average treatment effect at the cutoff" to "Local treatment effect at the cutoff" (dropping "average" removes the LATE association) or use a more explicit label like "Sharp RD treatment effect at the cutoff."

---

## 2. Inaccuracies / Omissions (Should Fix)

### 2.1 IPW weighting schemes are incomplete

**Lines 208–209**

The document lists three `weighting_scheme` options:

> `"raw"`, `"robust"` → ATE; `"overlap"` → overlap population estimand

But the actual implementation (`inverse_propensity_weighting.py`, line 42-44) supports **four** options:

- `"raw"` — ATE
- `"robust"` — ATE
- `"overlap"` — overlap population (ATO)
- **`"doubly robust"` / `"doubly_robust"`** — omitted from the document

**Suggested fix**: Add `"doubly_robust"` to the list and clarify which estimand it targets.

### 2.2 Quick Reference table is inconsistent with implementation

**Line 228**

The table lists RD's computation as "Coefficient-based". Per finding 1.1, this should be "Prediction-based" or "G-computation (local)".

| Method | Empirical Estimand | Computation |
|--------|-------------------|-------------|
| Regression Discontinuity | Local treatment effect at cutoff | **Prediction-based** |

### 2.3 SC "Convex hull coverage" assumption may not apply to all CausalPy models

**Line 152**

The "convex hull coverage" assumption comes from the original Abadie et al. synthetic control framework, which requires non-negative weights summing to one. CausalPy's Bayesian SC implementation does not necessarily enforce these constraints — the Bayesian model can produce weights outside [0, 1] or that don't sum to one.

**Suggested fix**: Qualify this assumption, e.g., "In the standard (non-Bayesian) formulation, the treated unit must lie within the convex hull of control units. CausalPy's Bayesian implementation relaxes this constraint."

---

## 3. Conceptual / Pedagogical Suggestions (Nice to Have)

### 3.1 DiD "No interference" assumption is SUTVA, not DiD-specific

**Line 83**

"No interference between units" (SUTVA) is a general assumption required by all causal methods covered in the document, not unique to DiD. Listing it only under DiD implies it's specific to that method. Consider either:
- Stating SUTVA once at the top as a universal assumption, or
- Listing it under all methods for consistency.

### 3.2 The ASCII pipeline diagram could be improved

**Lines 12–14**

```
Theoretical Estimand     -->  Empirical Estimand      -->  Estimator
```

This could be improved with a proper Mermaid diagram or a styled admonition. The plain ASCII is functional but feels out of place in otherwise well-formatted documentation. For example:

```
:::{mermaid}
graph LR
    A["Theoretical Estimand<br/>(research question)"] --> B["Empirical Estimand<br/>(ATT, ATE, LATE, ...)"]
    B --> C["Estimator<br/>(model + computation)"]
:::
```

### 3.3 Cross-reference style inconsistency

**Lines 86, 118, 156, 187**

Links to `quasi_dags` use raw Markdown link syntax:

```markdown
[Difference in Differences section of quasi_dags](quasi_dags.ipynb#difference-in-differences)
```

While the rest of the document uses MyST cross-reference syntax (`{doc}`, `{term}`, `{cite:t}`). The auto-generated section anchors from notebook headers are fragile — renaming a header silently breaks the link. Consider using `{ref}` with explicit labels if possible, or at minimum note this fragility.

### 3.4 ITS and SC could mention `effect_summary()` more prominently

The ITS section (line 129) mentions `effect_summary()` but SC does not, despite both supporting it. Since `effect_summary()` is the primary programmatic way users access results, it would be helpful to mention it for SC as well.

### 3.5 IV and ANCOVA are referenced but have limited CausalPy support

**Lines 206–216, 230**

The "Context-Dependence" note references IV, and the closing paragraph mentions IV, IPW, and ANCOVA. However:
- **IV**: `plot()` and `summary()` both raise `NotImplementedError`.
- **IPW**: Bayesian-only (`supports_ols = False`).
- **ANCOVA (PrePostNEGD)**: Bayesian-only.

Users reading this doc might expect full feature parity. A brief note about maturity levels would help set expectations.

### 3.6 Minor: "G-computation" terminology

The use of "G-computation" for ITS and SC is technically defensible (predict-under-intervention), but in the epidemiology literature, G-computation (Robins, 1986) refers specifically to standardization over the joint distribution of confounders. Readers from that background may find the term surprising. A brief parenthetical — e.g., "G-computation (counterfactual prediction)" — on first use would help.

---

## 4. Summary Table

| # | Issue | Severity | Lines |
|---|-------|----------|-------|
| 1.1 | RD estimator is NOT "Coefficient-based" | **Error** | 190–193, 228 |
| 1.2 | RD does NOT use "local polynomials" | **Error** | 191 |
| 1.3 | RD "LATE" terminology conflicts with IV/glossary | **Error** | 175 |
| 2.1 | IPW missing `"doubly_robust"` weighting scheme | Omission | 208–209 |
| 2.2 | Quick Reference table wrong for RD | Inaccuracy | 228 |
| 2.3 | SC convex hull assumption may not apply to Bayesian SC | Inaccuracy | 152 |
| 3.1 | SUTVA listed only under DiD | Pedagogical | 83 |
| 3.2 | ASCII diagram could be a Mermaid diagram | Style | 12–14 |
| 3.3 | Cross-reference style inconsistency | Style | 86, 118, 156, 187 |
| 3.4 | SC section doesn't mention `effect_summary()` | Omission | ~165 |
| 3.5 | IV/ANCOVA maturity not noted | Clarity | 206–216, 230 |
| 3.6 | "G-computation" may confuse epi readers | Clarity | 121, 159 |

---

## 5. What Works Well

- **The Lundberg et al. framework** is an excellent organizing choice and provides a principled way for users to think about estimands.
- **Consistent per-method structure** (question → estimand → assumptions → estimator → interpretation) makes the document scannable and easy to reference.
- **Explicit identification assumptions** per method is very valuable — users often skip this step.
- **Interpretation notes** calling out common pitfalls (e.g., DiD's coefficient vs. counterfactual plot, ITS temporal variation, RD locality) are practical and helpful.
- **The Quick Reference table** is a nice summary (once corrected).
- **Glossary term linking** via `{term}` is done correctly throughout.
- **The Lundberg citation** is properly included in `references.bib` and cited via `{cite:t}`.

drbenvincent (approved):

PR #687, opened 2026-01-24
Add zizmor GHA security linter and Dependabot

maresb:
## Summary

- Add zizmor GitHub Action to scan workflows for security vulnerabilities
- Add Dependabot configuration for automated GitHub Actions updates with 7-day cooldown
- Fix all zizmor security findings in existing workflows:
  - Add explicit permissions blocks to restrict GITHUB_TOKEN scope
  - Add `persist-credentials: false` to checkout actions
  - Fix spoofable bot actor check in label-precommit-prs.yml
  - Fix template injection in release.yml using environment variables
  - Replace `write-all` with scoped permissions in uml.yml

## Test plan

- [ ] Verify zizmor GHA runs successfully on this PR
- [ ] Verify CI workflow still passes
- [ ] Verify release workflow still works (manual test on next release)

----
📚 Documentation preview 📚: https://causalpy--687.org.readthedocs.build/en/687/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    cooldown:
+      default-days: 7
+    labels:
+      - "devops"
+      - "no releasenotes"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,7 @@ updates:
     labels:
       - "devops"
       - "no releasenotes"
+    groups:
+      actions:
+        patterns:
+          - "*"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,15 +6,21 @@ on:
   push:
     branches: [main]
 
+permissions: {}
+
 jobs:
   test:
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     strategy:
       matrix:
         python-version: ["3.11", "3.14"]
 
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - name: Set up Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,9 +14,9 @@ jobs:
         python-version: ["3.11", "3.14"]
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: ${{ matrix.python-version }}
       - name: Update pip and setuptools
@@ -34,7 +34,7 @@ jobs:
           python ./docs/source/.codespell/notebook_to_markdown.py --tempdir tmp_markdown
           codespell
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
         with:
           token: ${{ secrets.CODECOV_TOKEN }} # use token for more robust uploads
           name: ${{ matrix.python-version }}
--- a/.github/workflows/pr-issue-label-sync.yml
+++ b/.github/workflows/pr-issue-label-sync.yml
@@ -4,11 +4,14 @@ on:
   pull_request_target: # zizmor: ignore[dangerous-triggers]
     types: [opened, edited]
 
+permissions: {}
+
 jobs:
   sync:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
+      issues: read  # to read labels from linked issues
 
     steps:
       - name: Sync labels from closing issues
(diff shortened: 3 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -6,17 +6,23 @@ on:
...
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     strategy:
       matrix:
         python-version: ["3.11", "3.14"]
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
OriolAbril, line 21: I agree with using commits instead of tags which are mutable (most github provided actions don't seem to have the immutable releases option active 🤷🏿 ) or aliases like v6 which will mean different things as releases are published. However, the latest release for the checkout action is 6.0.1. The commit being pinned is a "random" one on main before being released. Given this will be updated automatically, can we check updates happen from release to release instead to happen with every commit to main (after a 7 day cooldown)?

I think it would be better to stick to actual releases (that is, pin to the commit from which the latest release tag was crafted)

Note: I can see there are a lot of options related to cooldown so I assume it is possible but I also haven't had time to look properly into all the options. If you already know it would be one think I can learn more easily but I can also look at the different options and see if I manage to get dependabot working like this if you agree with the dependency update policy.

.github/workflows/pr-issue-label-sync.yml
OriolAbril, line 13: Do we need the `permissions: {}` at the top level yaml to then override the specific permissions we want here or is it indifferent? I see a few actions are updated to use the pattern but some like this one aren't. I think this will mean the behaviour will depend on the org/repo settings regarding the GITHUB_TOKEN permissions so it is probably better to be explicit, but I am not completely sure about the differences or behaviour

.github/workflows/rtd-link-preview.yml
OriolAbril, line 1: This job should be deleted as it is deprecated. If there are a lot of checks and finding the PR preview link is different between all of them there is the option to add a comment to the PR through the RTD app. This is a setting managed through `readthedocs.org` dashboard.

Example of comment: https://github.com/arviz-devs/arviz-base/pull/135#issuecomment-3760058233

.github/dependabot.yml
OriolAbril, line 1: Given the number of workflows I would add some groupings to avoid the number of dependabot PRs to be excessive. I would probably start with the pattern below and we can create more focused groupings if we see there are too many different updates roped together.

```
  groups:
    actions:
      patterns:
        - "*"
```

.github/workflows/rtd-link-preview.yml
maresb: I opened https://github.com/pymc-labs/CausalPy/issues/688 for this. I'm not so familiar with RTD. Also I don't have access to the advanced settings, so I don't really know if the app is set up or if there's anything that needs to be done before deleting this workflow. Hence I figured it'd be easier to split this task into the separate issue.

.github/workflows/ci.yml
maresb: Hmm, this is actually a legit tagged release. It seems that they just didn't mark it as "Latest" for some reason:

https://github.com/actions/checkout/releases

I haven't looked in depth at the cooldown options. It just seems like 7 days is common, and seems reasonable to me.

.github/dependabot.yml
maresb: Oh, I didn't know about this. This is really cool, thanks!!!

.github/workflows/pr-issue-label-sync.yml
maresb: Good catch!

.github/workflows/rtd-link-preview.yml
OriolAbril: I don't have access to the rtd dashboard for causalpy either but it is only a matter of clicking on a checkmark: "Settings" (navbar) -> "Pull Request Builds" (left sidebar) -> "Show build overview in a comment" (checkmark in body)

maresb: > I don't have access to the rtd dashboard for causalpy either but it is only a matter of clicking on a checkmark: "Settings" (navbar) -> "Pull Request Builds" (left sidebar) -> "Show build overview in a comment" (checkmark in body)

@drbenvincent

.github/workflows/ci.yml
OriolAbril: This happens for only looking at "latest release" in the homepage sidebar instead of entering the releases page. Marking as resolved already, I also think 7 days is a good default

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,38 @@
...
+  pull_request:
+    branches: ["**"]
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: zizmor latest via PyPI
+    runs-on: ubuntu-latest
+    permissions:
```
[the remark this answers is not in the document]
maresb: This is wrong.

.github/workflows/release.yml
```diff
@@ -32,18 +37,24 @@ jobs:
...
           venv-bdist/bin/python -m pip install ../dist/causalpy*.whl
           echo "Checking import and version number (on release)"
-          venv-bdist/bin/python -c "import causalpy; assert causalpy.__version__ == '${{ github.ref_name }}' if '${{ github.ref_type }}' == 'tag' else causalpy.__version__; print(causalpy.__version__)"
+          venv-bdist/bin/python -c "import causalpy; assert causalpy.__version__ == '${REF_NAME}' if '${REF_TYPE}' == 'tag' else causalpy.__version__; print(causalpy.__version__)"
           cd ..
-      - uses: actions/upload-artifact@v4
+        env:
+          REF_NAME: ${{ github.ref_name }}
+          REF_TYPE: ${{ github.ref_type }}
+      - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
```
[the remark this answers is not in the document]
maresb: This is wrong.

.github/dependabot.yml
maresb: Done

.github/workflows/pr-issue-label-sync.yml
maresb: Done

OriolAbril (approved):

.github/workflows/rtd-link-preview.yml
drbenvincent: Done, I believe
Image

drbenvincent (approved):

PR #693, opened 2026-02-02
Avoid mutating sklearn estimators in DiD experiments

### LINKED ISSUE #664: DifferenceInDifferences mutates fit_intercept on user models
## Description

`DifferenceInDifferences` mutates sklearn models by forcing `fit_intercept=False` inside the experiment constructor. This is a hidden side effect: callers pass in a configured estimator, but the experiment overrides one of its key hyperparameters without explicit opt-in. This can be surprising in interactive workflows and makes it hard to reuse the same estimator instance across experiments because its configuration is modified in place.

## Steps to Reproduce

1. Create a sklearn regressor with `fit_intercept=True`.
2. Pass it into `DifferenceInDifferences`.
3. Observe that `fit_intercept` has been mutated to `False`.

```python
from sklearn.linear_model import LinearRegression
import causalpy as cp

model = LinearRegression(fit_intercept=True)
_ = cp.DifferenceInDifferences(..., model=model)
assert model.fit_intercept is False  # mutated by experiment
```

## Expected Behavior

Experiments should not mutate user-supplied models implicitly. If the experiment requires `fit_intercept=False` for correctness, it should either document this requirement, validate and raise, or create an internal copy of the estimator with the required settings.

## Actual Behavior

`DifferenceInDifferences` sets `model.fit_intercept = False` without warning or user consent.

## Environment

- Python version: 3.11+
- OS: macOS (darwin)

- Preferred direction: document and validate. Require `fit_intercept=False` and raise a clear error if not set.
- Clone the estimator (e.g., `sklearn.base.clone`) and adjust the copy instead of mutating the original.
- Move this behavior into `ScikitLearnAdaptor` or a dedicated wrapper so the mutation is explicit and centralized.

## Additional Context

Current code in `DifferenceInDifferences`:

```python
if hasattr(self.model, "fit_intercept"):
    self.model.fit_intercept = False
```

drbenvincent:
## Summary
- Avoid mutating user-supplied sklearn estimators by cloning when `fit_intercept=False` is required
- Emit a `UserWarning` when the clone occurs so users are aware and can fix their code
- Centralize `fit_intercept` handling in `BaseExperiment._ensure_sklearn_fit_intercept_false()`
- Add integration coverage to assert original estimators are unchanged and the warning fires
- Fix `codecov/patch` failure by ensuring full branch coverage of the new method

## Details

DiD experiments require `fit_intercept=False` because the design matrix already contains an explicit intercept column (`~ 1 + ...`). Previously, the experiment constructor mutated the user's model in-place (`model.fit_intercept = False`), which was a hidden side effect. Now the method clones the estimator via `sklearn.base.clone()`, sets the parameter on the copy, and issues a warning explaining why.

### Coverage fix

The `_ensure_sklearn_fit_intercept_false()` method in `base.py` has a defensive isinstance guard that returns early for non-sklearn models (line 70). Originally, the method was called from **inside** `elif isinstance(self.model, RegressorMixin):` blocks in both `diff_in_diff.py` and `staggered_did.py`, which made the guard unreachable — the caller had already confirmed the model was sklearn, so the isinstance check could never be `True`. This caused the `codecov/patch` check to fail.

The fix moves the `_ensure_sklearn_fit_intercept_false()` call to **before** the model-type dispatch (the `if PyMCModel / elif RegressorMixin` block) in both experiment classes. Since the method already has a built-in guard for non-sklearn models, calling it unconditionally is safe and semantically cleaner. Now any PyMC integration test naturally exercises the early-return path, providing full branch coverage without artificial unit tests.

A new integration test `test_did_sklearn_fit_intercept_false` was also added to cover the "no warning needed" path (when the model already has `fit_intercept=False`).

## Testing
```
MPLCONFIGDIR=/tmp/mplconfig XDG_CACHE_HOME=/tmp conda run -n CausalPy python -m pytest -o addopts= causalpy/tests/test_integration_skl_examples.py::test_did causalpy/tests/test_integration_skl_examples.py::test_did_sklearn_fit_intercept_false causalpy/tests/test_staggered_did.py::test_staggered_did_sklearn causalpy/tests/test_staggered_did.py::test_staggered_did_sklearn_model_without_fit_intercept
```

## Issues
- Closes #664

### DIFF
--- a/causalpy/experiments/base.py
+++ b/causalpy/experiments/base.py
@@ -18,6 +18,7 @@ Base class for quasi experimental designs.
 from __future__ import annotations
 
 import contextlib
+import warnings
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Literal
@@ -136,16 +137,30 @@ class BaseExperiment(ABC):
             raise ValueError("OLS models not supported.")
 
     def _ensure_sklearn_fit_intercept_false(self) -> None:
-        """Ensure scikit-learn models use fit_intercept=False without mutating user models."""
+        """Ensure scikit-learn models use ``fit_intercept=False`` without mutating
+        user-supplied estimators.
+
+        When the formula includes an explicit intercept (e.g. ``~ 1 + ...``), the
+        design matrix already contains an intercept column.  Letting sklearn *also*
+        fit its own intercept would double-count it and hide the intercept from the
+        reported coefficients.  This method detects the mismatch, clones the
+        estimator, and swaps in the corrected copy so the caller's original object
+        is never modified.
+        """
         if not isinstance(self.model, RegressorMixin):
             return
-        if not hasattr(self.model, "fit_intercept"):
-            return
-        if getattr(self.model, "fit_intercept", False) is False:
+        if not getattr(self.model, "fit_intercept", False):
             return
 
+        warnings.warn(
+            "fit_intercept=True was set on your estimator, but this experiment "
+            "requires fit_intercept=False (the design matrix already contains an "
+            "intercept column). A cloned copy with fit_intercept=False will be used.",
+            UserWarning,
+            stacklevel=3,
+        )
         try:
-            cloned_model = clone(self.model)
+            cloned = clone(self.model)
         except (
             Exception
         ) as exc:  # pragma: no cover - defensive for non-cloneable estimators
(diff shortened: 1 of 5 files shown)


drbenvincent:
Closing in favour of #942, which fixes the same issue (#664) with broader scope: centralises clone + `fit_intercept` handling in `BaseExperiment.__init__` and removes the mutation from all four affected experiment classes (`DifferenceInDifferences`, `StaggeredDifferenceInDifferences`, `PanelRegression`, `PiecewiseITS`), not just DiD/StaggeredDiD.

Useful patterns from this PR (lazy clone when `fit_intercept=True`, `set_params`, integration tests for the no-warning path) can be cherry-picked onto #942 if we want them — not blockers.

Thanks for the groundwork here.

PR #694, opened 2026-02-02
Enforce BaseExperiment abstract methods

### LINKED ISSUE #660: BaseExperiment abstract methods not enforced without ABC
## Description

`BaseExperiment` uses `@abstractmethod` decorators but does not inherit from `ABC`,
so Python does not enforce abstract methods. This allows instantiating incomplete
experiment subclasses without errors, which can lead to runtime failures later.

## Steps to Reproduce

1. Define a subclass of `BaseExperiment` without implementing abstract methods.
2. Instantiate the subclass.

## Expected Behavior

Instantiation should fail with a `TypeError` until all abstract methods are implemented.

## Actual Behavior

Instantiation succeeds because `BaseExperiment` does not inherit from `ABC`.

## Environment

- Python version: 3.11+
- OS: macOS (darwin)

Make `BaseExperiment` inherit from `abc.ABC` to enforce abstract methods.

## Additional Context

Relevant file: `causalpy/experiments/base.py`.

drbenvincent:
## Summary
- make BaseExperiment inherit from ABC and relax abstract requirements based on supports flags
- add explicit method stubs to experiments missing required abstract methods
- add a test to enforce abstract-method instantiation failures

## Testing
- conda run -n CausalPy python -m pytest causalpy/tests/test_base_experiment_abc.py (fails: coverage fail-under=80 when running single test)

Closes #660

### DIFF
--- a/causalpy/experiments/base.py
+++ b/causalpy/experiments/base.py
@@ -15,7 +15,7 @@
 Base class for quasi experimental designs.
 """
 
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from typing import Any, Literal
 
 import arviz as az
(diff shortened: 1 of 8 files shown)


cetagostini (approved):
All good here!

PR #695, opened 2026-02-02
Fix custom exceptions to preserve messages

### LINKED ISSUE #665: Custom exceptions drop message without super().__init__
## Description

Custom exceptions (`BadIndexException`, `FormulaException`, `DataException`) store `message` on the instance but do not call `Exception.__init__`. As a result, `str(exc)` and default exception rendering are empty, which makes error messages disappear in tracebacks and user-facing outputs that rely on the exception string. This is easy to miss because the message exists as an attribute, but standard exception behavior is effectively lost.

## Steps to Reproduce

1. Raise one of the custom exceptions with a message.
2. Print or log the exception object.

```python
from causalpy.custom_exceptions import DataException

try:
    raise DataException("Dataset is missing required columns")
except Exception as exc:
    print(str(exc))  # prints "" (empty string)
```

## Expected Behavior

`str(exc)` should return the message passed at construction, and tracebacks should show the message content.

## Actual Behavior

The exception message is empty because `Exception.__init__` is never called.

## Environment

- Python version: 3.11+
- OS: macOS (darwin)

- Preferred: call `super().__init__(message)` in each exception class and optionally keep `self.message` for backward compatibility. This preserves standard exception behavior (`str(exc)`, tracebacks) and is the smallest change.
- Alternative (simplest): remove custom `__init__` entirely and let the base `Exception` store the message; the class body can be empty with just a docstring.
- Alternative (back-compat focused): keep `self.message` but also override `__str__` to return it, in case any downstream code depends on that attribute (still less idiomatic than calling `super()` but explicit).
- Alternative (typing nicety): convert to `@dataclass` with a `message: str` field and call `Exception.__init__` in `__post_init__`; adds structure but more overhead than needed.
- Add a regression test to assert that `str(exc)` includes the message for each custom exception.

## Additional Context

Current implementation:

```python
class DataException(Exception):
    def __init__(self, message: str):
        self.message = message
```

drbenvincent:
## Summary
- Call `super().__init__(message)` in custom exceptions while preserving `self.message`.
- Add regression tests for exception string rendering.

## Testing
- `conda run -n CausalPy python -m pytest` (fails: PyTensor compiledir permission error)
- `PYTENSOR_FLAGS="compiledir=/tmp/pytensor" conda run -n CausalPy python -m pytest` (timed out at 120s)

Closes #665

### DIFF
--- a/causalpy/custom_exceptions.py
+++ b/causalpy/custom_exceptions.py
@@ -21,6 +21,7 @@ class BadIndexException(Exception):
     index and an event, typically a treatment or intervention."""
 
     def __init__(self, message: str):
+        super().__init__(message)
         self.message = message
 
 
@@ -29,6 +30,7 @@ class FormulaException(Exception):
     formula"""
 
     def __init__(self, message: str):
+        super().__init__(message)
         self.message = message
 
 
@@ -36,4 +38,5 @@ class DataException(Exception):
     """Exception raised given when there is some error in user-provided dataframe"""
 
     def __init__(self, message: str):
+        super().__init__(message)
         self.message = message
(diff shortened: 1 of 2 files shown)


drbenvincent:
bugbot review

PR #707, opened 2026-02-11
Add default PyMC models when model=None for all experiment classes

### LINKED ISSUE #661: Docs claim default models but BaseExperiment requires explicit model
## Description

Several experiment classes accept `model=None` in their signatures and explicitly
state in docstrings that a default PyMC model will be used when `model` is not
provided. In practice, this never happens: `BaseExperiment` immediately raises if
`model` is missing. The result is a mismatch between the documented API and actual
behavior that surfaces as a `ValueError` in the constructor, even in the most
common "getting started" usage where a user omits the model.

This is more than a doc nit: it prevents users from relying on advertised defaults,
and makes notebooks/examples that omit `model` brittle. It also causes the
constructor to fail before any input validation or helpful error messaging can run.

## Steps to Reproduce

1. Run an experiment without passing `model`.
2. Observe the error from `BaseExperiment`.

```python
import causalpy as cp
import pandas as pd

df = (
    cp.load_data("its")
    .assign(date=lambda x: pd.to_datetime(x["date"]))
    .set_index("date")
)
treatment_time = pd.to_datetime("2017-01-01")

# Docstring says this should default to a PyMC LinearRegression
cp.InterruptedTimeSeries(df, treatment_time, formula="y ~ 1 + t + C(month)")
```

## Expected Behavior

If `model=None`, the experiment should initialize a default PyMC model as described
in the docstrings (or the docstrings should state that a model is required).

## Actual Behavior

`BaseExperiment` raises immediately when `model` is missing.

```python
ValueError: model not set or passed.
```

## Environment

- Python version: 3.11+
- OS: macOS (darwin)

- Developer decision needed: align behavior with docs (defaults) or update docs/signatures.
- Option A: Update experiment constructors to set default models when `model is None`.
- Option B: Update docs and signatures to require an explicit model.

## Additional Context

Relevant code:

```python
if self.model is None:
    raise ValueError("model not set or passed.")
```

Docstring example (ITS):

```text
model : Union[PyMCModel, RegressorMixin], optional
    A PyMC (Bayesian) or sklearn (OLS) model. If None, defaults to a PyMC
    LinearRegression model.
```

There are likely other experiments with the same signature/doc pattern. A small
scan of experiments that accept `model=None` should confirm which docstrings need
alignment with the actual behavior (or where defaults should be implemented).

ErikRingen:
## Summary

Closes #661.

Experiment docstrings claimed default PyMC models would be used when `model` was omitted, but `BaseExperiment` immediately raised `ValueError("model not set or passed.")`. This aligns actual behavior with the documented API.

- Add `_default_model_class` class attribute to `BaseExperiment` with fallback logic that instantiates the default when `model is None`
- Set appropriate defaults on all 9 experiment subclasses: `LinearRegression` for most, `WeightedSumFitter` for `SyntheticControl`, `InstrumentalVariableRegression` for `InstrumentalVariable`, `PropensityScore` for `InversePropensityWeighting`
- Fix incorrect type hints (`BaseExperiment` -> `PyMCModel`) in `InstrumentalVariable`, `InversePropensityWeighting`, and `RegressionKink`
- Remove redundant `self.model = model` in `InstrumentalVariable` that would have overwritten the default set by `super().__init__()`

## Test plan

- [x] New parametrized integration test covers all 9 experiment classes with `model=None`
- [x] Explicit model precedence test confirms user-provided models are not overridden
- [x] Existing `test_model_experiment_compatability` tests pass (OLS rejection still works)
- [ ] Full CI integration test suite passes


drbenvincent (commented):
## Adversarial review: Default PyMC models when `model=None`

### What the PR does well
- Fixes the real bug: docstrings promised defaults but `BaseExperiment` raised when `model` was omitted.
- Single, clear mechanism: `_default_model_class` + fallback in base `__init__`.
- All nine experiment classes get an explicit default; type hints are corrected where they were wrong.
- Removing the redundant `self.model = model` in `InstrumentalVariable` is correct.
- Parametrized test covers every experiment with `model=None`; precedence test guards against overriding an explicit model.

---

### Critical / high-impact points

**1. `BaseExperiment` can raise `AttributeError` instead of `ValueError`**

If a subclass does **not** set `_default_model_class` and the user passes `model=None`, the base never assigns `self.model`. The code then hits `if isinstance(self.model, PyMCModel) ...` and raises `AttributeError: '...' object has no attribute 'model'` instead of the intended `ValueError("model not set or passed.")`.

**Recommendation:** Make the final check defensive, e.g. `if getattr(self, "model", None) is None: raise ValueError(...)` so that a future experiment that forgets `_default_model_class` still gets a clear, consistent error.

**2. Tests don’t assert that the default model is usable**

The new tests only check `isinstance(result.model, expected_model_class)` after construction. They never call `.fit()` (or any method that uses the model). So you don’t verify that the experiment’s `fit()` is actually compatible with the default model’s interface, or that there isn’t a bug in how the default is wired (e.g. wrong constructor or missing attributes).

**Recommendation:** Add at least one minimal integration test that constructs an experiment with `model=None` and calls `.fit()` (e.g. DiD with mocked sampling), and optionally asserts that fitting completes without error. That would catch wiring bugs that type checks alone miss.

**3. Default is always PyMC, never OLS**

For experiments with `supports_ols=True` (e.g. DiD, ITS, RD, SyntheticControl, StaggeredDiD), `model=None` still yields the default **PyMC** model (e.g. `LinearRegression()`), not an OLS/sklearn default. That’s consistent with “default PyMC model” in the docs but may surprise users who assume “no model” means “simple OLS”.

**Recommendation:** In the relevant experiment docstrings (or a shared note), state explicitly that `model=None` means “use the default **Bayesian (PyMC)** model” and that OLS requires passing an sklearn (or compatible) model. That avoids the “I didn’t pass a model but it’s still doing MCMC” confusion.

---

### Medium / consistency and scope

**4. `RegressionKink` docstring not updated**

Other experiments were updated to say “Defaults to LinearRegression” (or the appropriate class). `RegressionKink`’s docstring still doesn’t mention the default. The type hint was fixed to `PyMCModel | None`, but the narrative doc is now inconsistent.

**Recommendation:** Add a short note in `RegressionKink`’s docstring that the model defaults to `LinearRegression` when omitted.

**5. Interrogate badge SVG in this PR**

The diff includes `docs/source/_static/interrogate_badge.svg` (96.9% → 97.0%). That’s a side effect of docstring edits, not of the default-model behavior.

**Recommendation:** Either drop that file from this PR and let a docs/coverage pass update it, or add a one-line note in the PR description so reviewers know it’s intentional. Keeps the PR focused and avoids “why is the badge changing?” noise.

**6. Future subclasses must remember `_default_model_class`**

The design assumes every concrete experiment sets `_default_model_class`. There’s no abstract requirement or runtime check. A new experiment that forgets it will trigger the `AttributeError` (or, after the suggested fix, the `ValueError`).

**Recommendation:** Either document in `BaseExperiment` (or a contributing guide) that subclasses must set `_default_model_class` if they want `model=None` to work, or consider a pattern that makes “no default” explicit (e.g. subclasses that don’t want a default set `_default_model_class = None` in a commented or documented way). The `getattr` fix above at least keeps the error message consistent.

---

### Lower priority

**7. Release notes**

Changing “model required” to “model optional with a default” is a visible API/UX change. Some users might have relied on the previous `ValueError` to catch missing model arguments.

**Recommendation:** In release notes, briefly state that `model` is now optional for all experiment classes and that the documented default PyMC model is used when it’s omitted.

**8. Param doc style**

Some experiments use a `Parameters` section, others use `:param model:`. The PR updates content but not style. Minor; could be cleaned up in a docs-only pass.

---

### Summary

The fix is correct and the design is simple. The main improvements to consider before or shortly after merge:

1. Use `getattr(self, "model", None)` in the final “model not set” check so subclasses without a default get `ValueError` instead of `AttributeError`.
2. Add at least one test that runs `.fit()` with `model=None` to validate the default-model wiring.
3. Document that `model=None` means “default **Bayesian (PyMC)** model” and that OLS requires an explicit sklearn (or compatible) model.
4. Align `RegressionKink` docstring with the others and optionally trim or explain the interrogate badge change.

drbenvincent (approved):

drbenvincent:
Just need to fix the failing tests @ErikRingen and we are good to merge :)

drbenvincent:
## CI doctest failures fixed

The two failing doctests were:
- `causalpy.experiments.instrumental_variable.InstrumentalVariable`
- `causalpy.pymc_models.InstrumentalVariableRegression`

### Root cause

This was **not** a CausalPy bug. numpy 2.4 changed the internal format of `np.einsum_path(..., einsum_call=True)` from 5-element tuples to 3-element tuples. PyTensor 2.37.0 expects 5 elements in its `einsum()` function (Case 2: known shapes path), so `pm.LKJCholeskyCov()` — used by `InstrumentalVariableRegression.build_model()` — crashes with:

```
ValueError: not enough values to unpack (expected 5, got 3)
```

Locally (conda), numpy 2.3.5 is installed and everything works. On CI (pip), numpy 2.4.x is resolved and triggers the incompatibility.

### Fix

Pinned `numpy<2.4` in `pyproject.toml` (commit 19678f00). This is consistent with the broader ecosystem (e.g., numba also requires `numpy<2.4`).

### Upstream issue

Filed: https://github.com/pymc-devs/pytensor/issues/1892

The pin can be lifted once PyTensor releases a version that handles numpy 2.4's new `einsum_path` format.

PR #731, opened 2026-02-21
Support datetime treatment_time in SyntheticControl plots

### LINKED ISSUE #725: Support datetime treatment_time in SyntheticControl plots
## Description

The `SyntheticControl._bayesian_plot()` method has a TODO noting that `treatment_time` does not fully work when it is a datetime value (line 532 of `experiments/synthetic_control.py`). The `axvline` call works with numeric values, but datetime-aware plotting requires conversion or special handling.

## Current behavior

The `axvline(x=self.treatment_time)` call works for numeric `treatment_time` but may fail or produce incorrect results when `treatment_time` is a `pd.Timestamp` or `datetime` object.

## Expected behavior

The intervention line should render correctly regardless of whether `treatment_time` is numeric or datetime.

## Context

Triaged from #666. The existing datetime formatting logic (added later in the method) handles axis tick labels, but the `axvline` call itself may need adjustment for datetime values.

drbenvincent:
## Summary
- convert `SyntheticControl` intervention line positions through each axis unit converter so datetime and numeric `treatment_time` values both render correctly
- apply the fix in both Bayesian and OLS plotting paths
- add an sklearn integration regression test for datetime-index synthetic control plotting

Fixes #725

## Test plan
- [x] `CONDA_EXE=$(for c in mamba micromamba conda; do command -v \"$c\" >/dev/null 2>&1 && echo \"$c\" && break; done) && \"$CONDA_EXE\" run -n CausalPy pytest --no-cov causalpy/tests/test_integration_skl_examples.py -k \"test_sc or test_sc_datetime_treatment_time_plot\"`
- [x] `CONDA_EXE=$(for c in mamba micromamba conda; do command -v \"$c\" >/dev/null 2>&1 && echo \"$c\" && break; done) && \"$CONDA_EXE\" run -n CausalPy pre-commit run --all-files`

### DIFF
--- a/causalpy/experiments/synthetic_control.py
+++ b/causalpy/experiments/synthetic_control.py
@@ -281,6 +281,18 @@ class SyntheticControl(BaseExperiment):
             print(f"Treated unit: {self.treated_units[0]}")
         self.print_coefficients(round_to)
 
+    @staticmethod
+    def _convert_treatment_time_for_axis(
+        axis: plt.Axes, treatment_time: int | float | pd.Timestamp
+    ) -> int | float | pd.Timestamp:
+        """
+        Convert treatment time into the plotting units expected by a specific axis.
+        """
+        try:
+            return axis.xaxis.convert_units(treatment_time)
+        except (TypeError, ValueError):
+            return treatment_time
+
     def _bayesian_plot(
         self,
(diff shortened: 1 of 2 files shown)


drbenvincent:
bugbot review

PR #741, opened 2026-02-27
Add knowledgebase guide for custom PyMC models

### LINKED ISSUE #740: Allow custom PyMC Models in Causalpy
I see a churn of users not using causalpy because they want to build their own models. Could `Causalpy` allow a simple way to make user pass their own pymc models and use all machinery from causalpy under the hood?

Things like:

- [ ] Plot suite.
- [ ] Effects tabular results.
- [ ] Idata objects

If we allow this, I feel more people could engage. On top, will allow to incorporate more PyMC Marketing because we'll have the same modular behavior.

Note of worry: People who reach with this concerns usually are doing very crazy wrong models, which give me nerve to allow people to add any random models, because we'll open to door to do wrong models and because output looks nice, they might thing are correct.

drbenvincent:
## Summary

Closes #740

- Adds a knowledgebase notebook (`docs/source/knowledgebase/custom_pymc_models.ipynb`) documenting how to customize PyMC models in CausalPy
- Documents **two levels** of customization:
  1. **Prior-based** (no subclassing) — change parameter priors and the likelihood distribution via the `Prior` class. References the existing Student-t example in `its_pymc_comparative.ipynb`.
  2. **Subclassing `PyMCModel`** — for structural changes like link functions, latent variables, or custom error structure
- Documents the **PyMCModel naming contract** (required variable and dimension names) as a reference table — this was previously undocumented outside docstring examples
- Includes a **worked `LogLinearRegression` example** (`mu = exp(X @ beta)`) for positive-valued outcomes, demonstrating the subclassing pattern with an ITS analysis on simulated revenue data

## Rough edges noted for follow-up

While researching this, I identified some rough edges that could be addressed in future PRs:
- Variable names (`"X"`, `"y"`, `"mu"`, `"y_hat"`, `"beta"`) are hardcoded strings — could be class attributes to make them overridable
- No validation after `build_model()` — violating the naming contract produces cryptic `KeyError`s rather than helpful messages
- `print_coefficients()` assumes the model has `"beta"` — custom models with different parameterizations silently break

## Test plan

- [x] Notebook executes successfully (MCMC sampling converges, no divergences)
- [x] `pre-commit run --all-files` passes
- [x] Summary, coefficients, and plots render correctly


cetagostini (approved):
Good, this is a great introduction and should help even to people to understand how to make new models in the API

PR #781, opened 2026-03-04
Remove numpy<2.4 pin now that pytensor fix is released

### LINKED ISSUE #723: Remove numpy<2.4 pin once pytensor fixes numpy 2.4 compatibility
## Context

In PR #707 (commit 19678f00) we pinned `numpy<2.4` in `pyproject.toml` because numpy 2.4 changed the internal format of `np.einsum_path(..., einsum_call=True)`, breaking pytensor's `einsum()` function. This causes `pm.LKJCholeskyCov` (used by `InstrumentalVariableRegression`) to crash.

Upstream issue: https://github.com/pymc-devs/pytensor/issues/1892

## Action needed

Once the pytensor issue is resolved and a fixed version is released:

1. Remove the `numpy<2.4` upper pin from `pyproject.toml`
2. Optionally add a `pytensor>=<fixed_version>` minimum pin if needed
3. Verify CI doctests pass with numpy >= 2.4

## How to check

- Watch https://github.com/pymc-devs/pytensor/issues/1892 for resolution
- Test with: `pip install numpy>=2.4 pytensor>=<new_version>` and run `make doctest`

drbenvincent:
## Summary
- remove the temporary `numpy<2.4` upper bound and require `pytensor>=2.38.1`, now that the upstream NumPy 2.4 compatibility fix is available
- keep `environment.yml` aligned with `pyproject.toml` for these dependency requirements
- fix downstream runtime breakages triggered by NumPy 2.4 stricter scalar conversion in reporting and plotting code paths
- make scalar extraction more robust by converting scalar-like xarray outputs at API boundaries (instead of relying on fragile array-to-float coercion)
- add regression coverage for singleton-dimension xarray cases to prevent reintroducing these failures

## Why this change
- the dependency pin can now be removed safely because PyTensor has shipped the needed compatibility fix
- once the pin is removed, CI and notebook execution exposed shape-related scalar conversion failures (`TypeError: only 0-dimensional arrays can be converted to Python scalars`)
- this PR addresses both parts together: dependency unpinning and hardening call sites so behavior remains stable on newer NumPy versions

## Files changed
- `pyproject.toml`
- `environment.yml`
- `causalpy/reporting.py`
- `causalpy/experiments/diff_in_diff.py`
- `causalpy/tests/test_reporting.py`

## Test plan
- [x] `$CONDA_EXE run -n CausalPy prek run --all-files`
- [x] `$CONDA_EXE run -n CausalPy pytest causalpy/tests/test_reporting.py -q --no-cov`
- [x] `$CONDA_EXE run -n CausalPy pytest causalpy/tests/test_integration_skl_examples.py::test_did causalpy/tests/test_integration_pymc_examples.py::test_did -q --no-cov`

Fixes #723.


causalpy/reporting.py
```diff
@@ -56,6 +56,11 @@ class EffectSummary:
 # ==============================================================================
 
 
+def _as_float(value) -> float:
+    """Convert scalar-like values (including singleton arrays) to Python float."""
+    return float(np.asarray(value).reshape(()))
```
juanitorduz: There should be in utils.py or somehwere. They have to be in the same place as `_as_scalar` :)

juanitorduz (dismissed):

causalpy/reporting.py
drbenvincent: Good catch — we had the same helper defined in two places (`_as_float` here and `_as_scalar` in `diff_in_diff.py`). Consolidated both into a single `_as_scalar` in `causalpy/utils.py` and updated all call sites + the test. Pushed in f812477.

pyproject.toml
```diff
@@ -43,10 +43,11 @@ dependencies = [
     "graphviz",
     "ipython!=8.7.0",
     "matplotlib>=3.5.3",
-    "numpy<2.4",
+    "numpy",
```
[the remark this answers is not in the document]
drbenvincent: Thanks for flagging this. I looked into it and the ITS patterns are actually safe — the data there has dims `["chain", "draw", "obs_ind"]` with no `treated_units` dimension, so `.mean(dim=["chain", "draw"])` collapses to a true 0-d array and `float()` works fine even on NumPy 2.4+.

That said, it would be good defensive practice to convert those remaining `float(x.values)` calls in `interrupted_time_series.py` and `piecewise_its.py` to `_as_scalar()` as well — for consistency and future-proofing in case those data shapes ever change. Not a blocker, but worth doing.

ricardoV94 (approved):

PR #794, opened 2026-03-07
Fix simulate_data.py reproducibility + remove synthetic CSVs   

### LINKED ISSUE #545: Fix reproducibility, refactor `simulate_data.py`, use functions in tests
## Problem

  Seems like `causalpy/data/simulate_data.py` module has some reproducibility issues and a bit of refactoring that could be done

  ### Reproducibility

  The module declares a seeded RNG but doesn't use it consistently:

  ```python
  rng = np.random.default_rng(RANDOM_SEED)  # Declared on line 27, only used once

  # Most functions use unseeded random:
  norm(0, 0.25).rvs(N)           # scipy.stats uses global numpy state
  np.random.choice(2, size=N)     # Uses global numpy state
```

  Result: Functions produce different data each run. Generated CSV files cannot be reproduced.

  CSV Usage:

  - Many CSVs committed to git
  - Cannot regenerate them deterministically

## Proposed Solution

1. Add seed parameter to all generation functions
2. Replace norm().rvs() with rng.normal(), dirichlet().rvs() with rng.dirichlet(), etc.
3. Delete generated CSV files; use pytest fixtures instead
4. Update tests to generate data dynamically
5. Fix bug: create_series() ignores length_scale parameter (line 488)
6. Reduce duplication (lines 87-93: repeated function calls)
7. Other light touch refactoring (separation of responsibility in functions, reduce LOC on `_smoothed_gaussian_random_walk` (lines 87-5))

louismagowan:
Closes #545

## Summary

Refactors synthetic dataset generation to be reproducible and fully programmatic, removing the dependency on committed CSV files.

## What changed

**Bug fixes**
- All synthetic data generators now accept a `seed` parameter and use a local `numpy.Generator` instance instead of global `np.random.*` state — results are now reproducible
- Fixed `create_series()` bug where `length_scale` parameter was ignored (hardcoded to `2`)
- Fixed deprecated pandas `freq="M"` → `freq="ME"`

**Refactor**
- `load_data()` now generates synthetic datasets on-the-fly using a seeded generator rather than reading from CSV; real-world datasets (banks, brexit, etc.) still load from shipped CSVs
- Split `DATASETS` dict into `SYNTHETIC_DATASETS` and `REAL_WORLD_DATASETS`
- Removed circular `import causalpy as cp` dependency in `datasets.py`

**Deleted files**
- 8 synthetic CSVs removed (~172KB): `did.csv`, `regression_discontinuity.csv`, `synthetic_control.csv`, `its.csv`, `its_simple.csv`, `ancova_generated.csv`, `geolift1.csv`, `geolift_multi_cell.csv`
- `gt_social_media_data.csv` removed (was never referenced in code)

**API changes** _(low risk — not imported via `__init__.py`)_
- `generate_seasonality`, `periodic_kernel`, `create_series` renamed to `_generate_seasonality`, `_periodic_kernel`, `_create_series` (internal helpers)
- `generate_time_series_data` deleted (superseded by `generate_time_series_data_seasonal` and `generate_time_series_data_simple`)

**Tests**
- Added session-scoped fixtures for all 8 synthetic datasets — generated once per test session, avoiding redundant calls
- Updated 8 test files to use fixtures instead of `cp.load_data()`
- Rewrote `test_data_loading.py` to parametrize over all datasets with reproducibility and unknown-key tests

## Notes for users

Calls to `load_data()` for synthetic datasets will now return different numeric values than before (expected — this is the fix). Users who pinned to specific values from the old CSVs will need to update their code.


drbenvincent (changes requested):
Thanks for this contribution, @louismagowan — nice work. The commit structure is clean, the approach is well thought-out, and it's great to see all CI checks green with coverage actually improving (+0.55%).

A few notes:

### API breakages to be aware of

This PR removes or renames several functions that were previously accessible (and documented in the API docs):

- `generate_seasonality`, `periodic_kernel`, `create_series` are renamed to private (`_` prefix)
- `generate_time_series_data` is deleted entirely

None of these are imported via `__init__.py`, so they're not top-level public API, but anyone doing `from causalpy.data.simulate_data import generate_seasonality` (etc.) would break. The risk is low, but we should mention these in the release notes.

### `load_data()` is now dynamic — no caching

Previously `load_data("sc")` read a small CSV. Now it runs full generation (LOWESS smoothing, GP kernels, etc.) on every call. The test suite handles this well with session-scoped fixtures, but users calling `load_data()` in a loop will see a performance hit. Not a blocker — just something to be aware of. We could add `@functools.lru_cache` or similar in a follow-up if it becomes an issue.

### Data values will change

This is inherent to the fix (and expected), but worth calling out: `load_data("did")` and friends now return completely different numeric values than the old CSVs. Any downstream user code that depended on specific values will see different results.

### `git-filter-repo` suggestion

Appreciate the thought here, but I'd recommend against it. The savings (~172KB) are negligible and the disruption is severe — all commit hashes rewritten, all open PRs need rebasing, all contributors need to re-clone, external SHA links break. Please do raise this as a separate issue though so we have it tracked; it just won't be high priority given the trade-offs.

### PR description

Could you flesh out the PR body with a summary of what changed and why? For a 20-file refactor like this, a brief bullet list (API changes, deleted files, new test fixtures, bug fixes) makes it much easier for maintainers and users browsing the changelog to understand the scope. The commit messages are well-structured — it's really just about surfacing that same information in the PR description so it's visible at a glance.

drbenvincent (approved):

PR #796, opened 2026-03-07
Clean up experiments/: typos, annotations, and constants

louismagowan:
Summary

Purely mechanical cleanup of causalpy/experiments/ — no behavioural changes.
Closes: https://github.com/pymc-labs/CausalPy/issues/795

- Fix typos: "treament" → "treatment" (diff_in_diff), "trestment" → "treatment" (prepostnegd)
- Fix **kwargs: dict → **kwargs: Any: 9 experiment files — dict is the wrong annotation for **kwargs values
- Replace is False with not: idiomatic boolean checks in dummy-coded validation (3 files)
- Fix misplaced docstring: DifferenceInDifferences.input_validation() had its docstring after the first executable line
- Standardize expt_type to instance attribute: remove dead class attribute in InterruptedTimeSeries, move to __init__ in PiecewiseITS
  - Extract LEGEND_FONT_SIZE to causalpy/experiments/constants.py (8 files)
  - Extract HDI_PROB to causalpy/utils.py, replace all hardcoded 0.94 defaults and 0.03/1 - 0.03 quantile bounds (10 files)

### DIFF
--- a/causalpy/utils.py
+++ b/causalpy/utils.py
@@ -20,8 +20,6 @@ from __future__ import annotations
 import re
 from typing import TYPE_CHECKING, Any, Literal
 
-HDI_PROB: float = 0.94
-
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
@@ -30,6 +28,8 @@ import xarray as xr
 
 if TYPE_CHECKING:
     from causalpy.experiments.synthetic_control import SyntheticControl
+
+HDI_PROB: float = 0.94
 
 
 def _is_variable_dummy_coded(series: pd.Series) -> bool:
@@ -108,7 +108,9 @@ def convert_to_string(x: float | xr.DataArray, round_to: int | None = 2) -> str:
         return f"{x:.2f}"
     elif isinstance(x, xr.DataArray):
         # In the case of an xarray object, we return the mean and 94% CI
-        percentiles = x.quantile([(1 - HDI_PROB) / 2, 1 - (1 - HDI_PROB) / 2]).to_numpy()
+        percentiles = x.quantile(
+            [(1 - HDI_PROB) / 2, 1 - (1 - HDI_PROB) / 2]
+        ).to_numpy()
         ci = (
             r"$CI_{94\%}$"
             + f"[{round_num(percentiles[0], round_to)}, {round_num(percentiles[1], round_to)}]"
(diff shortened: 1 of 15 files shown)


causalpy/utils.py
```diff
@@ -29,6 +29,8 @@
 if TYPE_CHECKING:
     from causalpy.experiments.synthetic_control import SyntheticControl
 
+HDI_PROB: float = 0.94
```
juanitorduz: Shall we move this ont into the new `constants.py` file? WDYT @drbenvincent ?

juanitorduz (commented):
Thanks @louismagowan ! This looks great. I just let a minor question :)

drbenvincent (changes requested):
Thanks for the cleanup — this is thorough and well-structured. All CI green, coverage maintained. A couple of things to address:

## 1. Hardcoded "94%" label strings should derive from `HDI_PROB`

The quantile *computations* now correctly use `HDI_PROB`, but the *display labels* are still hardcoded as `"94%"`. This means the constant extraction is half-done — if someone changes `HDI_PROB`, the math updates but the labels would be wrong.

Locations:
- `pymc_models.py:420` — `"94% HDI [...]"`
- `utils.py:111` — `r"$CI_{94\%}$"`
- `regression_kink.py:279` — `r"$CI_{94\%}$"`
- `regression_discontinuity.py:338` — `r"$CI_{94\%}$"`
- `prepostnegd.py:219` — `r"$CI_{94%}$"` (also note: missing `\` before `%`, unlike the other files)

These should derive from the constant, e.g. `f"{HDI_PROB*100:.0f}% HDI"` and `rf"$CI_{{{HDI_PROB*100:.0f}\%}}$"`.

## 2. Agree with @juanitorduz — move `HDI_PROB` into `constants.py`

Having constants split across `experiments/constants.py` (for `LEGEND_FONT_SIZE`) and `utils.py` (for `HDI_PROB`) is a bit scattered. Since `HDI_PROB` is used outside `experiments/` too (`utils.py`, `plot_utils.py`, `pymc_models.py`), a top-level `causalpy/constants.py` might be the cleanest home for both constants. That way there's one obvious place to look.

drbenvincent (approved):

drbenvincent:
Thanks for this @louismagowan !

PR #808, opened 2026-04-01
Clarify reuse-first Python environment guidance

drbenvincent:
## Summary
- clarify in `AGENTS.md` and the `python-environment` skill that agents should reuse an existing `CausalPy` environment and only create or update one when needed
- explain when the project environment is actually required versus when plain `python` is enough for simple local inspection helpers
- add a `run -p <full-prefix>` fallback plus guidance for editable installs in git worktrees and reuse on persistent remote machines

This reduces unnecessary setup work and token/time overhead while preserving the existing guidance for tasks that truly need the project environment.

## Test plan
- [x] `mamba run -p "/Users/benjamv/mambaforge/envs/CausalPy" prek run --all-files`

### DIFF
--- a/.github/skills/python-environment/SKILL.md
+++ b/.github/skills/python-environment/SKILL.md
@@ -1,11 +1,31 @@
 ---
 name: python-environment
-description: Detect and configure a conda-compatible tool, create the CausalPy environment, and run commands inside it. Use before any task that requires Python execution.
+description: Detect and configure a conda-compatible tool, reuse an existing CausalPy environment when possible, create it only when needed, and run commands inside it. Use before tasks that need the project environment, such as importing project code, running tests, building docs, or invoking repo tooling.
 ---
 
 # Python Environment
 
 Set up and run commands inside the CausalPy conda environment.
+
+## Decide whether the env is required
+
+Use the `CausalPy` env when the command:
+
+- imports project code (for example `import causalpy` or project modules)
+- runs tests
+- builds docs
+- invokes repo tooling such as `make`, `prek`, or notebook execution
+
+For simple inspection helpers that only read local text/JSON or use the Python standard library, any Python on `PATH` is acceptable.
+
+## Reuse before creating
+
+Do the least work that will get the task done:
+
+1. Reuse an existing `CausalPy` env if one is already available.
+2. If `run -n CausalPy` cannot resolve the env, check whether it exists under a different prefix and use `run -p`.
+3. Only create the env if no suitable existing env is available.
+4. Only update the env or rerun `make setup` when dependencies changed, the editable install is stale, or the current checkout has not been installed into that env yet.
 
 ## Detect the conda tool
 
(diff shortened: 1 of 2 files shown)


.github/skills/python-environment/SKILL.md
maresb:
```suggestion
Never use `$CONDA_EXE activate`, instead use `$CONDA_EXE run -n CausalPy <command>`.
```

.github/skills/python-environment/SKILL.md
```diff
@@ -1,12 +1,32 @@
 ---
 name: python-environment
-description: Detect and configure a conda-compatible tool, create the CausalPy environment, and run commands inside it. Use before any task that requires Python execution.
+description: Detect and configure a conda-compatible tool, reuse an existing CausalPy environment when possible, create it only when needed, and run commands inside it. Use before tasks that need the project environment, such as importing project code, running tests, building docs, or invoking repo tooling.
```
maresb: Seems overly verbose.

```suggestion
description: Detect and configure, and use a conda-compatible tool. Use before tasks that need the project environment, such as importing project code, running tests, building docs, or invoking repo tooling.
```

.github/skills/python-environment/SKILL.md
```diff
@@ -54,6 +78,27 @@ $CONDA_EXE env update --file environment.yml --prune
...
+
+```bash
+$CONDA_EXE env list
+$CONDA_EXE run -p "/full/path/to/CausalPy" <command>
+```
+
+Keep using `run -p` with that full prefix for the rest of the session.
+
+Note: a shell `conda` function may still resolve through the same libmamba installation as `mamba`, so switching from `mamba` to `conda` does not necessarily change which env store is searched.
+
```
maresb: I don't understand what this is trying to convey. Is it necessary?

maresb (commented):
Looks good, just a few suggestions.

.github/skills/python-environment/SKILL.md
drbenvincent: Good call, the one-liner doesn't need to carry the reuse/create nuance — that's covered in the body. Applying with a small grammar tweak (comma placement):

```
description: Detect, configure, and use a conda-compatible tool. Use before tasks that need the project environment, such as importing project code, running tests, building docs, or invoking repo tooling.
```

.github/skills/python-environment/SKILL.md
drbenvincent: Agreed — `AGENTS.md` already says "never use `$CONDA_EXE activate`", so the skill file should match. Applied.

.github/skills/python-environment/SKILL.md
drbenvincent: The intent was to explain why switching from `mamba` to `conda` (a natural troubleshooting step) often doesn't help resolve a missing env — they can share the same libmamba backend and env store. But you're right, it's not necessary. The actionable fix is the `run -p` fallback above it, and that's already clear. Removing.

drbenvincent:
Thanks @maresb. Everything done in https://github.com/pymc-labs/CausalPy/pull/808/commits/a2da1def277618392f55ed71b2f3c61957ecedf9

maresb (approved):
Looks great!

PR #817, opened 2026-04-02
Add sub-issue discovery to github-issues skill

drbenvincent:
## Summary

Add sub-issue discovery guidance to the github-issues skill so agents check for native GitHub sub-issues when evaluating issues.

- **issue-evaluation.md**: Add a "Check for sub-issues" step using the `subIssues` / `parent` GraphQL fields, and note to review sub-issues during context analysis.
- **parent-child-issues.md**: Add a "Discover existing sub-issues" section (step 6) with the correct GraphQL query and a warning not to use the older `trackedIssues` / `trackedInIssues` fields.

## Motivation

GitHub's native sub-issues use the `subIssues` GraphQL field, not `trackedIssues`. Without this guidance, agents default to the wrong API and silently miss existing sub-issues on a parent issue.

## Test plan

- [x] Skill files are valid markdown
- [x] No code changes; documentation-only

### DIFF
--- a/.github/skills/github-issues/reference/issue-evaluation.md
+++ b/.github/skills/github-issues/reference/issue-evaluation.md
@@ -10,10 +10,26 @@ description: Analyze an issue, check current relevance, and propose next steps o
 gh issue view <issue_number> --json number,title,body,state,labels,comments,assignees
 ```
 
+## Check for sub-issues (native hierarchy)
+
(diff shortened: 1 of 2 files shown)


drbenvincent:
@cetagostini could we review/merge this one (and #808) before I review your #837 ? Logic is that my 2 are open agent PR's and yours looks to try to unify things, so makes sense to do that after?

maresb (approved):
Sorry, just now seeing this.

Also just now learning about subissues! I think I should be using them.

Regarding this PR, this looks very benign and I think you should just merge it.

PR #819, opened 2026-04-02
Add geo-experiment design workflow and sensitivity check plotting for Synthetic Control

drbenvincent:
## Summary

Adds prospective experiment-design capabilities and first-class sensitivity check plotting to `SyntheticControl`.

**Design-phase methods** — so practitioners can assess whether a geo-experiment will work before committing budget:

- `SyntheticControl.from_pre_period()` creates a design-phase instance from pre-period data only
- `validate_design()` — dress rehearsal: injects a known effect and checks recovery
- `power_analysis()` — simulation-based Bayesian power curve
- `donor_pool_quality()` — composite quality score (correlation, convex hull, weight concentration)
- `DressRehearsalCheck` wraps dress rehearsal as a `Check` for pipeline integration
- Result classes (`DressRehearsalResult`, `PowerCurveResult`, `DonorPoolQualityResult`) with `plot()` and `summary()` methods

**Sensitivity check plotting** — previously, check visualizations lived as ~80 lines of custom matplotlib in the notebook. Now they are part of the library:

- New `causalpy/checks/_plot_helpers.py` with shared `forest_plot()` and `null_distribution_plot()` helpers
- `plot()` staticmethods on `PlaceboInSpace`, `PlaceboInTime`, `LeaveOneOut`, and `PriorSensitivity`
- Each check's `run()` auto-populates `CheckResult.figures` with matplotlib figures
- `GenerateReport` now renders check figures in the HTML report (base64-encoded PNGs)
- Notebook custom plot cells replaced with single-line library calls (e.g. `PlaceboInSpace.plot(result, baseline_stats=stats)`)

**Notebook overhaul** (`sc_pymc.ipynb`):

- Switched to the California Proposition 99 dataset — the canonical SC example
- Restructured as a full workflow: design assessment before analysis, robustness checks after
- Literature-grounded narrative with citations to Abadie (2003, 2010, 2015, 2021), Athey (2017), Brodersen (2015)

## Test plan

- [x] 27 integration tests pass (`test_sc_design.py`)
- [x] 13 unit tests for check plotting (`test_check_plots.py`)
- [x] All prek checks pass (ruff, mypy, codespell, notebook schema) — `interrogate` failure is pre-existing (84% vs 85%)
- [ ] Verify notebook renders correctly via `make html`
- [ ] Run full test suite to check for regressions


drbenvincent:
## Question on the noise-injection step in `power_analysis()`

Flagging this for a closer look before we merge — I want to make sure we're on solid methodological ground here.

Quick recap of what the algorithm does, so the question is concrete. For each candidate effect size, and for each simulation within it, `power_analysis()` does roughly this: it takes a copy of the pre-period data, draws a fresh vector of Gaussian noise scaled by the residual standard deviation from the real pre-period fit, adds that noise to the treated unit's pre-period values, builds a new design-phase SC on the noisy data, runs `validate_design()` at the current effect size, and records whether the chosen detection criterion fires. Tally detections, divide by `n_simulations`, repeat across the effect-size grid, and that's the power curve.

The bit I want to sanity-check is the noise injection itself. It's the only source of simulation-to-simulation variability — the injected effect is deterministic, the donor pool is fixed, the model is fixed — so whatever statistical properties we attribute to the power curve are inherited entirely from that step. Two specific worries:

1. **Is the noise model right?** We're drawing iid Gaussian noise with `sigma = residual_std`, where `residual_std` is the std of the pre-period residuals' posterior-mean trajectory. That's a single scalar applied uniformly across time, treating residuals as homoscedastic and independent. For most real treated time-series — including Prop 99 — pre-period residuals are autocorrelated and often heteroscedastic. An iid Gaussian draw will under-represent the kind of variation we actually expect to see in a fresh realisation of the same data-generating process, which would tend to make the power curve look tighter / more optimistic than it should.

2. **Is sampling-variation-on-the-treated-unit-only the right notion of "a fresh experiment" for SC?** In a frequentist power calculation we'd usually resample from the assumed DGP. Here we're perturbing the observed treated trajectory while holding the donor pool fixed at its single realised path. That's a defensible choice (donors are the "design", treated is the "outcome"), but it's not obviously the same thing as sampling-distribution-style power, and I haven't found a clean reference in the synthetic-control literature that endorses exactly this resampling scheme. Abadie-style inference uses placebo-in-space permutations rather than parametric noise injection; the Bayesian-power literature (e.g. Kruschke-style assurance, design prior approaches) typically simulates from the full posterior predictive rather than perturbing one series.

Concrete things I'd like us to check before this lands:

- Is there a paper / standard reference we can cite for this specific resampling scheme? If yes, let's add the citation in the docstring and the notebook. If no, we should be explicit in the docs that this is a heuristic and describe its limitations.
- Should the noise model at least respect pre-period autocorrelation (e.g. block bootstrap of residuals, or an AR(1) draw calibrated from the pre-period residuals) rather than iid Gaussian?
- Should we offer an alternative variability source — e.g. resampling from the posterior predictive of the design-phase fit — as a non-default option, so users can compare?
- At minimum: does the docstring make it clear what assumption the iid-Gaussian-on-treated step is making, and what kinds of mis-specification will bias the resulting power curve?

Happy to take the lead on any of the above if we agree it's worth doing in this PR rather than as a follow-up. My instinct is that the first bullet (find a citation or be honest that there isn't one) is the minimum bar for landing, and the others can become a follow-up issue.

cetagostini:
Hey, amazing work and dig here, I had this idea many times before and kinda like it because sound intuitive but, I guess this is re-implementing an existing capability already in CausalPy.

Placebo in time was build with the goal of run a "power analysis" (bayesian assurance - power is more freq term). The approach solve and catch many of the issues or concerns you raise. Allowing you an output like this one: Check here -> 

image

> ps: The outcome shows detection probability as a continuous function of effect magnitude — the same thing as a power curve — but decomposed into three regions (Correct Detection, Misclassification, Non-Detection) rather than a single binary threshold, and it's derived from the placebo-calibrated null model, not IID noise injection. Not available now, but we only need a function to make the plot.

On the other hand, this power analysis give you information only about if the information holds. You don't need to re-run several times to get this, you can run the model over the most recent pre-period, get a posterior for different trajectories, estimate a CI, and then you can estimate properly what effect size would be greater than Z either in average or cumulative. You could simulate as many trajectories as you want and be creative here, but this only give you "power" based on given model uncertainty, and loop adds no information.

Additionally, adding effect has the flaws you already detected, increasing complexity without solve previous point.

-----

You are right to mention about: design prior approaches. This is well documented, and it's the complement of the null model already coming from placebo estimation, you can say, "Based on prior knowledge, my expected cumulative effect (design prior) is N" then draw a full estimation based on it, and see where it lands, helping you to estimate the curve showed above. The "curve" comes out by construction — it's the joint integration of detection probability against the design prior, weighted by plausibility. Bayesian Assurance (O'Hagan, 2005) gives you the operating characteristics by integration — the "curve" emerges naturally without a brute-force simulation loop. CausalPy already implements this via PlaceboInTime's expected_effect_prior argument (#826 ).

You can loop based on the different outcomes of this method to estimate best combination of donors or other characteristics. [Read here](https://github.com/cetagostini/placebo-in-time-quasi-experiment/blob/main/pdf/paper.pdf) and check the references!

My take in short: It's a great job but reimplement's existing logic, after #826 I can make a PR with new plots only and a notebook to show this full pipeline and how this things are solved. Unless, I'm missing something here which is very probable.

Issue #47, opened 2022-11-02
Add Synthetic Difference-in-Differences


drbenvincent:
Suggestion by @juanitorduz 

# Resources

* https://github.com/MasaAsami/pysynthdid
* https://matheusfacure.github.io/python-causality-handbook/25-Synthetic-Diff-in-Diff.html

aeturrell:
Came here to say synthetic diff-in-diff would be useful and glad to see it is on the road map. Susan Athey has published R and Stata packages that might be useful for reference:

https://synth-inference.github.io/synthdid/

https://arxiv.org/pdf/2301.11859.pdf

Thanks for the amazing work!

PR #822, opened 2026-04-03
Add SoftmaxWeightedSumFitter for synthetic control

thomaspinder:
Introduces a `_softmax_simplex_weights` helper and `SoftmaxWeightedSumFitter`, providing an alternative to `WeightedSumFitter` for synthetic control that places Normal priors on unconstrained logits and maps to the simplex via softmax (first logit pinned to zero). This parameterisation enables continuous regularisation control via the prior scale, bridging between DiD-uniform and SC-sparse weight regimes.

This PR is the first of two that work towards implementing SDiD 

Relevant Issue: #47

### DIFF
--- a/causalpy/pymc_models.py
+++ b/causalpy/pymc_models.py
@@ -678,6 +678,187 @@ class WeightedSumFitter(PyMCModel):
             self.priors["y_hat"].create_likelihood_variable("y_hat", mu=mu, observed=y)
 
 
+def _softmax_simplex_weights(
+    name: str,
+    prior: "Prior",
+    n_rows: int,
+    dims: list[str],
+) -> "pt.TensorVariable":
+    """Create simplex weights via softmax-over-Normal-logits with pinned reference.
+
+    Pins the first logit to zero (removing softmax shift invariance), samples
+    ``N - 1`` unconstrained Normal logits, and applies softmax to produce
+    weights on the simplex.
+
+    Parameters
+    ----------
+    name : str
+        Name for the PyMC Deterministic variable (e.g., "beta", "omega").
+    prior : Prior
+        Prior for the raw logits. Must be a Normal prior with appropriate dims.
+    n_rows : int
+        Number of rows in the weight matrix (e.g., n_treated for SC, 1 for SDiD).
+    dims : list[str]
+        Dimension names for the output Deterministic.
+
+    Returns
+    -------
+    pt.TensorVariable
+        Simplex weights as a PyMC Deterministic.
+    """
+    raw = prior.create_variable(f"{name}_raw")
+    if n_rows == 1 and raw.ndim == 1:
+        # 1D case: raw is shape (N-1,), produce 1D simplex of shape (N,)
+        zero_logit = pt.zeros((1,))
+        tilde = pt.concatenate([zero_logit, raw], axis=0)
+        return pm.Deterministic(name, pt.special.softmax(tilde, axis=-1), dims=dims)
+    zero_logit = pt.zeros((n_rows, 1))
+    tilde = pt.concatenate([zero_logit, raw], axis=-1)
+    return pm.Deterministic(name, pt.special.softmax(tilde, axis=-1), dims=dims)
+
+
+class SoftmaxWeightedSumFitter(PyMCModel):
+    r"""
+    Weighted sum model with softmax-over-Normal-logits parameterization.
+
+    An alternative to :class:`WeightedSumFitter` for synthetic control experiments.
+    Instead of a Dirichlet prior on the simplex weights, this model places Normal
+    priors on unconstrained logits and maps them to the simplex via the softmax
+    transform. The first logit is pinned to zero to remove the softmax's shift
+    invariance.
+
+    Defines the PyMC model:
+
+    .. math::
+        \tilde{\beta}_1 &= 0 \\
+        \tilde{\beta}_{j} &\sim \mathrm{Normal}(0, \sigma) \quad j = 2, \ldots, N \\
+        \beta &= \mathrm{softmax}(\tilde{\beta}) \\
+        \mu &= X \cdot \beta \\
+        y &\sim \mathrm{Normal}(\mu, \sigma_y) \\
+
+    Notes
+    -----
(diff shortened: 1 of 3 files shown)


drbenvincent (changes requested):
## Overall assessment

Great first contribution! This is a clean, well-tested, self-contained addition that provides an alternative weight parameterization for synthetic control. The softmax-over-Normal-logits approach is well-motivated by the SDiD literature (Arkhangelsky et al., 2021) and serves as a solid foundation for the Bayesian SDiD work in PR #823. The docstrings are excellent — the Notes section clearly contrasts the two parameterizations and their regularization behaviour.

A few minor suggestions below. Most are non-blocking, but comment 3 (`priors_from_data` ignoring the data) deserves a response — I'd like to understand the reasoning before approving.

---

## Code comments

### 1. `_softmax_simplex_weights`: consider validating the prior distribution

The docstring says the prior "Must be a Normal prior", but there's no runtime check. If someone passes a Dirichlet or Cauchy prior here the model would build without complaint but produce unexpected behaviour. A lightweight guard would help, especially since this helper will also be used in PR #823.

```python
# suggestion (optional — depends on how public this helper is intended to be)
if prior.distribution != "Normal":
    raise ValueError(
        f"_softmax_simplex_weights expects a Normal prior, got {prior.distribution}"
    )
```

Not blocking — this is an internal helper prefixed with `_`, so the audience is narrow. But worth considering since #823 will call it from a second model class.

### 2. `_softmax_simplex_weights`: the 1D branch

The 1D/2D branching logic works correctly, but the condition `n_rows == 1 and raw.ndim == 1` couples two independent checks. In practice this branch is only hit when `priors_from_data` happens to omit the `treated_units` dim, producing a 1D raw tensor. A brief inline comment explaining *when* each branch triggers would help future readers. The current comment says "1D case" but doesn't explain *why* it would be 1D.

### 3. `priors_from_data` ignores the data

Unlike `WeightedSumFitter.priors_from_data`, which adapts the Dirichlet shape to `X.shape[1]`, this implementation always returns `sigma=1.0` regardless of the data. This is fine as a default, but it means users who want tighter regularization (more DiD-like) or looser regularization (more SC-like) need to know to pass custom priors.

A docstring note explaining *how* to override sigma would be helpful for users:

```python
# Example of overriding the logit prior scale for tighter regularization:
model = SoftmaxWeightedSumFitter(
    priors={"beta_raw": Prior("Normal", mu=0, sigma=0.1, dims=["treated_units", "coeffs_raw"])}
)
```

### 4. `build_model`: `coeffs_raw` fallback

```python
coeffs_raw = (
    coords["coeffs"][1:]
    if coords and "coeffs" in coords
    else list(range(1, X.shape[1]))
)
```

The fallback `list(range(1, X.shape[1]))` produces integer coordinate labels, while the primary path produces string labels (sliced from the user's coords). This asymmetry is harmless in practice since coords are always passed through the experiment classes, but it's a bit surprising. Could this fallback raise instead? If `coords` is missing `"coeffs"`, that's likely a usage error rather than an expected path.

### 5. Test: `test_softmax_weighted_sum_fitter_priors_from_data` could also verify dims adapt to data shape

The test verifies that `priors_from_data` returns the right distribution and fixed parameters, but since the current implementation doesn't actually use `X` or `y`, it doesn't test that the prior shape would be wrong for a different number of control units. This is fine for now but worth noting as a gap — if `priors_from_data` later adapts sigma to data scale, the test should be updated to verify that.

---

## Questions for the author

1. **Pinning the first logit vs the last:** The choice to pin the *first* logit to zero means the first control unit in `coords["coeffs"]` becomes the reference. This is arbitrary and mathematically equivalent to pinning any other position, but is there a reason you chose the first? Just want to confirm this is intentional and not a consequence of the concatenation order. (It matches the blog post's parameterization, so I suspect it's deliberate.)

2. **Connection to PR #823:** The `_softmax_simplex_weights` helper is designed to be reused in the SDiD weight fitter. Have you verified that the 1D branch (which is never exercised in PR #822's tests) is actually needed for #823, or could the helper be simplified to always use the 2D path?

---

## Summary

| Category | Verdict |
|---|---|
| Correctness | No issues found |
| Tests | Comprehensive — unit, integration, and prior tests all present |
| CI | All 17 checks pass |
| Documentation | Excellent docstrings with math, contrast to Dirichlet, and runnable examples |
| Code style | Consistent with existing `WeightedSumFitter` patterns |
| Public API | Consistent — accessible via `cp.pymc_models.SoftmaxWeightedSumFitter`, same pattern as `WeightedSumFitter` |

**Recommendation:** Approve with minor suggestions, pending a response on comment 3 (`priors_from_data` not adapting to the data).

thomaspinder:
On Code Comment:
1. Good callout - updated.
2. Docstring updated
3. This is deliberate, and the mechanism differs from WeightedSumFitter. The Dirichlet dist. used in `WeightedSumFitter` requires an explicit length-N concentration vector `a=np.ones(n_predictors)`, so `priors_from_data` must read `X.shape[1]` to set the shape. In this class, the Normal dist.'s logits prior broadcasts via `dims=["treated_units", "coeffs_raw"]`, so the shape adapts without inspecting the data. The `sigma=1.0` is the analogue of `a=1`, imposing moderate regularisation. Documenting how to override sigma is a good suggestion. I'll add a docstring example showing tighter (sigma=0.1, DiD-like) and looser (sigma=10, SC-like) configurations.
4. Agreed — if coords is missing "coeffs" that's a usage error. Will replace the fallback with a ValueError.
5. Acknowledged. The test will become more meaningful if priors_from_data later adapts to data scale.

On Questions:
1. Pinning the first logit is deliberate as it matches the standard softmax identifiability convention. Pinning any position is mathematically equivalent - I just chose the first.
2. The SDiD weight fitter calls `_softmax_simplex_weights(n_rows=1, ...)` for both omega (unit weights) and lam (time weights), both producing 1D outputs. Without the 1D branch, the concatenation would produce an incorrect (1, N) shape for what should be a 1D simplex.

drbenvincent (approved):
Approved. Thanks for this! 

Hoping to get to review the other, bigger, PR soon.

PR #831, opened 2026-04-08
feat: add legend_kwargs parameter to BaseExperiment.plot()

### LINKED ISSUE #341: Fix legend overlapping with lines on causal impact plots
In a number of situations, the legend is overlapping with data in the causal impact plots. Here's an example:
Screenshot 2024-05-08 at 21 06 46

- [ ] Create a kwarg in the `plot` method which allows users to specify that they want the legend outside the plot. It might take some experimentation to get this right. For example, does it look best to put the legend on the side of the plots (but this may look strange) or perhaps under the bottom?

haoyu-haoyu:
## Summary

Add `legend_kwargs` parameter to `BaseExperiment.plot()` to allow users to control legend placement, fixing the overlap issue in causal impact plots.

## Usage

```python
result.plot(legend_kwargs={"loc": "upper left", "bbox_to_anchor": (1.04, 1)})
```

## Implementation

- New `legend_kwargs: dict | None` parameter in `BaseExperiment.plot()`
- Applied to all axes that have a legend after the plot is created
- Handles single Axes, lists, and numpy arrays of axes
- Skips axes without legends (no-op safety)

## Design decision

Per maintainer feedback, this uses arbitrary kwargs passed to `ax.legend()` rather than a string-based position enum, giving users full control while keeping good defaults.

Closes #341

### DIFF
--- a/causalpy/experiments/base.py
+++ b/causalpy/experiments/base.py
@@ -157,7 +157,13 @@ class BaseExperiment(ABC):
         """Optional maketables plugin hook for default statistic rows."""
         return get_maketables_adapter(self.model).default_stat_keys(self)
 
-    def plot(self, *args: Any, show: bool = True, **kwargs: Any) -> tuple:
+    def plot(
+        self,
+        *args: Any,
+        show: bool = True,
+        legend_kwargs: dict[str, Any] | None = None,
+        **kwargs: Any,
+    ) -> tuple:
         """Plot the model.
 
         Internally, this function dispatches to either `_bayesian_plot` or `_ols_plot`
@@ -168,6 +174,10 @@ class BaseExperiment(ABC):
         show : bool, optional
             Whether to automatically display the plot. Defaults to True.
             Set to False if you want to modify the figure before displaying it.
+        legend_kwargs : dict, optional
+            Keyword arguments passed to ``ax.legend()`` to control legend
+            placement. For example, ``legend_kwargs={"loc": "upper left",
+            "bbox_to_anchor": (1.04, 1)}`` moves the legend outside the axes.
         """
         # Apply arviz-darkgrid style only during plotting, then revert
         with plt.style.context(az.style.library["arviz-darkgrid"]):
@@ -177,6 +187,18 @@ class BaseExperiment(ABC):
                 fig, ax = self._ols_plot(*args, **kwargs)
             else:
                 raise ValueError("Unsupported model type")
+
+        # Apply legend customization if requested
+        if legend_kwargs is not None:
+            axes = ax if isinstance(ax, list) else [ax]
+            try:
+                # Handle numpy arrays of axes (e.g., from plt.subplots(nrows=2))
+                axes = list(ax.flat) if hasattr(ax, "flat") else axes
+            except (TypeError, AttributeError):
+                pass
+            for a in axes:
+                if a.get_legend() is not None:
+                    a.legend(**legend_kwargs)
 
         if show:
             plt.show()
--- a/causalpy/experiments/base.py
+++ b/causalpy/experiments/base.py
@@ -33,6 +33,57 @@ from causalpy.reporting import EffectSummary
 from causalpy.skl_models import create_causalpy_compatible_class
 
 
+def _apply_legend_kwargs(legend: Any, kwargs: dict[str, Any]) -> None:
+    """Mutate an existing Legend in place without recreating it.
+
+    This preserves custom handles (e.g. ``(Line2D, PolyCollection)`` tuples
+    built by :func:`~causalpy.plot_utils.plot_xY`) that would be lost if the
+    legend were rebuilt with ``ax.legend()``.
+
+    Supported keys: ``loc``, ``bbox_to_anchor``, ``bbox_transform`` (only
+    with ``bbox_to_anchor``), ``fontsize``, ``frameon``, ``title``.
+
+    Raises
+    ------
+    TypeError
+        If *kwargs* contains keys that cannot be applied in place.
+    """
+    _SUPPORTED = {"loc", "bbox_to_anchor", "bbox_transform", "fontsize", "frameon", "title"}
+    unsupported = set(kwargs) - _SUPPORTED
+    if unsupported:
+        raise TypeError(
+            f"legend_kwargs keys not supported for in-place mutation: "
+            f"{sorted(unsupported)}. Supported keys: {sorted(_SUPPORTED)}"
+        )
+    if "bbox_transform" in kwargs and "bbox_to_anchor" not in kwargs:
+        raise TypeError(
+            "bbox_transform requires bbox_to_anchor to be specified as well"
+        )
+
+    if "loc" in kwargs:
+        loc = kwargs["loc"]
+        # set_loc is public in matplotlib >= 3.8; fall back to the stable
+        # private helper for older versions, converting string names to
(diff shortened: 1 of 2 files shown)


drbenvincent (changes requested):
Thanks for this contribution @haoyu-haoyu. The API direction is good and I do think we should merge this functionality, but the current implementation is not quite merge-ready yet. The main thing to fix is making sure legend customization preserves the existing legend rather than changing what the plot shows.

1. **Blocker — Legend recreation changes existing plot semantics**

   Calling `a.legend(**legend_kwargs)` after the plot is built creates a new legend rather than modifying the existing one. That is risky in this codebase because several experiment plots already construct custom legends with explicit `handles` / `labels`, often using `(line, patch)` tuples from `plot_xY()` to show a mean line plus HDI band as one legend entry. A second bare `legend()` call can silently drop those curated entries and reset styling like `fontsize`.

   **Suggested direction:** do not call `legend()` "blindly" after the fact. Either mutate the existing `Legend` object in place when possible, or if recreation is necessary, rebuild from the currently displayed legend handles/labels and preserve its existing styling. If fully generic `legend_kwargs` proves awkward to support safely, it would be better to narrow the feature to placement-oriented kwargs than to risk changing legend contents.

2. **Major — The PR does not reliably "fix overlap" yet**

   Some plots call layout helpers before control returns to `BaseExperiment.plot()`; for example, `piecewise_its` currently does a `tight_layout()` inside the plot method. If the legend is moved afterward with something like `bbox_to_anchor`, it can still be clipped because layout is not recomputed.

   **Suggested direction:** after applying legend placement, rerun layout on the figure, or explicitly document that outside-the-axes legend placement may require `show=False` plus a follow-up layout adjustment by the caller. If that is out of scope for this PR, I would soften the PR description from "fixes the overlap issue" to "adds legend placement control."

3. **Major — Missing regression test for the new contract**

   This should come with a focused test. A happy-path manual check is not enough because the main risk is regressions in existing custom legends.

   **Suggested direction:** add a `show=False` test for one representative experiment that already uses a custom legend (for example DiD, ITS, or synthetic control). Assert that `legend_kwargs` changes placement while preserving legend presence and legend content (for example text labels or entry count). That would document the intended behavior and protect the implementation going forward.

4. **Nice to have — Add a small user-facing example**

   This feature is useful enough that people will look for it in examples before they inspect the API signature.

   **Suggested direction:** add one short snippet in a docstring or notebook showing `result.plot(show=False, legend_kwargs={...})`, ideally paired with whatever layout call is needed when placing the legend outside the axes. Not a blocker for correctness, but good polish for closing `#341`.

haoyu-haoyu (commented):
Thanks for the thorough review @drbenvincent! I've addressed all four points in the latest commit (503d8cd):

**1. Legend recreation → in-place mutation (Blocker)**

You were right — calling `ax.legend(**legend_kwargs)` was silently destroying the custom `(Line2D, PolyCollection)` tuple handles built by `plot_xY()`. I replaced this with a new `_apply_legend_kwargs()` helper that mutates the existing `Legend` object in place using public APIs (`set_loc()`, `set_bbox_to_anchor()`, `set_frame_on()`, `set_title()`, and per-text `set_fontsize()`). The legend is never recreated, so all custom handles and styling survive exactly as the subclass created them.

For matplotlib < 3.8 (where `set_loc()` isn't public), the fallback converts string location names to numeric codes via `Legend.codes` before calling `_set_loc()`.

**2. Layout recomputation (Major)**

`fig.tight_layout()` is now called **only** when `bbox_to_anchor` is in `legend_kwargs` (the case where clipping actually happens), avoiding a redundant second `tight_layout()` call on experiments like `PiecewiseITS` that already run it internally.

**3. Regression tests (Major)**

Added 10 new tests covering:
- Label + handle count preservation (PyMC and OLS)
- Location change preserves labels
- `bbox_to_anchor` triggers `fig.tight_layout()` (mocked assertion)
- `frameon` and `title` in-place mutation
- Fontsize preservation when not overridden + explicit override
- `None` no-op
- Multi-axis ITS (exercises list-of-axes flattening)
- `TypeError` on unsupported keys (fail-fast)
- `TypeError` on `bbox_transform` without `bbox_to_anchor`

**4. Docstring example (Nice-to-have)**

Added NumPy-style docstring with supported keys list and a usage example showing `show=False` + `legend_kwargs`.

The supported keys are narrowed to placement-oriented properties (`loc`, `bbox_to_anchor`, `bbox_transform`, `fontsize`, `frameon`, `title`) as you suggested, rather than accepting arbitrary kwargs that could risk changing legend contents. Unsupported keys raise `TypeError` with a clear message.

All 15 tests pass and `ruff check` is clean.

drbenvincent (changes requested):
Thanks for the thorough follow-up, @haoyu-haoyu — I can see you've put real thought into addressing each point. The in-place mutation approach via `_apply_legend_kwargs()` is the right design, and narrowing the API surface to placement-oriented keys was a good call. The tests are well-structured and cover the important cases.

A few things to address before we can merge:

## Must fix

**1. CI is failing — please run `prek run --all-files`**

The `prek` check fails because `ruff format` wants to reformat both changed files. The fixes are trivial (expanding a set literal across multiple lines, collapsing/expanding a couple of expressions) — but they need to be applied before CI goes green.

Please run:

```bash
prek run --all-files
```

before pushing. That's the canonical pre-commit check for this repo and it runs the full suite: ruff lint, **ruff format**, mypy, interrogate, codespell, etc. Running `ruff check` alone only covers the linter — it doesn't run the formatter, which is what's failing here.

A note on `ruff check` scope: if you run `ruff check .` against the whole repo you'll see ~28 pre-existing errors in `docs/` notebooks — those aren't yours. Our pre-commit config (`exclude: ^docs/`) deliberately skips docs for ruff lint, but that exclusion only lives in `.pre-commit-config.yaml`, not in ruff's own config in `pyproject.toml`. So bare `ruff check .` doesn't know about it. `prek` handles the scoping correctly.

## Suggestion (non-blocking)

**2. The `contextlib.suppress(TypeError, AttributeError)` on the axes-flattening logic is unnecessary**

```python
axes = ax if isinstance(ax, list) else [ax]
with contextlib.suppress(TypeError, AttributeError):
    axes = list(ax.flat) if hasattr(ax, "flat") else axes
```

The `hasattr(ax, "flat")` guard already prevents the `AttributeError`, and `list()` on a numpy flat iterator won't raise `TypeError`. The suppress could mask a real bug from a future subclass. A clearer alternative:

```python
if hasattr(ax, "flat"):
    axes = list(ax.flat)
elif isinstance(ax, list):
    axes = ax
else:
    axes = [ax]
```

Not blocking on this, but it would make the intent clearer and avoid silently swallowing unexpected errors.

---

Overall this is looking close — the core design is solid and the tests are comprehensive. Fix the formatting and this should be ready to go.

haoyu-haoyu:
Thanks @drbenvincent! Both items addressed in 1228b2e (pushed yesterday after your review):

1. **Ruff format** — ran `prek run --all-files`. CI is now green across the board (`prek`, `pre-commit.ci`, both Python versions, notebooks).
2. **Drop `contextlib.suppress` on axes flatten** — adopted your suggested explicit `if hasattr(ax, "flat") / elif isinstance(ax, list) / else` form. Intent is clearer and we no longer silently swallow unexpected errors.

Ready for another look when you have a moment.

drbenvincent (approved):
Great work @haoyu-haoyu !

PR #842, opened 2026-04-14
fix(ipw): preserve treatment design_info instead of overwriting with covariates

### LINKED ISSUE #835: InversePropensityWeighting: _build_design_matrices overwrites _t_design_info
### Description

In `InversePropensityWeighting._build_design_matrices`, `self._t_design_info` is assigned twice in a row: first to the treatment matrix `t` design info from patsy, then immediately overwritten with the covariate matrix `X` design info.

```python
t, X = dmatrices(self.formula, self.data)
self._t_design_info = t.design_info
self._t_design_info = X.design_info
self.labels = X.design_info.column_names
```

So the attribute name suggests treatment design metadata, but only `X.design_info` is retained. Elsewhere in the codebase (e.g. `instrumental_variable.py`), `_t_design_info` is set once from `t.design_info`, which matches the name.

### Expected behavior

Treatment and covariate design info should both be available if stored—typically as two attributes (e.g. `_t_design_info` and something like `_x_design_info` / reuse `X.design_info` explicitly) or by renaming the single stored object to match what is actually kept.

### Actual behavior

Treatment `design_info` is discarded; any code or future use that expects `_t_design_info` to refer to the treatment side of the formula will be wrong.

### Additional context

FYI @NathanielF (original author of this experiment)—no expectation you fix this; logging for visibility and whoever picks it up.

haoyu-haoyu:
Fixes #835.

### Description

\`InversePropensityWeighting._build_design_matrices\` assigned \`self._t_design_info\` twice in a row — first to the treatment matrix's \`design_info\`, then immediately overwriting it with the covariate matrix's \`design_info\`. The attribute name suggests treatment metadata but only the covariate side was being kept.

\`\`\`python
t, X = dmatrices(self.formula, self.data)
self._t_design_info = t.design_info
self._t_design_info = X.design_info     # ← overwrites the line above
\`\`\`

### Fix

Store the covariate side under \`self._x_design_info\`, matching the existing convention used elsewhere in this codebase:

- \`causalpy/experiments/instrumental_variable.py:153,159\`
- \`causalpy/experiments/interrupted_time_series.py:162\`
- \`causalpy/experiments/regression_discontinuity.py:155\`
- \`causalpy/experiments/prepostnegd.py:120\`

The treatment design now lives under \`self._t_design_info\` as the name implies.

### Tests

Added a \`TestDesignInfoAttributes\` suite to \`causalpy/tests/test_ipw_get_ate.py\` that:

- Verifies \`_t_design_info.column_names == ["trt"]\` (the LHS of the test formula \`trt ~ 1 + age + race\`).
- Verifies \`_x_design_info.column_names\` contains the covariates.
- Verifies the two attributes are distinct objects.

The class would have failed on master because \`_t_design_info\` would have held the covariate columns and \`_x_design_info\` would not have existed at all.


drbenvincent (approved):
Approving, but would appreciate a sanity check from @NathanielF before we merge

PR #847, opened 2026-04-15
Add validation to _data_setter for non-standard data node names

### LINKED ISSUE #663: PyMCModel._data_setter hard-codes X/y data nodes
## Description

`PyMCModel._data_setter()` assumes the underlying PyMC model uses data nodes named `"X"` and `"y"`, and it always injects a zero-filled `y` with shape `(n_obs, n_treated_units)` during prediction. This is a tight coupling to the current internal model templates and can silently break if a subclass uses different data node names or additional observed variables. The method documentation acknowledges this assumption but the interface does not allow customization, so custom models can fail at predict time even if they otherwise follow the `PyMCModel` API.

## Steps to Reproduce

1. Create a `PyMCModel` subclass that uses different data node names (e.g., `"X_design"` and `"y_obs"`).
2. Fit the model and call `predict()`.

```python
class CustomModel(PyMCModel):
    def build_model(self, X, y, coords):
        with self:
            self.add_coords(coords)
            X_design = pm.Data("X_design", X)
            y_obs = pm.Data("y_obs", y)
            ...

# predict() calls _data_setter(), which updates only "X" and "y"
```

## Expected Behavior

`predict()` should update the data nodes that the subclass actually defines, or the base class should allow subclasses to override the data node names in a supported way. At minimum, the base behavior should be safe for subclasses that rename nodes.

## Actual Behavior

`_data_setter()` calls `pm.set_data({"X": X, "y": zeros(...)})`, which fails or silently leaves the real data nodes untouched if they are named differently, causing incorrect predictions or runtime errors.

## Environment

- Python version: 3.11+
- OS: macOS (darwin)

More thought and decision making is needed here to balance flexibility, safety, and API stability. Possible directions:

- Explicit configuration: add class attributes or init args like `data_node_names = {"X": "X", "y": "y"}` so subclasses can override without re-implementing `_data_setter`.
- Override contract: make `_data_setter` abstract or clearly documented for subclasses to override, with a helper utility for common cases (e.g., `set_data_with_obs_ind`).
- Graph inspection: discover `pm.Data` containers programmatically from the model graph and update any matching dims; safer but more complex and may be brittle across PyMC versions.
- Model-level registration: provide a `register_data_nodes()` helper during `build_model()` to record which data containers should be updated for prediction.
- Validate on fit: store the names of data nodes used in `build_model()` and assert they exist during `predict()`, raising a clear error if `"X"`/`"y"` are missing instead of silently misbehaving.

## Additional Context

Current implementation:

```python
pm.set_data(
    {"X": X, "y": np.zeros((new_no_of_observations, n_treated_units))},
    coords={"obs_ind": obs_coords},
)
```

drbenvincent:
## Summary

`PyMCModel._data_setter()` hard-codes `"X"` and `"y"` as data node names. If a subclass uses different names, it silently misbehaves or throws a cryptic PyMC error. This PR adds validation that raises a clear `ValueError` with actionable guidance when expected data nodes are missing.

Fixes #663

## Changes

- **`causalpy/pymc_models.py`**:
  - Added validation at the top of `_data_setter()` that checks `"X"` and `"y"` exist in `self.named_vars` before proceeding
  - Raises `ValueError` with guidance to override `_data_setter()` when nodes are missing
  - Updated class and method docstrings to document the override contract
  - No behavioral changes to existing logic — the `pm.set_data` call is unchanged

- **`causalpy/experiments/panel_regression.py`**:
  - Fixed pre-existing mypy `type: ignore` codes (`attr-defined` → `union-attr`)

- **`causalpy/tests/test_pymc_models.py`**:
  - Added `NonStandardDataModel` test subclass using renamed data nodes
  - Added `TestDataSetterValidation` with 2 tests:
    - `test_mismatched_X_raises`: verifies clear error when `"X"` is renamed
    - `test_missing_y_raises`: verifies clear error when only `"y"` is renamed

## Design rationale

The issue proposed 5 possible directions. After examining all existing subclasses, I found that:

- All standard models (`LinearRegression`, `WeightedSumFitter`, `SoftmaxWeightedSumFitter`) use `"X"` and `"y"` — they work as-is
- Complex models (`BayesianBasisExpansionTimeSeries`) already override `_data_setter` entirely
- Models with fundamentally different data (`StateSpaceTimeSeries`, `InstrumentalVariableRegression`, `PropensityScore`) override `predict` or `fit` entirely

So the override contract already works for all existing subclasses. The missing piece was **validation**: a subclass that forgets to override `_data_setter` previously got a cryptic error. Now it gets a clear `ValueError` pointing to the solution.

I deliberately kept this minimal — no new class attributes or API surface — because the name-remapping-only case doesn't arise in any existing subclass and would add complexity for a speculative benefit.

## Testing

- [x] 2 new validation tests pass
- [x] All 49 existing tests in `test_pymc_models.py` pass
- [x] All 30 PyMC integration tests pass
- [x] All 21 timeseries model coverage tests pass

## Checklist

- [x] Backwards compatible (no behavioral changes)
- [x] No new API surface
- [x] Tests cover both X-missing and y-missing error paths
- [x] Docstrings updated

### DIFF
--- a/causalpy/experiments/panel_regression.py
+++ b/causalpy/experiments/panel_regression.py
@@ -557,7 +557,7 @@ class PanelRegression(BaseExperiment):
         """
         # Get posterior predictions
         if isinstance(self.model, PyMCModel):
-            mu = self.model.idata.posterior["mu"]  # type: ignore[attr-defined]
+            mu = self.model.idata.posterior["mu"]  # type: ignore[union-attr]
             pred_mean = mu.mean(dim=["chain", "draw"]).values.flatten()
             pred_lower = mu.quantile(0.025, dim=["chain", "draw"]).values.flatten()
             pred_upper = mu.quantile(0.975, dim=["chain", "draw"]).values.flatten()
@@ -673,7 +673,7 @@ class PanelRegression(BaseExperiment):
 
         if isinstance(self.model, PyMCModel):
             # Bayesian: get posterior means
-            beta = self.model.idata.posterior["beta"]  # type: ignore[attr-defined]
+            beta = self.model.idata.posterior["beta"]  # type: ignore[union-attr]
             unit_fe_indices = [self.labels.index(name) for name in unit_fe_names]
 
             # Get mean and std for each unit FE
--- a/causalpy/experiments/panel_regression.py
+++ b/causalpy/experiments/panel_regression.py
@@ -11,28 +11,30 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
-"""Panel Regression with Fixed Effects."""
+"""
+Panel Regression with Fixed Effects
+"""
 
-import re
 from typing import Any, Literal
 
 import arviz as az
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+import xarray as xr
 from matplotlib.gridspec import GridSpec
-from patsy import ModelDesc, dmatrices
+from patsy import dmatrices
 from scipy import stats
 from sklearn.base import RegressorMixin
 
-from causalpy.constants import HDI_PROB
 from causalpy.custom_exceptions import DataException
-from causalpy.experiments.model_adapter import build_coords
 from causalpy.pymc_models import PyMCModel
 from causalpy.reporting import EffectSummary
 from causalpy.utils import round_num
 
 from .base import BaseExperiment
+
+LEGEND_FONT_SIZE = 12
 
 
 class PanelRegression(BaseExperiment):
@@ -59,7 +61,6 @@ class PanelRegression(BaseExperiment):
         If provided, time fixed effects will be included. Default is None.
     fe_method : {"dummies", "demeaned"}, default="dummies"
         Method for handling fixed effects:
-
         - "dummies": Use unpooled dummy-variable fixed effects
           (``C(unit)``/``C(time)`` in formula). Gets individual unit effect
           estimates but creates N-1 dummy columns. Best for small N.
@@ -67,8 +68,6 @@ class PanelRegression(BaseExperiment):
           but doesn't directly estimate individual unit effects.
     model : PyMCModel or RegressorMixin, optional
         A PyMC (Bayesian) or sklearn (OLS) model. If None, a model must be provided.
-    **kwargs
-        Additional keyword arguments forwarded to :class:`BaseExperiment`.
 
     Attributes
     ----------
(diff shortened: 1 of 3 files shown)


drbenvincent:
## Automated triage

**Why:**
- Contained internal fix: adds validation to `_data_setter()` in `pymc_models.py` (20 lines added) with 84 new test lines.
- No API change, no behavioural change to existing code — only raises a clearer error when data node names are non-standard.
- `codecov/patch` failed (likely a threshold issue for the small diff). PR is conflicted (DIRTY).

**Review focus:**
1. Verify the override contract is documented (the PR says docstrings were updated).
2. Check the `panel_regression.py` mypy fix (`attr-defined` → `union-attr`).
3. Assess whether the `codecov/patch` failure is a false positive or indicates uncovered edge cases.

**Confidence:** high

drbenvincent:
## Automated review — CausalPy PR review automation (head: 08ee2c3b3ec925138cc34ed24e431c0d0153d193)

### Change contract
Adds validation to `PyMCModel._data_setter()` that raises a clear `ValueError` when expected mutable data nodes `"X"` or `"y"` are missing from the model graph. Existing subclasses that use standard names are unaffected. Subclasses that renamed nodes and forgot to override `_data_setter()` previously hit cryptic PyMC errors; they now get an actionable message directing them to override the method. No API surface change. Pre-existing mypy `type: ignore` codes in `panel_regression.py` corrected from `attr-defined` to `union-attr`. Fixes #663.

### Evidence and preconditions
- **Changed files**: `causalpy/pymc_models.py` (+validation, +docstrings), `causalpy/tests/test_pymc_models.py` (+2 tests, +1 test helper class), `causalpy/experiments/panel_regression.py` (mypy fix only)
- **CI**: All required checks pass (prek, tests on 3.11 and 3.14, notebooks x3, zizmor, build, environment-yml).
- **codecov/patch**: failing — the patch adds lines to `pymc_models.py` (the `ValueError` raise) that are not covered by the test suite. The two new tests exercise the error paths but do not cover the validation's `for name in ("X", "y"):` loop body in a non-error scenario.
- **Merge conflict**: `mergeable_state: dirty` — the branch has conflicts with `main` and must be updated before merging.

### Documentation preview
RTD build succeeded: https://causalpy--847.org.readthedocs.build/en/847/

No documentation source file changed; API docs for `PyMCModel` and `_data_setter()` will render the updated docstrings, which describe the override contract. The human should verify that:
- The class-level note about mutable data node naming renders clearly without breaking the surrounding docstring layout.
- The `_data_setter()` method docstring renders the subclass-guidance paragraph correctly.

### Review focus
- Verify the `ValueError` message is sufficiently actionable for a downstream subclass author who encounters it — does `"override _data_setter()"` give enough context to actually implement the override?
- Confirm that `self.named_vars` is guaranteed to exist on all `PyMCModel` instances by the time `_data_setter()` is called (it is set during `build_model` → `pm.Model.__init__`, but consider if any code path calls `_data_setter` before `fit`).
- Check that the merge conflict resolution does not accidentally drop the validation logic or the docstring additions.

### Findings
- **Merge conflict with main** — the branch cannot merge as-is. This is a hard blocker; the conflict must be resolved before a human can evaluate the final diff.
- **codecov/patch failure** — the validation loop body (`for name in ("X", "y"):`) has no coverage when both nodes are present. The two tests only exercise the error branches. A trivially simple test that calls `predict` on a standard model (which exercises the happy path through the loop) would resolve this. This is not a correctness concern but gap-filling CI noise.
- The validation uses a loop over two names but raises on the first missing one. If both nodes are missing, the user sees only the `"X"` error, fixes it, reruns, and hits the `"y"` error. This is a deliberate and reasonable design choice, not a defect.

### Recommendation
human review required — merge conflict with `main` and the failing codecov/patch check must be addressed before approval.

Issue #848, opened 2026-04-15
Update notebooks to use new xr.Dataset API


drbenvincent:
> **Depends on:** #849 must be merged first.

## Update notebooks to use new `xr.Dataset` API

Following the consolidation of experiment design-matrix attributes into `xr.Dataset` objects (#199), the old attribute names (`pre_X`, `pre_y`, `post_X`, `post_y`, `datapre_control`, `datapre_treated`, `datapost_control`, `datapost_treated`, `X`, `y`) are deprecated. They still work via backward-compatible `@property` accessors that emit `DeprecationWarning`, but notebooks should be migrated to the new API.

### New API

| Experiment type | Old attributes | New access pattern |
|---|---|---|
| InterruptedTimeSeries | `result.pre_X`, `result.pre_y`, `result.post_X`, `result.post_y` | `result.pre_design["X"]`, `result.pre_design["y"]`, `result.post_design["X"]`, `result.post_design["y"]` |
| SyntheticControl | `result.datapre_control`, `result.datapre_treated`, `result.datapost_control`, `result.datapost_treated` | `result.pre_design["control"]`, `result.pre_design["treated"]`, `result.post_design["control"]`, `result.post_design["treated"]` |
| Formula-based (DiD, RD, RK, PrePostNEGD, PiecewiseITS, PanelRegression) | `result.X`, `result.y` | `result.design["X"]`, `result.design["y"]` |

### Affected notebooks

These notebooks access deprecated attributes directly and should be updated:

- [ ] `docs/source/notebooks/did_pymc.ipynb` — `result.causal_impact` (not deprecated, but review for other attrs)
- [ ] `docs/source/notebooks/did_pymc_banks.ipynb` — `result.causal_impact`
- [ ] `docs/source/notebooks/rd_donut_pymc.ipynb` — `result.discontinuity_at_threshold`, `result.fit_data`
- [ ] `docs/source/notebooks/its_pymc.ipynb` — `result.post_impact`, `result.datapost`
- [ ] `docs/source/notebooks/sc_pymc.ipynb` — `result.post_impact_cumulative`
- [ ] `docs/source/notebooks/geolift1.ipynb` — `result.post_impact_cumulative`

### Notes

- The deprecated properties will be removed in a future release.
- Notebooks should be re-run after updating to verify outputs are unchanged.
- Some notebooks may access attributes that were not deprecated (e.g., `result.causal_impact`, `result.post_impact`); these should be reviewed but may not need changes.

PR #823, opened 2026-04-03
Add Bayesian Synthetic Difference-in-Differences

thomaspinder:
Implements the cut-posterior SDiD formulation of SDiD as a new experiment class. Unit and time weights are estimated jointly in a single `SyntheticDifferenceInDifferencesWeightFitter` model, and the ATT is computed analytically via the double-difference formula. Includes the California Proposition 99 dataset, a Jupyter notebook demo, and full test coverage.

Depends on `feat/softmax-weighted-sum-fitter` (PR #822) for the `_softmax_simplex_weights` helper and `SoftmaxWeightedSumFitter` base.

Relevant Issue: #47

PR #849, opened 2026-04-15
Consolidate experiment design-matrix attributes into xr.Dataset

### LINKED ISSUE #199: Simplify `ExperimentalDesign` classes by consolidating relevant properties into `idata` object

drbenvincent:
## Summary

- Bundles loose `xr.DataArray` attributes on 9 experiment classes into `xr.Dataset` objects, reducing attribute sprawl while keeping related design matrices together.
  - **Pre/post classes** (ITS, SC, SDiD): 4 DataArrays consolidated into `pre_design` / `post_design` Datasets.
  - **Formula-based classes** (DiD, RD, RK, PrePostNEGD, PiecewiseITS, PanelRegression): 2 DataArrays consolidated into a single `design` Dataset.
- All old attribute names preserved via a centralized deprecation mechanism (`BaseExperiment.__getattr__` driven by a declarative `_deprecated_design_aliases` mapping) that emits `DeprecationWarning` for backward compatibility.
- Updates `reporting.py`, `maketables_adapters.py`, `checks/convex_hull.py`, and tests to use the new Dataset access patterns internally, so CausalPy never warns against itself.
- `InversePropensityWeighting`, `InstrumentalVariable`, and `StaggeredDiD` are intentionally **not** migrated: their design matrices have non-standard layouts (IPW pairs covariates with a treatment vector `t`, IV has an instrument matrix `Z` plus an endogenous treatment, StaggeredDiD maintains a train/full split) that don't fit the shared `X`/`y` Dataset shape, so they keep their existing attributes with no deprecation shims.
- UML diagrams (`docs/source/_static/classes.png` / `packages.png`) regenerated via `make uml` to reflect the consolidated attributes.

Closes #199. Follow-up notebook migration tracked in #848. Removal of the deprecation shims is tracked in draft PR #958 (do not merge that until a release has shipped with the warnings in place).

## Test plan

- [x] All 1102 existing tests pass (0 failures), including a parametrized backward-compatibility suite (`test_deprecated_design_aliases.py`) covering every alias on every migrated class
- [x] `prek run --all-files` passes (lint, format, mypy, codespell, notebook validation)
- [x] Deprecated properties still work (notebooks and external code unaffected)


drbenvincent (commented):
Adversarial review from `gpt-5.4-xhigh`.

1. `causalpy/checks/convex_hull.py` still calls deprecated aliases. `ConvexHullCheck.run()` reads `sc.datapre_control` and `sc.datapre_treated`, which on this branch now route through the new deprecation shims. I reproduced this locally and the check emits:
   - `datapre_control is deprecated, use pre_design['control']`
   - `datapre_treated is deprecated, use pre_design['treated']`

   That means a normal synthetic-control workflow now makes CausalPy warn against itself. Immediate fix direction: do a repo-wide sweep for internal uses of the deprecated names, switch those callers to `design[...]` / `pre_design[...]` / `post_design[...]`, and add a warning-focused test in `causalpy/tests/test_cross_cutting_checks.py` asserting that `ConvexHullCheck.run()` stays warning-free.

2. The refactor still has a lot of duplicate migration boilerplate, and that is making the change less elegant than it could be. Right now the same migration pattern is hand-written across the experiment classes:
   - build one or two `xr.Dataset`s with near-identical `X`/`y` or `control`/`treated` wrapping logic,
   - preserve temporary raw arrays just long enough to build those datasets,
   - add near-identical deprecated properties that warn and then forward to the new dataset-backed location.

   You can see this pattern repeated in `diff_in_diff.py`, `panel_regression.py`, `piecewise_its.py`, `prepostnegd.py`, `regression_discontinuity.py`, `regression_kink.py`, plus the split pre/post variants in `interrupted_time_series.py` and `synthetic_control.py`. The practical problem is not just aesthetics: every repeated shim and dataset wrapper is another place to miss an internal migration, drift in warning text, or leave coverage behind. The `ConvexHullCheck` miss feels like a direct symptom of that copy-paste surface area.

   Suggested fix direction: centralize more of this in `BaseExperiment` (or a small mixin/helper module) so subclasses only describe what is experiment-specific. For example:
   - one helper for the common single-design case that builds the standard `{"X", "y"}` dataset given raw arrays, observation index, labels, and treated-unit labels;
   - one helper for split designs so ITS-style classes can reuse the same builder for pre/post periods;
   - one shared deprecation-forwarding helper so the compatibility properties become one-liners instead of each class repeating its own `warnings.warn(...); return self.design[...]` block.

   I would prefer that over adding more bespoke per-class migration code. If you want to avoid magic, you do not need a dynamic `__getattr__`; even explicit properties backed by a shared helper would already remove most of the duplication.

3. The backward-compatibility layer is still under-tested, and the remote results reflect that. Every non-Codecov check is green, but `codecov/patch` is failing at `84.71%` with `37` missing lines, concentrated in the new deprecated properties across the experiment classes. Since preserving old attribute access is one of the main claims of the PR, I’d expect targeted tests that exercise those legacy aliases and assert both the warning and the data equivalence to the new dataset-backed API.

   Suggested fix direction: add a small parametrized compatibility test suite rather than lots of bespoke tests. I would cover:
   - representative single-design aliases (`X`, `y`) and split-design aliases (`pre_X`, `pre_y`, `post_X`, `post_y`);
   - the synthetic-control aliases (`datapre_control`, `datapre_treated`, `datapost_control`, `datapost_treated`);
   - equality of the returned object/data with the new dataset-backed access path;
   - correct deprecation behavior via `pytest.deprecated_call()` or explicit warning capture.

   If you centralize the alias metadata, you can drive these tests from the same mapping and shrink both implementation duplication and test duplication at the same time.

Net: I like the direction of moving related arrays into `xr.Dataset`s, but I don’t think this is ready yet. The current version is functionally plausible, but not yet simple or especially elegant because the migration logic is still spread across too many classes and one real regression has already slipped through. I’d first eliminate internal uses of deprecated aliases, then centralize the migration helpers, then land this with focused compatibility coverage.

drbenvincent (commented):
Follow-up adversarial review from `gpt-5.4-xhigh`.

No new blocking findings from me on the latest revision.

What satisfies my earlier review:

- The duplicate migration boilerplate is now materially reduced. Moving deprecated alias forwarding into `BaseExperiment.__getattr__` with a declarative `_deprecated_design_aliases` mapping is the kind of centralization I was asking for, and `_build_design_dataset()` usefully consolidates the common `design["X"]` / `design["y"]` wrapping path.
- The self-warning regression is fixed. `causalpy/checks/convex_hull.py` now reads `pre_design[...]` directly, and `causalpy/maketables_adapters.py` also no longer relies on deprecated design aliases internally.
- The backward-compatibility contract is now tested in the right way. `causalpy/tests/test_deprecated_design_aliases.py` checks both deprecation behavior and data identity, and it adds a warning-free test for the convex-hull path.
- The remote results now support the implementation: `codecov/patch` is green, both `test` jobs are green, notebooks are green, docs are green, and `prek` is green.

I also ran the two most relevant local test targets from this follow-up:

- `pytest causalpy/tests/test_deprecated_design_aliases.py -q`
- `pytest causalpy/tests/test_maketables_plugin.py -q`

Both passed locally. The standalone invocations tripped the repo-wide coverage floor, which is expected for narrow pytest runs here and not a concern for the PR itself.

Residual note, not a blocker: the shared dataset helper currently covers the standard `X`/`y` layout but not the `SyntheticControl` `control`/`treated` layout, so there is still some experiment-specific wrapping code there. That no longer looks like problematic duplication to me; the important compatibility and warning-forwarding boilerplate is now centralized.

This satisfies my earlier objections.

causalpy/experiments/base.py
```diff
@@ -114,6 +117,67 @@ class BaseExperiment(ABC):
 
     _default_model_class: type[PyMCModel] | None = None
 
+    _deprecated_design_aliases: dict[str, tuple[str, str]] = {}
```
cetagostini, line 127: Quick one — `InversePropensityWeighting`, `InstrumentalVariable`, and `StaggeredDiD` didn't get migrated to `design` (still using numpy `self.X` / `self.y`, see e.g. `causalpy/experiments/inverse_propensity_weighting.py:111`, `instrumental_variable.py:155`, `staggered_did.py:332-338`) and accordingly don't show up in `_deprecated_design_aliases` here. Assuming intentional given their non-standard layouts (IPW has `t` + outcome, IV has `Z` + endogenous treatment, staggered has the train/full split)? If so, maybe worth a sentence in the PR body so it's not mistaken for an oversight.

causalpy/experiments/synthetic_control.py
```diff
@@ -126,12 +132,12 @@ def _check_convex_hull(self) -> None:
...
         n_units = len(self.treated_units)
-        n_pre_points = self.datapre_treated.shape[0]
+        n_pre_points = self.pre_design["treated"].shape[0]
 
         for i in range(n_units):
             unit_check = check_convex_hull_violation(
-                self.datapre_treated.isel(treated_units=i).values,
-                self.datapre_control.values,
+                self.pre_design["treated"].isel(treated_units=i).values,
+                self.pre_design["control"].values,
```
cetagostini: Heads-up on the input convention: this call site converts to numpy via `.values`, but the new `ConvexHullCheck.run` in `causalpy/checks/convex_hull.py:60-62` hands the same data to `check_convex_hull_violation` as `xarray.DataArray` directly (no `.values`). Both work — I checked end-to-end with violation cases and got identical results — but it's worth normalizing on one style across the two call sites and adding a one-liner in the helper's docstring saying it accepts numpy or xarray. Otherwise this becomes a quiet footgun the first time someone changes the helper to assume one shape.

causalpy/checks/convex_hull.py
```diff
@@ -50,8 +50,8 @@ def run(
     ) -> CheckResult:
         """Run the convex hull violation check on pre-treatment data."""
         sc = experiment
-        datapre_control = sc.datapre_control  # type: ignore[attr-defined]
-        datapre_treated = sc.datapre_treated  # type: ignore[attr-defined]
+        datapre_control = sc.pre_design["control"]  # type: ignore[attr-defined]
+        datapre_treated = sc.pre_design["treated"]  # type: ignore[attr-defined]
```
cetagostini: Tiny nit: since `applicable_methods = {SyntheticControl}` and `validate()` already enforces the type, you could drop both `# type: ignore[attr-defined]` markers by adding `assert isinstance(experiment, SyntheticControl)` (or just calling `self.validate(experiment)`) at the top of `run()` — that narrows `sc` to `SyntheticControl` and `pre_design` becomes a known attribute for mypy. Same effect, no escape hatches.

causalpy/pymc_models.py
```diff
@@ -253,18 +258,24 @@ def build_model(
...
-        in the future we get more complex models with more data nodes, then we'll need
-        to update all of them - ideally programmatically.
+        The base implementation updates mutable data nodes named ``"X"`` and
+        ``"y"``.  Subclasses that use different data nodes should override this
+        method.  See :class:`BayesianBasisExpansionTimeSeries` for an example.
         """
+        for name in ("X", "y"):
+            if name not in self.named_vars:
+                raise ValueError(
+                    f"Data node '{name}' not found in model. "
```
cetagostini, line 304: Optional polish: the class docstring above already points users at `BayesianBasisExpansionTimeSeries` as a concrete override example. Worth echoing that in the runtime message itself so users hitting this in a notebook don't have to dig — e.g. `f"... override _data_setter() (see BayesianBasisExpansionTimeSeries for an example)."`. Pure DX nicety, not blocking.

causalpy/experiments/base.py
drbenvincent: Yes, intentional — IPW, IV, and StaggeredDiD have non-standard design layouts (IPW pairs covariates with a treatment vector, IV has an instrument matrix plus endogenous treatment, StaggeredDiD keeps a train/full split) that don't fit the shared X/y Dataset shape. I've added a sentence to the PR body making this explicit. Note that SyntheticDifferenceInDifferences (added on main in #823 after this branched) *did* fit the SC-style four-quadrant pattern, so it has now been migrated to pre_design/post_design with the same deprecated aliases.

causalpy/experiments/synthetic_control.py
drbenvincent: Good catch — normalized both call sites on xarray: SyntheticControl._check_convex_hull now passes the DataArrays directly (no .values), matching ConvexHullCheck.run, and check_convex_hull_violation's docstring and type hints now state it accepts np.ndarray or xr.DataArray.

causalpy/checks/convex_hull.py
drbenvincent: Done — run() now calls self.validate(experiment) followed by an isinstance assert to narrow the type, and both type: ignore markers are gone. mypy passes via prek.

causalpy/pymc_models.py
drbenvincent: Done — the runtime message now ends with "override _data_setter() (see BayesianBasisExpansionTimeSeries for an example)."

PR #851, opened 2026-04-15
Implement summary method for InstrumentalVariable

### LINKED ISSUE #360: Implement `summary` method for `cp.pymc_experiments.InstrumentalVariable`
At the moment `cp.pymc_experiments.InstrumentalVariable` does not have an implementation of the `summary` method. It would be good to add this, and to add test coverage (e.g. in `test_iv_reg` in `test_integration_pymc_examples.py`)

Tagging @NathanielF - feel free to assign yourself if you want to take it up :)

anevolbap:
## Summary

Implement the `summary` method for `InstrumentalVariable`, replacing the `NotImplementedError` stub.

Fixes #360

## Approach

Issue #360 floated overriding the base `PyMCModel.print_coefficients` method. This PR overrides `summary` directly instead, because IV has five distinct coefficient blocks that don't fit the single-stage shape of `print_coefficients`:

1. Naive OLS coefficients
2. 2SLS first-stage coefficients
3. 2SLS second-stage coefficients
4. Bayesian instrument-stage coefficients (`beta_t` over `instruments`)
5. Bayesian outcome-stage coefficients (`beta_z` over `covariates`)

The base `print_coefficients` assumes one `beta` variable. Reusing it would require either splitting the call into multiple shapes or branching inside the base, both of which leak IV-specific structure into the parent class. Inlining the formatting in `InstrumentalVariable.summary` keeps the base unchanged.

The Bayesian section uses the same `HDI` label and quantile computation pattern as the base `print_coefficients` (in `causalpy/pymc_models.py`). Whether to switch the whole codebase to `az.hdi` is out of scope for this PR.

## Changes

- `InstrumentalVariable.summary()` prints OLS, 2SLS, and Bayesian coefficient blocks (instrument and outcome stages with HDI-style intervals).
- Removed `"summary"` from `test_iv_not_implemented_methods` parametrize list.
- Added `result.summary()` call to `test_iv_reg`.

## Testing

- All 4 IV integration tests pass.
- Pre-commit hooks pass on the latest commit (the earlier mypy failure was caused by the now-merged `panel_regression` issue fixed in #853).


drbenvincent:
This PR's failing `test (3.11)` / `test (3.14)` jobs are caused by `test_panel_regression_plot_unit_effects` (`TypeError: only 0-dimensional arrays can be converted to Python scalars`), which was fixed in #853 (now merged to `main`). Updating this branch from `main` should turn CI green — no code changes needed here.

anevolbap:
@drbenvincent this is ready for review. CI fully green after merging `main`. Just pushed one fix (`63f5f26a`) so `round_to=0` is no longer overridden to 2.

Two things to call out for review:

1. **Override target**: in #360 you suggested overriding `print_coefficients`. I went with overriding `summary` directly because IV has five coefficient blocks (OLS, 2SLS first stage, 2SLS second stage, Bayesian instrument stage, Bayesian outcome stage), and the base `print_coefficients` assumes a single `beta` variable. The PR description has the full reasoning. Happy to change to a custom `print_coefficients` if you prefer.

2. **HDI labeling**: the Bayesian block uses the same `X% HDI [lo, hi]` label with equal-tailed quantiles that `PyMCModel.print_coefficients` uses. I matched the existing convention rather than diverging. If we want to switch the whole codebase to `az.hdi`, I can open a follow-up issue.

Thanks!

NathanielF (approved):
Looks good to me.

causalpy/tests/test_instrumental_variable.py
```diff
@@ -399,7 +399,7 @@ def test_iv_coords_set(iv_data, sample_kwargs):
 # =============================================================================
 
 
-@pytest.mark.parametrize("method", ["plot", "summary", "effect_summary"])
+@pytest.mark.parametrize("method", ["plot", "effect_summary"])
```
anevolbap, line 402: The CI failure in `test_three_period_its.py::test_analyze_persistence_sklearn` is triggered by this line, but the bug is not here. The `rng` fixture in `conftest.py` is session-scoped and `iv_data` draws from it. Removing one parametrize case means one fewer `iv_data` call, which shifts the shared random stream for every test collected after this file. `datetime_data` in `test_three_period_its.py` builds its noise from that stream, and its `persistence_ratio >= 0` assertion only holds for some draws.

I reproduced this on main: running `test_instrumental_variable.py` followed by `test_analyze_persistence_sklearn` fails there too, so the test is order-dependent and pre-existing. I will fix it in a separate PR (give `datetime_data` its own seeded generator) and rerun checks here.

drbenvincent:
Thanks @anevolbap

PR #853, opened 2026-04-16
Fix mypy type: ignore codes in panel_regression.py

### LINKED ISSUE #852: mypy: wrong type: ignore code in panel_regression.py breaks pre-commit
## Description

`panel_regression.py` has two bugs:

1. Lines 560 and 676 use `# type: ignore[attr-defined]`, but mypy emits `[union-attr]` errors. The mismatched code means the ignores have no effect, and mypy fails during pre-commit.

2. `plot_unit_effects` (line 683) uses `float(...)` on an xarray result that can have a leftover `treated_units` dimension, causing `TypeError: only 0-dimensional arrays can be converted to Python scalars`. This makes `test_panel_regression_plot_unit_effects` fail on main (Python 3.14).

## Root cause

1. Introduced in commit 1cf3387 (PR #628). `self.idata` can be `None`, so accessing `.posterior` is a `union-attr` error, not `attr-defined`.
2. `beta.sel(...).mean(dim=["chain", "draw"])` returns an array with a `treated_units` dimension, not a scalar.

## Fix

1. Replace `[attr-defined]` with `[union-attr]` on lines 560 and 676.
2. Replace `float(...)` with `.item()` on line 683.

anevolbap:
## Summary

Fix two bugs in `panel_regression.py`:
1. Wrong `type: ignore` error codes that cause mypy to fail during pre-commit
2. `plot_unit_effects` crashes with `TypeError` when posterior has a `treated_units` dimension

Fixes #852

## Changes

- `panel_regression.py:560` -- `[attr-defined]` replaced with `[union-attr]`
- `panel_regression.py:676` -- `[attr-defined]` replaced with `[union-attr]`
- `panel_regression.py:683` -- `float(...)` replaced with `.squeeze("treated_units", drop=True).item()` to handle non-0d xarray result, matching existing idiom at lines 852-858

## Testing

- `pre-commit run --files causalpy/experiments/panel_regression.py` passes (mypy green)
- `test_panel_regression_plot_unit_effects` fails on main with the same `TypeError`; the `.item()` fix resolves it
- Test hardened with `filterwarnings` to promote the ndim>0 scalar `DeprecationWarning` to an error, catching this regression on all Python versions

---
*Edit (2026-04-17): addressed review feedback -- added `.squeeze("treated_units", drop=True)` and `filterwarnings` test hardening.*

### DIFF
--- a/causalpy/experiments/panel_regression.py
+++ b/causalpy/experiments/panel_regression.py
@@ -557,7 +557,7 @@ class PanelRegression(BaseExperiment):
         """
         # Get posterior predictions
         if isinstance(self.model, PyMCModel):
-            mu = self.model.idata.posterior["mu"]  # type: ignore[attr-defined]
+            mu = self.model.idata.posterior["mu"]  # type: ignore[union-attr]
             pred_mean = mu.mean(dim=["chain", "draw"]).values.flatten()
             pred_lower = mu.quantile(0.025, dim=["chain", "draw"]).values.flatten()
             pred_upper = mu.quantile(0.975, dim=["chain", "draw"]).values.flatten()
@@ -673,7 +673,7 @@ class PanelRegression(BaseExperiment):
 
         if isinstance(self.model, PyMCModel):
             # Bayesian: get posterior means
-            beta = self.model.idata.posterior["beta"]  # type: ignore[attr-defined]
+            beta = self.model.idata.posterior["beta"]  # type: ignore[union-attr]
             unit_fe_indices = [self.labels.index(name) for name in unit_fe_names]
 
             # Get mean and std for each unit FE
--- a/causalpy/experiments/panel_regression.py
+++ b/causalpy/experiments/panel_regression.py
@@ -680,7 +680,7 @@ class PanelRegression(BaseExperiment):
             fe_means = []
             for idx in unit_fe_indices:
                 fe_means.append(
-                    float(beta.sel(coeffs=self.labels[idx]).mean(dim=["chain", "draw"]))
+                    beta.sel(coeffs=self.labels[idx]).mean(dim=["chain", "draw"]).item()
                 )
 
             ax.hist(
--- a/causalpy/experiments/panel_regression.py
+++ b/causalpy/experiments/panel_regression.py
@@ -680,7 +680,10 @@ class PanelRegression(BaseExperiment):
             fe_means = []
             for idx in unit_fe_indices:
                 fe_means.append(
-                    beta.sel(coeffs=self.labels[idx]).mean(dim=["chain", "draw"]).item()
+                    beta.sel(coeffs=self.labels[idx])
+                    .mean(dim=["chain", "draw"])
+                    .squeeze("treated_units", drop=True)
+                    .item()
                 )
 
             ax.hist(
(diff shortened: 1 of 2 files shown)


causalpy/experiments/panel_regression.py
```diff
@@ -673,14 +673,14 @@ def plot_unit_effects(
...
-            beta = self.model.idata.posterior["beta"]  # type: ignore[attr-defined]
+            beta = self.model.idata.posterior["beta"]  # type: ignore[union-attr]
             unit_fe_indices = [self.labels.index(name) for name in unit_fe_names]
 
             # Get mean and std for each unit FE
             fe_means = []
             for idx in unit_fe_indices:
                 fe_means.append(
-                    float(beta.sel(coeffs=self.labels[idx]).mean(dim=["chain", "draw"]))
+                    beta.sel(coeffs=self.labels[idx]).mean(dim=["chain", "draw"]).item()
```
drbenvincent: For consistency with the existing pattern in this file (lines 852-858 already do `.squeeze("treated_units", drop=True)` on posterior slices), consider making the scalar extraction explicit rather than relying on `.item()` silently collapsing a size-1 array:

```suggestion
                    beta.sel(coeffs=self.labels[idx])
                    .mean(dim=["chain", "draw"])
                    .squeeze("treated_units", drop=True)
                    .item()
```

Functionally identical today (since `treated_units` is always `["unit_0"]`), but self-documenting and robust if that dimension is ever generalised.

drbenvincent (commented):
Thanks for the quick turnaround on this, @anevolbap — I validated locally and can confirm `main` is red on both `prek` (mypy) and on `test_panel_regression_plot_unit_effects` under strict `DeprecationWarning` handling; this branch fixes both cleanly, all 27 `test_panel_regression.py` tests pass, and `prek run --all-files` is green. The swap from `[attr-defined]` to `[union-attr]` is the right diagnosis (`self.model.idata: InferenceData | None` is an Optional, not a missing attribute), and `.item()` correctly collapses the residual `treated_units` dim that broke `float()` on Py3.14.

Two small follow-ups worth folding in — both non-blocking:

**1. Harden the regression test so this class of bug can't sneak back in.** The existing `test_panel_regression_plot_unit_effects` only asserts `isinstance(fig, plt.Figure)`, which is exactly why the `float(ndim>0)` call went unnoticed on Python versions where it was still a `DeprecationWarning` rather than a `TypeError`. Promoting that specific deprecation to an error on this test makes the regression catchable on every Python version, not just 3.14:

```python
@pytest.mark.integration
@pytest.mark.filterwarnings(
    "error:Conversion of an array with ndim > 0 to a scalar:DeprecationWarning"
)
def test_panel_regression_plot_unit_effects(mock_pymc_sample, small_panel_data):
```

Scoped narrowly to the xarray/NumPy message so it won't trip on unrelated warnings. Happy for this to be a follow-up PR if you'd rather keep this one minimal.

**2. Match the file's existing idiom for the `treated_units` dim** — see inline comment on line 683.

anevolbap:
Thanks for the thorough review @drbenvincent! Both suggestions addressed:

- `fbe8f71` — `.squeeze("treated_units", drop=True)` before `.item()` to match the existing pattern at lines 852–858
- `c062136` — `@pytest.mark.filterwarnings("error:Conversion of an array with ndim > 0 to a scalar:DeprecationWarning")` on the test so the regression is caught on all Python versions

drbenvincent (approved):
Thanks!

Issue #637, opened 2026-01-07
Add All Contributors bot for contributor recognition


drbenvincent:
# Add All Contributors bot for contributor recognition

## Summary

Implement the [All Contributors](https://allcontributors.org) specification and bot to maintain an automated contributor gallery in our README. This recognizes all types of contributions—not just code—and helps build community by making contributors feel valued.

## Motivation

Currently, contributors are only visible through GitHub's default contributors graph (which only shows code commits). This misses valuable contributions like:

- Documentation improvements
- Bug reports and triage
- Ideas and feature suggestions
- Code review
- Examples and tutorials
- Community support

A visible contributor gallery in the README:
- **Recognizes all contributors** publicly
- **Encourages participation** (people see their name/avatar in the project)
- **Builds community** by showing the project is collaborative
- **Complements our new welcome workflow** (PR #636)

## What is All Contributors?

A GitHub bot + specification used by 2000+ projects (including formik, p5.js, snipe-it). It:

1. Maintains a contributor table in the README
2. Uses an emoji key to show contribution types (💻 code, 📖 docs, 🐛 bugs, 💡 ideas, etc.)
3. Automates updates via GitHub comments (e.g., `@all-contributors add @user for docs`)

### Step 1: Install the GitHub App

- Install from: https://github.com/apps/allcontributors/installations/new
- Grant access to `pymc-labs/CausalPy`

### Step 2: Initialize the configuration

```bash
npx all-contributors-cli init
```

This creates:
- `.all-contributorsrc` — JSON config file
- Updates `README.md` with a contributors section placeholder

### Step 3: Add existing contributors

Retroactively add current contributors using the CLI or bot:

```bash
# Example
npx all-contributors-cli add drbenvincent code,doc,ideas
npx all-contributors-cli add contributor2 code,test
```

Or via GitHub comments after the bot is installed:
```
@all-contributors please add @drbenvincent for code, doc, ideas
```

### Step 4: Update CONTRIBUTING.md

Add a new section to `CONTRIBUTING.md` (after "Contributing code via pull requests") explaining:

```markdown
## Recognition of contributors

We use [All Contributors](https://allcontributors.org) to recognize all types of contributions—not just code. After your contribution is merged, a maintainer will add you to our contributors gallery in the README.

**Contribution types we recognize:**
- 💻 Code
- 📖 Documentation
- 🐛 Bug reports
- 💡 Ideas & suggestions
- ⚠️ Tests
- 👀 Code review
[...]

Issue #17, opened 2022-10-22
split the __init__ of the ExperimentalDesigns up


drbenvincent:
Split up the following into smaller methods:
- `TimeSeriesExperiment.__init__()`
- `DifferenceInDifferences.__init__()`
- `RegressionDiscontinuity.__init__()`

Issue #11, opened 2022-10-21
regression discontinuity: add quantitative output of the discontinuity estimate


drbenvincent:
- [x] Evaluate the model prediction either side of the threshold and report the discontinuity value.
- [x] Remove evaluation of the causal impact and cumulative causal impact. This makes less sense in the RD setting.
- [x] (Bayesian model) It makes more sense to be plotting the model expectation, not the posterior predictive. Similarly, we should calculate the discontinuity at the threshold from `mu`, not the `yhat`
- [x] Add method which prints text summary output about the discontinuity at the threshold
- [x] Improve/finish `summary` method for the Bayesian model. Might be best to create some get methods to avoid repeating this task multiple times.
- [x] Bayesian model: bring back a subplot, but this time plot the posterior distribution of the discontinuity at threshold. Or maybe best to create a separate plot function for this.

Issue #10, opened 2022-10-21
add control units to synthetic design plot


drbenvincent:
It would improve the plot if we add the untreated units to the plot (e.g. in light grey).

This will deviate from the `plot` method in the `TimeSeriesExperiment` class. So it's probably best to override this plot method where we call the superclass method then additionally plot the untreated units.

PR #856, opened 2026-04-17
Self-hosted contributors image + weekly regen workflow

drbenvincent:
## Summary

Adds a "Thanks to our contributors" section to the README, backed by an in-repo SVG regenerated weekly by a GitHub Actions workflow that opens a PR when the contributor set changes.

**Why not `contrib.rocks`?** The service stalled for this repo and upstream users report multi-week refresh latencies. This approach is fully self-hosted with no third-party runtime dependencies.

## What changes

- `README.md` — new "Thanks to our contributors" section linking to `docs/source/_static/contributors.svg`.
- `docs/source/_static/contributors.svg` — initial 33-contributor grid. Circular avatars in a 12-column grid, each linking to the contributor's profile. Avatars are base64-embedded so the SVG renders on GitHub (external `<image>` refs are blocked by GitHub's SVG CSP).
- `docs/source/_static/contributors.json` — sidecar manifest recording the sorted login set. Used to short-circuit regeneration when nothing changed.
- `.github/workflows/contributors.yml` — weekly cron (`0 6 * * 1`) + manual dispatch. `actions/github-script` fetches contributors, `peter-evans/create-pull-request` opens a PR on change. All third-party actions SHA-pinned.

## Grid ordering

Avatars appear in the order returned by the GitHub contributors API, which is **sorted by commit count descending**. The sidecar manifest uses a separate `localeCompare` sort purely for stable set-equality comparisons; it doesn't affect what you see.

## Hardening (from review)

- Per-avatar try/catch with a neutral grey placeholder, 15s request timeout, redirect handling, parallel fetches (concurrency=8) — a single flaky avatar no longer breaks the weekly cron.
- Host pin (`avatars.githubusercontent.com`) and MIME allowlist (png/jpeg/gif/webp) on the avatar fetch.
- Short-circuit via `contributors.json`: if the login set is unchanged, skip fetching and writing entirely. Avatar byte-level re-encodes no longer produce noise PRs.

## Reliability characteristics

- No third-party service at view time (SVG in repo, avatars fetched only at workflow run time).
- If the workflow breaks, the committed SVG keeps rendering; failure is visible in the Actions tab.
- `GITHUB_TOKEN` gives 1000 req/hr, trivially sufficient.
- Top-level `permissions: {}`; job-scoped `contents: write` + `pull-requests: write`; `persist-credentials: false` on checkout.

## Relationship to #637

Lightweight, automated recognition of **code contributors**. Does not replace the richer All Contributors bot proposed in #637 (docs, ideas, reviews, etc.); #637 can be layered on later.

## Test plan

- [ ] View the rendered README on the branch page — confirm the 33-avatar grid renders.
- [ ] Manually dispatch `Update contributors image` once merged — with the seeded manifest, the job should short-circuit and open no PR.
- [ ] `zizmor` passes in CI (clean locally).


drbenvincent:
## Critical review — security and reliability

_Posted by: claude-opus-4-7-xhigh_

### Summary of what I verified

- All three third-party action SHAs are legitimate and match their claimed tags:
  - `actions/checkout@de0fac2e…` = v6.0.2 ✓
  - `actions/github-script@ed597411…` = v8.0.0 ✓ (matches `welcome.yml`)
  - `peter-evans/create-pull-request@5f6978fa…` = v8.1.1 ✓
- `zizmor` CI check passes on the PR.
- SHA pinning pattern matches the repo's existing conventions (`.github/workflows/zizmor.yml`, `welcome.yml`).
- `permissions: {}` at top level and scoped `contents: write` + `pull-requests: write` at job level is correct.
- `persist-credentials: false` on checkout is correct.

The PR description is unusually accurate — nothing misleading. Below are the issues I still think deserve attention, sorted by severity.

---

### Security

#### 1. Sanity-check overall: low risk surface

Important framing: this workflow only triggers on `schedule` and `workflow_dispatch`. It never runs on `pull_request` or `pull_request_target`. That eliminates the entire class of "malicious PR pwns our CI" attacks. No untrusted input ever reaches this workflow.

That said, there are still a few things worth flagging.

#### 2. No SSRF pin on the avatar URL host — LOW

`fetchDataUri()` is called with `${c.avatar_url}?s=128`, where `avatar_url` comes from the GitHub contributors API. In practice GitHub always returns `https://avatars.githubusercontent.com/...`, but the code doesn't validate this. A compromised or spoofed response could in theory redirect the fetch elsewhere.

Mitigations are already quite strong:
- GitHub API is itself trusted (same threat model as every workflow using `GITHUB_TOKEN`).
- The fetched content only becomes base64 bytes inside an SVG — it isn't executed.
- GitHub's SVG serving strips scripts, so even a crafted PNG/SVG payload wouldn't execute.

**Suggestion (optional, nice-to-have):** add an explicit host allowlist:

```js
const u = new URL(url);
if (u.host !== 'avatars.githubusercontent.com') {
  return reject(new Error(\`Unexpected avatar host: \${u.host}\`));
}
```

Defence-in-depth, not a real vulnerability.

#### 3. No redirect handling + no timeout — LOW reliability, not security

`https.get` does not follow redirects and has no timeout. If `avatars.githubusercontent.com` ever returns a 302, every avatar would fail. No timeout means a single hung connection could stall the entire job up to the 6-hour GitHub default.

**Fix:** set a request timeout, e.g.

```js
req.setTimeout(15000, () => req.destroy(new Error(\`timeout for \${url}\`)));
```

and either accept 301/302 with one retry, or log & skip a failing avatar rather than aborting the whole run.

#### 4. Data URIs without content-type validation — LOW

`contentType` is taken from the response header and written straight into the `data:` URI. A malicious server could return `content-type: text/html` and the SVG would embed `<image href=\"data:text/html;base64,...\">`. GitHub's SVG sanitiser strips scripts, and `<image>` only renders bitmap content regardless of declared type, so this doesn't lead to execution. Still, a strict allowlist is cleaner:

```js
const ct = res.headers['content-type'] || '';
if (!['image/png', 'image/jpeg', 'image/gif', 'image/webp'].includes(ct)) {
  return reject(new Error(\`Unexpected content-type \${ct} for \${url}\`));
}
```

#### 5. Raw README link points at `main` — LOW

Anyone with write access to `main` can replace the SVG at any time and it will render on the project's README (GitHub does aggressively cache SVGs, which slightly mitigates). Same trust model as any other in-repo asset, so acceptable. Just worth stating: reviewers of the auto-generated PR should not rubber-stamp it.

---

### Reliability

#### 6. All-or-nothing rendering — MEDIUM (my biggest practical concern)

The current loop is strictly sequential and has no error handling. If avatar #17 of 33 returns a 5xx, the whole workflow run fails and no PR is opened. Because this runs weekly on a cron, that silently means \"contributor image never refreshes until someone notices\". The README still renders the last committed SVG, which is why the impact is bounded — but the \"weekly auto-PR\" story in the PR description quietly degrades.

**Suggested fix (small):** wrap each `fetchDataUri` in try/catch; on failure either skip that contributor or fall back to a neutral placeholder circle. Also: parallelise with `Promise.all` or a small concurrency limiter — 33 serial HTTPS fetches is slow for no reason.

#### 7. `listContributors` returns at most 500 + excludes anonymous by default — LOW

The endpoint caps at the top 500 contributors. Non-issue for a 33-contributor project today, but worth a one-line comment if the project ever grows large.

#### 8. Bot filter is mostly redundant — LOW

You're already filtering on `c.type === 'User'`, which excludes `Bot`-type accounts (GitHub App bots show up as type `Bot`). The regex is mostly belt-and-braces, which is fine — just noting it.

#### 9. Whitespace/byte-level diffs may re-open PRs — LOW

`peter-evans/create-pull-request` opens a new PR whenever the SVG content differs. Since the SVG has 33 data URIs fetched fresh each run, any time GitHub re-encodes an avatar with different bytes, a diff will appear even when no contributor changed. The PR description already accepts this (\"weekly PR to approve\"), but consider rounding by storing only the `login` set in a sidecar file and short-circuiting generation if the set hasn't changed.

#### 10. `labels: documentation` may fail silently — TRIVIAL

If the `documentation` label doesn't exist, the action warns (not fails) and opens the PR without the label. Worth a one-line check before merge, or just swap for an existing label.

#### 11. SVG size scales linearly with contributor count — LOW

Each embedded 128px avatar is ~5–15 KB base64-encoded. At 33 contributors the SVG is 12 KB; at 500 it'd be 2–6 MB. Fine for the foreseeable future; worth being aware the approach doesn't scale to a PyMC-sized community.

---

### Correctness / code-review nits

- `xlink` namespace is declared on `<svg>` root but never used (code uses `href`, not `xlink:href`). Harmless but dead.
- `rel=\"noopener\"` on the links is correct; consider `rel=\"noopener noreferrer\"` to also strip referrer info.
- The inline comment says GitHub \"strips `<image href=\"https://...\">`\" — more accurately, GitHub's SVG renderer blocks external-origin image references via CSP. Behaviour is correct; phrasing is slightly off.
- `preserveAspectRatio=\"xMidYMid slice\"` with `clipPath` ensures non-square avatars centre-crop nicely — good touch.

---

### What I would require before merge

Nothing is blocking in a security sense. The one genuine change I'd ask for is **(6) per-avatar error handling**, because without it a single flaky avatar fetch breaks the weekly cron and you'll quietly find out in six months that the image is stale.

Everything else ((2), (3), (4), (8)–(11)) is polish and defence-in-depth. I'd happily approve this as-is if you're comfortable accepting that the workflow will simply fail visibly in the Actions tab when GitHub has a bad day — consistent with the \"committed SVG keeps rendering; workflow fails visibly\" guarantee already stated in the PR description.

### Verdict

**Approve with a strong nudge on reliability fix (6), and optional polish on (2)/(3)/(4).** The security posture of this PR is genuinely good — better than most CI changes. The reliability failure modes are the more realistic risk.

The pinning is correct. The permissions model is correct. The attack surface is small because the trigger is cron-only. The design choice (self-host the SVG, regenerate weekly) is the right call for this project, especially given the `contrib.rocks` reliability issues cited.

drbenvincent:
Addressing review in a sequence of commits. Commit 1: review points **#6** and **#3** — per-avatar try/catch with a neutral placeholder on failure, 15s request timeout, redirect handling (301/302/307/308, up to 3 hops), and parallel fetches with concurrency=8. A single flaky avatar now logs a warning and keeps going instead of breaking the weekly cron.

drbenvincent:
Commit 2: review points **#2** and **#4** — defence-in-depth allowlists. `fetchDataUri` now rejects URLs whose host isn't `avatars.githubusercontent.com` (so a spoofed redirect can't send us off-net) and rejects responses whose content-type isn't a raster image (`png`/`jpeg`/`gif`/`webp`). The validated MIME is what gets emitted in the `data:` URI, not the server's raw header.

drbenvincent:
Commit 3: correctness nits from review. Dropped the unused `xmlns:xlink` namespace, strengthened the avatar link `rel` to `noopener noreferrer`, and reworded the workflow's CSP comment to match what GitHub actually does (external `<image>` refs are *blocked* by CSP, not *stripped*). Also confirmed the `documentation` label (#10) exists on the repo so no silent-fail there.

drbenvincent:
Commit 4: review point **#9** — sidecar manifest to short-circuit unchanged runs. The workflow now maintains `docs/source/_static/contributors.json` with a sorted login list. At the top of each run we compare against the committed manifest: if the set is unchanged we skip avatar fetching and SVG writing entirely, so avatar PNG re-encodes on GitHub's side no longer produce weekly noise PRs. The manifest is seeded with the 33 current logins so the first scheduled run starts clean.

drbenvincent:
Commit 5: review points **#7**, **#8**, **#11** — inline comments documenting the 500-contributor cap on `listContributors` (plus anonymous exclusion), the rationale for keeping the BOT_PATTERN regex as belt-and-braces alongside `c.type === 'User'`, and that SVG size scales linearly with contributor count.

That covers every point raised in the review (#2, #3, #4, #6 with genuine behaviour changes; #7, #8, #10, #11, and the correctness nits with lighter-touch changes). **#1** and **#5** were framing/trust-model context rather than actionable items.

williambdean:
Screenshot 2026-04-17 at 4 32 31 PM

Doesn't render in the branch's README

drbenvincent:
This is allegedly an svg issue. But I'll dig into it more

cetagostini (approved):

williambdean:
> This is allegedly an svg issue. But I'll dig into it more

I see the svg file is fine while viewing from the repo.

drbenvincent:
This PR's failing `test (3.11)` / `test (3.14)` jobs are caused by `test_panel_regression_plot_unit_effects` (`TypeError: only 0-dimensional arrays can be converted to Python scalars`), which was fixed in #853 (now merged to `main`). Updating this branch from `main` should turn CI green — no code changes needed here.

PR #872, opened 2026-04-28
Add RD/RKink sensitivity-checks walkthrough

drbenvincent:
Closes #790

## Summary

Follow-up to #785: documents how to use the pipeline-style sensitivity checks for {term}\`Regression discontinuity design\` and {term}\`Regression kink design\` workflows, mirroring the staggered-DiD walkthrough that landed in \`staggered_did_pymc.ipynb\`.

- New **Sensitivity analysis** section in \`rd_pymc.ipynb\` with a Related-checks summary, numbered walkthrough, executed pipeline cell running \`BandwidthSensitivity\` + \`McCraryDensityTest\`, an HTML-report iframe, pass/fail interpretation, and an *If these checks fail* troubleshooting block.
- Refresh of the existing sensitivity section in \`rkink_pymc.ipynb\` with the same structure: links to \`pipeline_workflow\` and the central \`sensitivity_checks\` guide, a Related-checks block clarifying that \`McCraryDensityTest\`, \`PlaceboInTime\`, \`LeaveOneOut\`, and \`PlaceboInSpace\` are not applicable to RKink, a numbered walkthrough, pass/fail interpretation, and an *If this check fails* troubleshooting block.
- Docs only — no code changes.

## Notes

- The bandwidth sweep in \`rd_pymc.ipynb\` uses the linear interaction formula \`y ~ 1 + x + treated + x:treated\` rather than basis splines, with an inline note explaining why: \`bs(x, df=6)\` bakes its knots into the full data and patsy raises an error if a narrower bandwidth subset asks the model to evaluate observations that fall outside those baked-in outer knots.
- The McCrary check fails on the bundled \`rd\` dataset purely because the running variable spans \`[-1, 1]\` while the threshold sits at \`0.5\`. The notebook flags this explicitly so readers don't read the failure as evidence of manipulation in the example data.

## Test plan

- [x] \`prek run --all-files\` passes (notebook schema, ruff format, codespell, mypy, BibTeX, etc.)
- [x] Both notebooks validate with \`nbformat.validate\`
- [x] \`rd_pymc.ipynb\` sensitivity pipeline cell executes successfully (real outputs committed)
- [ ] CI passes on this PR
- [ ] Visual review of the rendered docs page

Made with [Cursor](https://cursor.com)

PR #871, opened 2026-04-28
Docs: Add Synthetic Control sensitivity checks walkthrough

drbenvincent:
## Summary

Adds a focused sensitivity-check walkthrough for Synthetic Control to the `sc_pymc.ipynb` notebook, addressing #789 (follow-up to #749 / #785).

This continues the docs work started in #818 (central guide) and #846 (Staggered DiD walkthrough), keeping the same structure:

- **Sensitivity analysis section** introducing the pipeline API (`EstimateEffect` → `SensitivityAnalysis` → `GenerateReport`) and linking back to the central guide and `pipeline_workflow`.
- **Related checks for synthetic control** subsection enumerating the SC-applicable checks (`PlaceboInTime`, `ConvexHullCheck`, `LeaveOneOut`, `PlaceboInSpace`, `PriorSensitivity`) with one-line interpretations and noting which checks belong to other estimators.
- **`PlaceboInTime` deep-dive** with how-it-works narrative grounded in citations (`abadie2010synthetic`, `abadie2021using`, `reichardt2019quasi`).
- **Walkthrough** with numbered steps and pass/fail interpretation guidance.
- **"If this check fails"** subsection with five concrete diagnostic actions: inspect fold summaries, prune the donor pool, add complementary SC checks, sanity-check the convex-hull condition, probe prior sensitivity.
- **Pipeline code cell** running `PlaceboInTime(n_folds=2, random_seed=seed)` on the existing SC dataset together with `GenerateReport(include_plots=True)`.
- **HTML report iframe** display cell, matching the pattern used in `staggered_did_pymc.ipynb`.

## Notes

- The walkthrough deliberately demonstrates a `NOT SUPPORTED` placebo-in-time result for the demo dataset (P=0.905, just below the 0.95 default threshold). This ties directly into the "If this check fails" troubleshooting section so readers see the failure path and the next steps.
- Existing cells were left untouched; the new cells were appended after the effect-summary section. The previous trailing empty placeholder cell was removed as part of inserting the new content.

Closes #789

## Test plan

- [x] `prek run --files docs/source/notebooks/sc_pymc.ipynb` passes
- [x] `python scripts/validate_notebooks.py docs/source/notebooks/sc_pymc.ipynb` passes
- [x] Notebook re-executed end-to-end via `jupyter nbconvert --to notebook --execute --inplace ...` so all outputs are populated
- [ ] CI notebooks job executes the SC notebook without errors
- [ ] ReadTheDocs preview renders the new sections correctly

[...]

PR #878, opened 2026-04-30
ci: pin pyproject2conda==0.22.1 to work around 0.23.0 regression

drbenvincent:
## Summary

The `check-environment-yml` CI job has been failing on every PR opened since **2026-04-29** with:

```
File ".../pyproject2conda/_normalized_requirements.py", line 115, in _iter_parts
    if self.channel:
       ^^^^^^^^^^^^
AttributeError: 'CondaRequirement' object has no attribute 'channel'
```

## Root cause

The job does `pip install pyproject2conda` with **no version constraint** ([ci.yml:51-52](https://github.com/pymc-labs/CausalPy/blob/main/.github/workflows/ci.yml#L51-L52)), so each CI run grabs whatever is latest on PyPI.

`pyproject2conda 0.23.0` was published to PyPI on 2026-04-29 and contains a regression where `CondaRequirement` instances no longer expose the `channel` attribute that `_iter_parts()` still tries to read. This is hit during `__hash__` of any conda requirement, so the tool blows up immediately when called with the args our workflow uses.

The local prek hook in `.pre-commit-config.yaml` is already pinned to `v0.22.1` (the last good release), so local `prek run --all-files` is unaffected — only CI breaks.

Filed upstream as [usnistgov/pyproject2conda#94](https://github.com/usnistgov/pyproject2conda/issues/94).

## Fix

Pin the CI `pip install` to `>=0.22.1,<0.23` so it tracks the same 0.22.x line as the local prek hook (`rev: v0.22.1`) while still allowing a hypothetical 0.22.x patch backport to ship without a workflow edit. Added an inline comment explaining the failure mode so future bumps remember to lift CI and the prek hook in lockstep.

## Why pin instead of work around?

The traceback lands in `CondaRequirement.__hash__` → `_iter_parts()`, which is invoked unconditionally as soon as `pyproject2conda` materialises any conda requirement. There is no plausible CLI flag, env extra, or `pyproject.toml` shape change that avoids that code path while still producing a usable conda yaml — the broken attribute access fires before any of our arguments matter. Pinning is therefore the right lever; alternative workarounds were not pursued because they cannot exist for this failure mode.

## Evidence this is environmental, not PR-specific

- PR #871 (`docs/789-sc-sensitivity-walkthrough`) has `check-environment-yml` failing despite touching only a single `.ipynb` file (no `pyproject.toml` or `environment.yml` changes).
- PR #872's last successful `check-environment-yml` run was on 2026-04-28 — it picked up `pyproject2conda 0.22.1` *before* 0.23.0 was published.
- All other open PRs that haven't been pushed to since 2026-04-29 still show the check as green.

## Test plan

- [ ] CI `check-environment-yml` job passes on this PR.
- [ ] After merge, rebase #871 onto main and confirm `check-environment-yml` goes green there too.

## Follow-up

When pyproject2conda ships a fixed release, bump both `.pre-commit-config.yaml` (`rev: v0.22.1` → newer) and `.github/workflows/ci.yml` (the version range → newer) together — or, preferably, merge #882 first so there is only one place to bump.

Tracked separately:

- Upstream regression report: [usnistgov/pyproject2conda#94](https://github.com/usnistgov/pyproject2conda/issues/94)
- Consolidating the CI invocation through the prek hook (removes the dual-pin surface): #879, implemented in #882
- Hardening the `tail -n +10` header comparison in this same job: #880
- Auditing pip-install steps in CI for consistent supply-chain pinning: #881

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,12 @@ jobs:
         with:
           python-version: "3.12"
       - name: Install pyproject2conda
-        run: pip install pyproject2conda
+        # Pinned to match the local prek hook rev (v0.22.1). pyproject2conda
+        # 0.23.0 (released 2026-04-29) regressed with
+        # `AttributeError: 'CondaRequirement' object has no attribute 'channel'`.
+        # Bump in lockstep with the .pre-commit-config.yaml rev once a fixed
+        # release is available.
+        run: pip install "pyproject2conda==0.22.1"
       - name: Regenerate environment.yml (same args as local prek hook config)
         run: |
           pyproject2conda yaml -e dev -o /tmp/environment.generated.yml -n CausalPy \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,12 +49,15 @@ jobs:
         with:
           python-version: "3.12"
       - name: Install pyproject2conda
-        # Pinned to match the local prek hook rev (v0.22.1). pyproject2conda
-        # 0.23.0 (released 2026-04-29) regressed with
-        # `AttributeError: 'CondaRequirement' object has no attribute 'channel'`.
-        # Bump in lockstep with the .pre-commit-config.yaml rev once a fixed
-        # release is available.
-        run: pip install "pyproject2conda==0.22.1"
+        # Pinned to the 0.22.x line to match the local prek hook rev (v0.22.1).
+        # 0.23.0 introduced a regression:
+        # `AttributeError: 'CondaRequirement' object has no attribute 'channel'`
+        # raised from `_iter_parts` during `__hash__` of any conda requirement,
+        # so any invocation that processes conda deps fails immediately.
+        # The upper bound allows a hypothetical 0.22.x patch backport while
+        # excluding 0.23.x. Bump in lockstep with the .pre-commit-config.yaml
+        # rev once a fixed release is available.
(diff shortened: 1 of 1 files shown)


cetagostini (approved):

drbenvincent:
Updated this PR off the back of an adversarial review. Two changes in `0de4a8e`:

1. **Widened the pin from `==0.22.1` to `>=0.22.1,<0.23`.** The strict equality was unnecessarily rigid — if upstream ever ships a `0.22.2` patch backport (e.g. a security fix), the range pin picks it up automatically while still excluding the broken 0.23.x line. The lower bound matches the prek hook's `rev: v0.22.1`; the upper bound is the meaningful one.

2. **Dropped the hardcoded release date from the inline comment** (`released 2026-04-29`). The version identifier `0.23.0` is the durable reference and `git blame` recovers the date if anyone needs it. Replaced that line with a more useful description of the failure mode (`_iter_parts` → `__hash__`) so future readers can tell at a glance whether a new release has actually fixed the right thing.

Also added a "Why pin instead of work around?" section to the PR body. Short version: the regression is in `CondaRequirement.__hash__`, which fires as soon as conda requirements are processed — there is no CLI flag or `pyproject.toml` shape that avoids that code path, so pinning is the only real lever and alternative workarounds were not pursued.

Follow-ups that came out of the review have been filed as separate issues / will be opened as a follow-up PR; links will be added to the PR body's "Follow-up" section as they land.

PR #895, opened 2026-05-01
Fix ITS plot when post-period has a single observation

drbenvincent:
## Summary

The `InterruptedTimeSeries` plot rendered all three panels as visually empty when there was only one post-treatment observation:

- the arviz HDI ribbon collapses to a zero-area polygon for a single x-coordinate,
- the median line has no neighbours to connect to (and no marker), and
- the (then top-of-zorder, `lw=3`, solid) treatment `axvline` landed exactly on top of the only post-period datum.

This PR fixes that with three focused changes scoped to the `InterruptedTimeSeries` plot path:

1. **Universal restyle** — the treatment `axvline` is now thin (`lw=1.5`), dashed, and at `zorder=1.5` (below data, above background). It reads as an annotation rather than data and never occludes observations or HDI ribbons. This is a small visual win for the normal multi-observation case too.
2. **Singleton overlay** — when `len(datapost) <= 1`, an explicit median dot + HDI errorbar is drawn on each of the three panels so the counterfactual and impact remain readable.
3. **Legend faithfulness** — in the singleton case the *Counterfactual* legend handle is swapped to the actual errorbar artist (instead of the degenerate line+ribbon tuple), and the *Causal impact* legend entry is dropped because its `fill_between` collapses to nothing. The legend now matches what is drawn.

## Reproducing the screenshots

The two figures below were generated with this fully self-contained snippet:

```python
import matplotlib
matplotlib.use("Agg")
import pandas as pd
import causalpy as cp

SAMPLE_KWARGS = dict(tune=500, draws=500, chains=2, random_seed=42, progressbar=False)

df = (
    cp.load_data("its")
    .assign(date=lambda x: pd.to_datetime(x["date"]))
    .set_index("date")
)

# 1. Singleton case: treat the very last observation as the only post-period datum.
res_singleton = cp.InterruptedTimeSeries(
    df,
    df.index[-1],
    formula="y ~ 1 + t + C(month)",
    model=cp.pymc_models.LinearRegression(sample_kwargs=SAMPLE_KWARGS),
)
fig, _ = res_singleton.plot()
fig.savefig("its_plot_singleton.png", dpi=120, bbox_inches="tight")

# 2. Multi-observation case (sanity-check the restyle).
res_multi = cp.InterruptedTimeSeries(
    df,
    pd.to_datetime("2017-01-01"),
    formula="y ~ 1 + t + C(month)",
    model=cp.pymc_models.LinearRegression(sample_kwargs=SAMPLE_KWARGS),
)
fig, _ = res_multi.plot()
fig.savefig("its_plot_multipoint.png", dpi=120, bbox_inches="tight")
```

### Singleton post-treatment observation

its_plot_singleton

### Normal multi-observation case (regression check)

its_plot_multipoint

## Test plan

- [x] New regression test `test_its_single_post_observation_plot` in `causalpy/tests/test_integration_pymc_examples.py` covers the singleton case end-to-end and asserts the restyle, the errorbar overlay, and the legend handle changes.
- [x] Existing ITS test suite (`test_its`, `test_its_covid`, `test_three_period_its`, `test_piecewise_its`, `test_plot_show_parameter`, ITS sklearn integration) still passes — 130+ tests green.
- [x] `prek run --all-files` passes.

## Scope

Two files changed (~168 insertions / 12 deletions):

- `causalpy/experiments/interrupted_time_series.py`
- `causalpy/tests/test_integration_pymc_examples.py`


ErikRingen (approved):
LGTM

PR #902, opened 2026-05-01
Widen numpydoc-validation checks across causalpy

### LINKED ISSUE #898: Widen numpydoc-validation checks beyond PR01/PR02 and beyond .plot methods
## Context

`numpydoc-validation` is currently configured (in `pyproject.toml`) to enforce only `PR01` and `PR02`, and only on `.plot` methods (negative-lookahead exclude). That narrow scope was intentional for #886, but the hook can do considerably more across the whole `causalpy/` tree.

```toml
[tool.numpydoc_validation]
checks = [
    "PR01",  # Parameters not documented
    "PR02",  # Unknown parameters
]
exclude = ['^(?!.*\.plot$).*$']
```

While addressing #892 it became clear that `numpydoc-validation` cannot catch the body-level rendering bug we hit there (it doesn't inspect section bodies), but it can catch a number of other defects that currently slip through.

## Proposed change

Expand the configured checks to a curated set, then drop the `.plot`-only exclude so the rules apply package-wide. Candidate rules to add (each can be landed independently to keep the noise manageable):

- `PR04` parameter has no type
- `PR07` parameter has no description
- `PR10` parameter requires a space before the colon
- `RT03` return value has no description
- `SS03` summary does not end with a period
- `GL07` sections in wrong order
- `GL08` object does not have a docstring (note overlap with `interrogate`; decide whether to keep one or both)

Other rules (`SS05` infinitive verb, `RT01` no Returns section, `EX01` / `SA01` / `ES01`) are likely too noisy to enable globally and should be considered case by case.

## Acceptance criteria

- [ ] Each enabled rule is added in a separate commit so violations can be reviewed and fixed in isolation.
- [ ] The exclude regex is removed once the package-wide pass is clean.
- [ ] Configuration in `pyproject.toml` is updated and documented.

## Out of scope

- Catching the inline-bullet-list rendering bug (handled by the local `validate-docstring-lists` pre-commit hook added under #892).
- Rewriting the `BaseExperiment.plot` carve-out (that lives under #886).

## Related

- #892: docstring inline-list rendering bug
- #886: hdi_prob default rollout / plot kwarg visibility

drbenvincent:
Closes #898.

## Summary

- Broaden the `numpydoc-validation` pre-commit hook from the original `.plot`-only scope to the whole `causalpy/` package (`tests/` and `data/` excluded), and trade the `.plot`-only exclude regex for a private-member exclude (`\._`).
- Enable the curated list of new rules requested in the issue, **one per commit** to keep the diff reviewable: `PR04`, `PR07`, `PR10`, `RT03`, `SS03`, `GL07`, `GL08`.
- Replace the standalone `interrogate` hook with `GL08` (per the agreed-upon resolution of the overlap discussed in the issue): the pre-commit hook, `[tool.interrogate]` config, extras entries, the `interrogate-badge` workflow, the `Makefile` invocation, and the README badge are all removed.
- Fix every violation that surfaces under the broadened scope. The bulk of the work is in commit 1: converting the remaining Sphinx `:param:` blocks in `causalpy/experiments/*` to numpydoc `Parameters` sections, documenting forwarded `*args`/`**kwargs`, fleshing out protocol/check/maketables/skl helper docstrings, and tidying a `See Also` entry in `utils.py` that previously broke the docstring parser. Subsequent commits include only small, rule-scoped fixes.

## Commit-by-commit breakdown

1. **Broaden numpydoc-validation scope across causalpy/** — config + all `PR01`/`PR02` violations under the new scope.
2. **Enable PR04** (parameter has no type) — config only, no violations.
3. **Enable PR07** (parameter has no description) — config only, no violations.
4. **Enable PR10** (space before colon) — config only, no violations.
5. **Enable RT03** (return value has no description) — flesh out `Check.run`'s return entry.
6. **Enable SS03** (summary ends with period) — fix module- and method-level summaries that lacked a trailing period.
7. **Enable GL07** (section ordering) — reorder `Examples`/`Notes`/`See Also` blocks in `InterruptedTimeSeries`, `PiecewiseITS`, `PyMCModel.priors_from_data`, and the variable-selection-prior classes (the latter previously used a non-standard singular `Example` heading).
8. **Enable GL08 and replace interrogate with numpydoc-validation** — drop the `interrogate` hook plumbing; add a top-level package docstring and flesh out abstract `fit`/`build_model`/`calculate_cumulative_impact` stubs; rename two nested helpers in `InversePropensityWeighting._plot_weights_propensity` to `_`-prefixed names so they fall under the configured private exclude.

## Decisions / clarifications recorded in the issue thread

- **GL08 vs `interrogate`**: replace `interrogate` with `GL08` (single source of truth for missing-docstring enforcement). All `interrogate` config, extras, the badge workflow, and the README badge are removed.
- **Single PR with 8 commits**: yes — one commit per enabled rule (plus the initial broadening + `PR01`/`PR02` cleanup), as suggested in the acceptance criteria.
- **Exclude private/dunder methods**: yes — `exclude = ['\._']` so violations on internal helpers don't drown out the public-API signal. The simulated-data helpers in `causalpy/data/` are also excluded via the pre-commit `files` regex; they intentionally retain Sphinx `:param:` style for didactic clarity.

## Test plan

- [x] `prek run --all-files` passes (every hook, including `numpydoc-validation`, `mypy`, ruff, codespell, PyMarkdown, blank-line-before-bullet-lists, environment.yml sync, etc.).
- [x] `python -c "import causalpy; print(causalpy.__doc__)"` succeeds and now prints the new package docstring.
- [x] `pytest causalpy/tests/test_misc.py -x -q` passes.
- [ ] CI to re-run the full unit-test suite and docs build.


drbenvincent:
bugbot run

PR #906, opened 2026-05-07
chore: pin pymc<6 and pytensor<3 to current major versions

### LINKED ISSUE #873: Pin pymc, pytensor, and arviz to current major versions before upcoming major releases
## Context

PyMC and PyTensor have new major releases coming soon (PyMC 6.0, PyTensor 3.0), and ArviZ has already released v1.0 with breaking changes. CausalPy's current dependency constraints are open-ended and would allow these new major versions to be installed, which could cause breakage.

## Current state

In `pyproject.toml`:
- `pymc>=5.15.1` — allows PyMC 6.0+
- `pytensor>=2.38.1` — allows PyTensor 3.0+
- No upper bound on `arviz` (if declared as a dependency or transitive)

## Proposed changes

Add upper-bound pins to keep dependencies within the current major version series:

```
pymc>=5.15.1,<6
pytensor>=2.38.1,<3
```

If `arviz` is a direct dependency, also pin it: `arviz>=0.x,<1` (or update to support 1.0 explicitly once compatibility is verified).

## Why

These pins protect CI and user installs from silently pulling in a new major version with breaking API changes. Once CausalPy is verified compatible with the new major releases, the upper bounds can be bumped.

anevolbap:
Closes #873.

Adds upper-bound pins to keep PyMC and PyTensor within their current major series. `arviz` is already pinned `<1.0`.

```
pymc>=5.15.1,<6
pytensor>=2.38.1,<3
```

This protects CI and user installs from silently pulling in PyMC 6.0 or PyTensor 3.0, which contain breaking changes. Once compatibility with the new majors is verified, the upper bounds can be bumped.

`environment.yml` was regenerated by the `pyproject2conda-yaml` pre-commit hook.

## Test plan
- [x] CI green on this branch


juanitorduz (approved):
Thanks @anevolbap

PR #908, opened 2026-05-07
ci: harden welcome.yml and silence zizmor dangerous-triggers

### LINKED ISSUE #888: Resolve zizmor `dangerous-triggers` alert on `welcome.yml`
## Background

Code scanning alert [#6](https://github.com/pymc-labs/CausalPy/security/code-scanning/6) (`zizmor/dangerous-triggers`) flags the `pull_request_target` trigger in [`.github/workflows/welcome.yml`](https://github.com/pymc-labs/CausalPy/blob/main/.github/workflows/welcome.yml#L3-L7):

```yaml
on:
  pull_request_target:
    types: [opened]
  issues:
    types: [opened]
```

`pull_request_target` runs in the **base** repo's context with access to repo secrets and a write-scoped `GITHUB_TOKEN`, but is triggered by PRs that may come from forks. The classic exploit ("pwn request") is when such a workflow checks out the PR head ref and executes anything from the untrusted PR — letting a fork PR exfiltrate secrets or push to `main`. zizmor flags the trigger categorically, regardless of how it is used.

## How risky is our specific usage?

Low, in practice:

- The workflow does **not** `actions/checkout` the PR.
- It does **not** execute any script from the PR.
- It only calls two well-known actions, both pinned by SHA: `actions/first-interaction` and `actions/github-script`.
- Permissions are already minimal at the workflow level: `issues: write`, `pull-requests: write`.
- The only user-controlled value inlined is `${{ github.actor }}`, which GitHub constrains to a tight charset and which is only inserted into a Markdown comment body (not a shell).

So this is a "trigger is categorically dangerous, but the usage here is one of the documented legitimate cases" situation — we need a write-scoped token on fork PRs in order to comment / label, and `pull_request` from a fork only gives a read-only `GITHUB_TOKEN`.

### 1. (Preferred) Inline zizmor ignore + defense-in-depth hardening

Add a `# zizmor: ignore[dangerous-triggers]` comment on the trigger with a written justification, and tighten the workflow as defense-in-depth:

- Hoist `github.actor` into `env:` and reference `\${ACTOR}` rather than inlining `${{ github.actor }}` directly into action inputs.
- Add an explicit top-level `permissions: {}` and grant only what's needed at the job level (already mostly the case).
- Keep SHA-pinned actions (already done).

**Why preferred:** Lowest effort, keeps the rationale next to the code (so the next person reading the workflow understands why the suppression exists), clears the alert, and we still pick up real hardening wins. The trigger genuinely is being used safely here, so suppression is honest rather than papering over a problem.

### 2. Dismiss the code-scanning alert as "used safely" with a written justification

Same outcome on the alert dashboard, but the rationale lives in GitHub's UI rather than in the repo.

**Why ranked below option 1:** Suppression rationale is invisible from the workflow file itself, easy to forget on the next refactor, and not portable if we ever migrate scanners.

### 3. Split the workflow into two files

One `issues`-only welcome (totally safe trigger) and one `pull_request_target` workflow for PRs.

**Why ranked below options 1–2:** Doesn't satisfy zizmor (the PR file still uses `pull_request_target`), but does narrow the surface that uses the dangerous trigger and makes the security boundary more obvious. Modest readability win at the cost of a second file to maintain.

### 4. Replace `pull_request_target` with the `workflow_run` pattern

A read-only `pull_request` workflow uploads minimal context as an artifact, then a privileged `workflow_run` workflow consumes it and posts the comment / adds the label. This is GitHub's officially recommended pattern when you genuinely need to act on fork PR data with elevated privileges.

**Why ranked below options 1–3:** zizmor-clean and architecturally "correct," but it's materially more YAML, two workflows that have to stay in sync, and harder to reason about. Overkill given that we don't even read PR contents — we just post a hardcoded greeting and add a label based on `author_association`.

### 5. Replace with a GitHub App / PAT-based bot triggered by `pull_request`

Use the read-only `pull_request` trigger, and have the App's token (or a PAT) handle the write actions.

**Why ranked below options 1–4:** Avoids `pull_request_target` entirely, but introduces ops burden (managing the App or PAT) and a larger blast radius if that secret ever leaks. Disproportionate for a welcome workflow.

### 6. Delete the workflow

If the value of the welcome message is judged to be small relative to the policy noise, drop it.

**Why ranked last:** It does provide value (first-time-contributor onboarding + labeling), and the actual security risk of the current workflow is low. Removing functionality should be a last resort, not a first response to a categorically-flagged trigger.

## Proposed action

Go with **option 1**: add an inline zizmor ignore with justification, hoist `github.actor` into `env:`, and add an explicit top-level `permissions: {}`. Reassess if the alert resurfaces or the workflow grows to do anything that touches PR contents.

---

Tagging @maresb because of your interest in security/devops :)

anevolbap:
# Resolve zizmor `dangerous-triggers` alert on `welcome.yml`

Closes #888.

## Summary

The welcome workflow uses `pull_request_target` so it can comment and label PRs from forks (the read-only `pull_request` token cannot write). zizmor flags this trigger categorically, but the usage here is one of the documented safe cases. This PR suppresses the alert with an inline justification and applies defense-in-depth hardening so the rationale lives next to the code.

## Changes

- Inline `# zizmor: ignore[dangerous-triggers]` on the trigger with a short justification.
- Top-level `permissions: {}` (deny-all), with `issues: write` and `pull-requests: write` moved to the job.
- Hoist `github.actor` into a job-level `env: ACTOR` and reference `${{ env.ACTOR }}` in the action inputs.

No behavior change.

## Testing

- `prek run --files .github/workflows/welcome.yml` passes.
- The workflow itself only runs on real `pull_request_target` / `issues` events, so end-to-end validation will happen on the next first-time contributor event.


anevolbap:
Going with option 1 from the issue: inline zizmor ignore with a written justification, top-level `permissions: {}` with the writes moved to job level, and `github.actor` hoisted into `env: ACTOR`. No behavior change. The other options stay on the table if the alert pattern resurfaces or the workflow grows to actually touch PR contents.

drbenvincent:
Thanks for opening this when #888 was filed — this is the PR we're taking forward.

Reviewed against option 1 in the issue: the inline zizmor ignore with justification, top-level `permissions: {}` with writes scoped to the job, and hoisting `github.actor` into `env.ACTOR` all look correct. No behavior change, and the rationale living in the workflow file is exactly what we wanted.

#953 was a duplicate (now closed); no further action needed on your side beyond getting this rebased onto current `main` — the branch is a bit behind. Once that's done and CI is green (especially zizmor), this should be good to merge.

drbenvincent (approved):

PR #412, opened 2024-09-16
Add pixi

maresb:
With this you can

```bash
pixi run init-jupyter-kernel
pixi shell
jupyter lab
```

---

Side note:

I'm working on a guide for migrating to pixi, and a tool to do it automatically via `conda-lock`.

The guide is here: https://github.com/conda/conda-lock/pull/708: ([preview](https://deploy-preview-708--conda-lock.netlify.app/pixi-migration/))

Note that because things are unreleased, to follow the guide you need to install conda-lock via

```
pipx install --force git+https://github.com/conda/conda-lock@main
```

----
📚 Documentation preview 📚: https://causalpy--412.org.readthedocs.build/en/412/

PR #910, opened 2026-05-07
ci: install test job env via micromamba (refs #279)

anevolbap:
Refs #279.

The CI `test` matrix still installs PyMC via pip. #304 moved local dev to a conda install but explicitly scoped out CI. Earlier attempts at the CI half stalled (#281 conda-lock, #412 pixi). #713 has since made `environment.yml` an autogenerated artifact synced with `pyproject.toml`, which makes a clean swap straightforward.

## What this PR does

Switches the `test` matrix in `.github/workflows/ci.yml` from `pip install -e .[test]` to `mamba-org/setup-micromamba` consuming the autogenerated `environment.yml`, so PyMC comes from `conda-forge` in CI.

- Pin `setup-micromamba` to v3.0.0 by SHA, matching the security style of the rest of the workflow.
- Add `defaults.run.shell: bash -el {0}` so the activated env persists across steps.
- Source the test-only deps (`pytest`, `pytest-cov`, `diff-cover`, `papermill`) from `environment.yml` by adding `-e test` to the `pyproject2conda` invocation in `.pre-commit-config.yaml` and the `check-environment-yml` job. Keeps `environment.yml` as the single source of truth and lets the conda env cover `make test-patch-cov`.
- Install the local package with `pip install --no-deps -e .` so pip cannot pull pip-built `pymc`/`pytensor` on top of the conda ones.

## Python 3.11 retained

The matrix stays `["3.11", "3.14"]`. `environment.yml` is generated with `--python-include "python>=3.11"` (in both the prek hook and the `check-environment-yml` job), and the conda env solves on 3.11 with `pymc`, `pytensor`, `numpy`, `scipy` from `conda-forge`. The earlier "would not solve" was an artifact of the `python>=3.12` pin colliding with `create-args: python=3.11`, not a real 3.11 constraint. The `requires-python = ">=3.11"` contract and the 3.11 classifier are unchanged.

## CI runtime impact

The `test` matrix went from ~3-5 min (pip) to roughly 5-8 min (conda) on the first run. `cache-environment` should bring subsequent runs closer to the pip baseline as long as `environment.yml` does not change.

## Verification

- Local run on Python 3.14: `pymc`, `pytensor`, `numpy`, `scipy` all from `conda-forge` with MKL BLAS, `blas_ldflags` references the conda env's `lib`, no `pytensor.tensor.blas` numpy-fallback warning. Doctests 35 passed / 11 skipped, extra tests 2 passed, full suite 936 passed / 17 skipped, coverage 94.97% in 3m27s.
- CI on this PR: both `test (3.11)` and `test (3.14)` pass. The `Show environment` step in each test job's log confirms `pymc` is installed from `conda-forge` and `blas_ldflags` references the conda env's `lib`.

## Out of scope

- The `release.yml`, `test_notebook.yml`, and `uml.yml` workflows also pip-install PyMC (transitively via `causalpy[docs]` etc.). They are in the spirit of #279 but a separate change.
- Migrating to conda-lock (#281) or pixi (#412).
- Touching the `prek` job or the `check-environment-yml` job beyond the `-e test` addition above.

Reviewed with Claude Opus 4.7 (adversarial review pass).


juanitorduz (approved):
thanks @anevolbap !

@drbenvincent can you take a look as this one is key ;) ?

drbenvincent:
Thanks for all these. I'm a bit behind on causal reviews - bear with me and I'll attempt to catch up

drbenvincent (changes requested):
Thanks for the PR — this is a useful direction, and I appreciate the careful notes about keeping PyMC/PyTensor on the conda-forge path.

I think there are two changes needed before merge:

1. Please keep the supported Python range aligned with the remote pytest suite. On current `main`, the core `test` job is the thing that defines the supported lower/upper bounds, and it currently spans Python 3.11 to 3.14. This PR changes that signal to 3.12 to 3.14 while `pyproject.toml` still advertises `requires-python = ">=3.11"` and the Python 3.11 classifier. Either re-add a 3.11 test leg with a conda env that solves, or make an explicit support-drop change that updates the package metadata and classifiers at the same time. My preference for this PR is to preserve the current support contract unless we deliberately decide otherwise.

2. Please make sure the generated conda dev/test env includes all dependencies needed for the documented local test workflow. Adding `-e test` brings in `pytest`, `pytest-cov`, and `papermill`, but the generated `environment.yml` still does not include `diff-cover`, even though `pyproject.toml` lists it in the `test` extra and `make test-patch-cov` depends on it. If pyproject2conda is skipping it because it is not available as a conda package, please add it explicitly in the generated env path, likely via the pip section or an explicit pyproject2conda argument, so the conda-based workflow remains complete.

The branch is also behind current `main` and there are overlapping updates in `.github/workflows/ci.yml` and `.pre-commit-config.yaml`, so please update/rebase and rerun CI after the changes.

anevolbap:
Addressed both points and synced with main.

1. Python 3.11: restored. I loosened the `pyproject2conda --python-include` from `python>=3.12` to `python>=3.11` in both the pre-commit hook and the `check-environment-yml` job, regenerated `environment.yml`, and re-added the `3.11` leg to the test matrix. The conda env solves cleanly on 3.11 (python 3.11.15, pymc 5.28.5, pytensor 2.38.3, all py311 builds from conda-forge), so the earlier drop was only an artifact of the `>=3.12` pin colliding with `create-args: python=3.11`, not a real 3.11 constraint. The `requires-python = ">=3.11"` contract is preserved, no metadata change needed.

2. diff-cover: fixed. Syncing with current `main` pulls `diff-cover` into the `test` extra (the patch-coverage gate, #951) and brings in the `test-patch-cov` Makefile target. I regenerated `environment.yml` so the `-e test` path includes it. `diff-cover` (10.3.0) is on conda-forge, so it lands as a normal conda dependency, no pip section needed, and I verified the env still solves on Python 3.11 with it included. The conda dev/test env now covers `make test-patch-cov`.

Branch is rebased on current `main` and CI is rerunning.

drbenvincent:
Thanks for the follow-up — the conda-forge test path looks solid.

Both earlier items look addressed on head: Python 3.11 is back in the matrix, `diff-cover` is in the generated `environment.yml` via `-e test`, and `pip install --no-deps -e .` is the right guard against pip-built PyMC/PyTensor.

Only blocker before merge: the branch conflicts with current `main` in `.pre-commit-config.yaml` (#957 hook bumps vs your pyproject2conda changes). Rebase, keep both sides, rerun CI, and this should be good to go.

anevolbap:
> Thanks for the follow-up — the conda-forge test path looks solid.
> 
> Both earlier items look addressed on head: Python 3.11 is back in the matrix, `diff-cover` is in the generated `environment.yml` via `-e test`, and `pip install --no-deps -e .` is the right guard against pip-built PyMC/PyTensor.
> 
> Only blocker before merge: the branch conflicts with current `main` in `.pre-commit-config.yaml` (#957 hook bumps vs your pyproject2conda changes). Rebase, keep both sides, rerun CI, and this should be good to go.

Done, rebased on current main, conflict resolved, all green. Ready for re-review to clear the earlier changes-requested.

drbenvincent (approved):
Thanks @anevolbap — all three points from my earlier review are addressed: Python 3.11 is back in the matrix, diff-cover is included in the environment, and the `.pre-commit-config.yaml` conflict is resolved. CI is green and the micromamba-based test env setup looks good. Approving — this clears my earlier changes-requested review.

PR #917, opened 2026-05-14
Add residual-noise sensitivity checks to power_analysis

### LINKED ISSUE #914: Methodology review: iid Gaussian noise injection in SyntheticControl.power_analysis()
## Summary

`SyntheticControl.power_analysis()` introduces simulation-to-simulation variability by drawing iid Gaussian noise scaled by the pre-period residual standard deviation and adding it to the treated unit's pre-period values. This is the *only* source of variability in the loop — the injected effect is deterministic, the donor pool is fixed, and the model is fixed. The choice deserves a closer methodological look: we have not found a clean reference endorsing exactly this resampling scheme, and the homoscedasticity / independence assumptions baked into iid Gaussian noise are unlikely to hold for the kinds of time-series SC is typically applied to.

This issue mirrors the concern raised in [PR #819 comment](https://github.com/pymc-labs/CausalPy/pull/819#issuecomment-4430258084) and lifts it out so it can be tracked independently of that PR.

## Current behaviour

Inside `power_analysis()`, for each simulation:

```python
residual_std = float(residuals.mean(dim=["chain", "draw"]).std(dim="obs_ind").values)
noise = rng.normal(0, residual_std, size=len(pre))
for col in self.treated_units:
    noise_data.loc[noise_data.index < self.treatment_time, col] += noise
```

A fresh SC is then fit on `noise_data` and `validate_design()` is called at the current candidate effect size. Detection rates across simulations become the y-axis of the power curve.

## Concerns

1. **Noise model is iid Gaussian.** Real pre-period residuals are typically autocorrelated and often heteroscedastic. An iid draw under-represents the variation expected in a fresh realisation of the same DGP and will tend to make the power curve look tighter / more optimistic than it should.
2. **Only the treated unit is perturbed.** The donor pool stays at its single observed path. This is a defensible choice — donors are the "design", treated is the "outcome" — but it is not standard sampling-distribution power. Abadie-style inference uses placebo-in-space permutations; the Bayesian-power literature (assurance, design priors) typically simulates from the full posterior predictive. Neither is what this code does.
3. **No citation in the docstring.** We do not currently point users at a paper that endorses this exact scheme. Either we can find one, or we should be explicit that this is a heuristic and describe its limitations.
4. **Residual-std estimation is fragile.** `residual_std = residuals.mean(["chain","draw"]).std("obs_ind")` collapses posterior uncertainty before taking a spread; if the pre-period fit is poor (see #pseudo_post_bias.md) this scalar may not reflect the true noise level the user cares about.

## Proposed work

In rough order of cost / value:

1. **Documentation.** Update the `power_analysis()` docstring and the `sc_pymc.ipynb` narrative to describe exactly what is being resampled and what assumptions that makes. Recommend cross-checking the power curve against placebo-in-space inference if the user is unsure.
2. **Alternative noise schemes.** Add an optional `noise_method` kwarg supporting at least:
   - `"iid_gaussian"` (current behaviour, default).
   - `"block_bootstrap"` — resample contiguous blocks of pre-period residuals, preserving short-range autocorrelation.
   - `"ar1"` — fit an AR(1) to the pre-period residuals and simulate from it.
3. **Posterior-predictive resampling.** As a separate `noise_method="posterior_predictive"`, draw fresh realisations of the treated pre-period from the design-phase fit's posterior predictive. This is the Bayesian-power-analysis canonical move.
4. **Cite literature.** If we can find a clean reference for any of (1)–(3), add it to the docstring and `references.bib`. Candidate starting points: SC simulation work in `{cite:t}`abadie2021using`` and review articles on Bayesian power analysis (Kruschke-style assurance).

## Acceptance criteria

- [ ] `power_analysis()` docstring describes the iid Gaussian resampling scheme and its assumptions explicitly.
- [ ] If no citation is found, the docstring flags the scheme as a heuristic and points at the alternatives below.
- [ ] At least one alternative noise method (block bootstrap or AR(1)) is implemented as an opt-in `noise_method` kwarg.
- [ ] Unit tests cover at least the default and one alternative, including a sanity check that the alternative produces visibly autocorrelated noise on a synthetic example.
- [ ] Notebook (`sc_pymc.ipynb`) optionally adds a one-cell comparison of the default and an alternative on Prop 99, framed as a sensitivity check on the power curve itself.

## Out of scope

- Replacing the default. Backwards compatibility argues for keeping iid Gaussian as the default and exposing alternatives via the kwarg.
- Fixing the underlying pre-period mis-fit bias (tracked separately in `pseudo_post_bias.md`).

bhemaraju138-pixel:
## Summary

Closes #914.

This PR keeps the existing iid Gaussian residual-noise simulation in `SyntheticControl.power_analysis()` as the default, and adds two opt-in residual simulation methods for sensitivity analysis:

- `noise_method="block_bootstrap"`
- `noise_method="ar1"`

It also documents the assumptions behind the default iid Gaussian scheme, clarifies that the resulting power curve is heuristic, and adds tests for the default and alternative methods.

## Motivation

The existing implementation perturbs only the treated unit’s pre-period values with iid Gaussian noise scaled by a scalar residual standard deviation. That is simple and backward-compatible, but it assumes independent and homoscedastic residual variation. For synthetic-control time series, residuals may be autocorrelated, seasonal, heteroscedastic, or affected by structural misfit.

This PR does not replace the default. Instead, it makes the assumptions explicit and gives users opt-in sensitivity checks for whether the power curve changes under residual-noise models that preserve short-range dependence.

## Design

- `iid_gaussian` preserves the current behavior.
- `block_bootstrap` resamples centered pre-period residuals in overlapping contiguous blocks. When `block_length=None`, it uses `ceil(sqrt(n_pre))`; users can override it.
- `ar1` centers the residual path, estimates `phi` by least squares on lagged residuals, clips it to the stationary range, estimates the innovation scale, and simulates a zero-mean AR(1) residual path.
- Very short pre-periods are handled explicitly: AR(1) requires at least three residual observations, while block bootstrap validates `block_length`.
- `random_seed` now makes both residual noise and generated refit seeds reproducible unless the user explicitly supplies `sample_kwargs["random_seed"]`.

These methods are intentionally framed as sensitivity checks, not as canonical SC sampling-distribution power analysis. Users should still compare against placebo-in-space and placebo-in-time diagnostics.

## Tests

Added coverage for:

- default iid Gaussian behavior
- block-bootstrap and AR(1) power-analysis calls
- invalid `noise_method`
- invalid `block_length`
- AR(1) short-pre-period validation
- reproducible helper draws under fixed seeds
- autocorrelation sanity checks on synthetic AR(1) residuals

## Documentation

Updated the `power_analysis()` docstring and the Prop 99 notebook narrative to describe the residual resampling assumptions and show a sensitivity comparison across noise methods.

## References

The iid Gaussian treated-only perturbation is documented as a heuristic because I did not find a clean SC reference endorsing that exact scheme. The block-bootstrap alternative is motivated by Künsch (1989), while Abadie (2021) is cited for general synthetic-control feasibility and placebo-diagnostic framing.

## Validation

- `.venv/bin/python -m pytest causalpy/tests/test_sc_design.py -k "PowerAnalysis or noise" --no-cov`
- `.venv/bin/prek run --all-files`

### DIFF
--- a/causalpy/experiments/sc_results.py
+++ b/causalpy/experiments/sc_results.py
@@ -171,12 +171,18 @@ class PowerCurveResult:
         The detection criterion used.
     raw_results : list[list[DressRehearsalResult]]
         Nested list: per effect size, per simulation.
+    noise_method : str
+        Residual-noise simulation method used.
+    block_length : int or None
+        Block length used for block-bootstrap residual noise, if applicable.
     """
 
     effect_sizes: list[float]
     detection_rates: list[float]
     criterion: str
     raw_results: list[list[DressRehearsalResult]] = field(repr=False)
+    noise_method: str = "iid_gaussian"
+    block_length: int | None = None
 
     def plot(self) -> tuple[plt.Figure, plt.Axes]:
         """Power curve: effect size vs detection rate."""
@@ -214,6 +220,8 @@ class PowerCurveResult:
                     "mean_recovery": np.mean(means),
                     "median_recovery": np.median(means),
                     "n_simulations": len(sim_results),
+                    "noise_method": self.noise_method,
+                    "block_length": self.block_length,
                 }
             )
         return pd.DataFrame(rows)
--- a/causalpy/experiments/synthetic_control.py
+++ b/causalpy/experiments/synthetic_control.py
@@ -1155,11 +1155,40 @@ class SyntheticControl(BaseExperiment):
         holdout_periods: int | None = None,
         sample_kwargs: dict | None = None,
         random_seed: int | None = None,
+        noise_method: Literal[
+            "iid_gaussian", "block_bootstrap", "ar1"
+        ] = "iid_gaussian",
+        block_length: int | None = None,
     ) -> PowerCurveResult:
         """Simulation-based Bayesian power curve.
 
         For each candidate effect size, runs :meth:`validate_design`
-        multiple times (with added noise) and records detection rates.
+        multiple times (with added residual noise) and records detection
+        rates. The simulation perturbs only the treated unit's pre-period
+        outcomes; donor paths, the model class, the candidate effect size,
+        and the design are fixed across simulations.
+
+        The default ``noise_method="iid_gaussian"`` preserves the original
+        implementation: it adds iid Gaussian noise with standard deviation
+        equal to the posterior-mean pre-period residual standard deviation.
+        This is a heuristic sensitivity calculation. It assumes independent,
+        homoscedastic Gaussian residual variation, and is not a canonical
+        synthetic-control sampling-distribution power analysis. No clean
+        synthetic-control reference is known for this exact treated-only
+        resampling scheme.
+
+        ``"block_bootstrap"`` and ``"ar1"`` are opt-in sensitivity checks
+        for short-range residual dependence. They are not universally better
+        replacements for iid Gaussian noise. All methods rely on
+        posterior-mean residuals and do not address donor-pool uncertainty,
+        heteroscedasticity, seasonality, posterior-predictive design priors,
+        or pseudo-post misfit bias. Users should compare noise methods and
+        cross-check the resulting curve with placebo-in-space and
+        placebo-in-time diagnostics. The block bootstrap samples overlapping
+        moving blocks of centered residuals. The AR(1) method centers the
+        residuals, estimates the lag-1 coefficient by least squares with no
+        intercept, clips it to ``[-0.99, 0.99]``, and simulates innovations
+        from the fitted residual innovation scale.
 
         Parameters
         ----------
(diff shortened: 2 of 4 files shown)


drbenvincent:
## Automated triage

**Why:**
- New public API: `noise_method` parameter (`block_bootstrap` and `ar1`) on `SyntheticControl.power_analysis()` — affects numerical behaviour of power analysis.
- Changes to `causalpy/experiments/synthetic_control.py` (225 lines added) and `sc_results.py`.
- Tests added; docs updated with a new notebook comparison.
- CI status is stale (last ran May 2026, no recent test results).

**Review focus:**
1. Verify the AR(1) estimation is robust (short pre-period, near-unit-root, non-stationary data).
2. Confirm the block-bootstrap default block length `ceil(sqrt(n_pre))` is reasonable.
3. Check that the iid Gaussian default remains unchanged and backward-compatible.

**Confidence:** high

PR #942, opened 2026-06-05
Fix #664: clone sklearn models to avoid mutating fit_intercept

### LINKED ISSUE #664: DifferenceInDifferences mutates fit_intercept on user models

aahammer-eu:
## Problem

`DifferenceInDifferences`, `PiecewiseITS`, `StaggeredDifferenceInDifferences`, and `PanelRegression` all silently mutate the user's sklearn model by setting `fit_intercept=False`. This is a hidden side effect: callers pass in a configured estimator, but the experiment overrides a key hyperparameter without consent. This makes it impossible to safely reuse the same model instance across experiments.

## Solution

Centralise the `fit_intercept` handling in `BaseExperiment.__init__`:

1. **Clone the model** before wrapping it, using `sklearn.base.clone()` (with `copy.deepcopy` fallback for custom regressors that lack `get_params()`). The original user instance is never mutated.
2. **Warn** via `UserWarning` when overriding `fit_intercept=True` on the clone, explaining why CausalPy requires `fit_intercept=False` (patsy already includes the intercept in the design matrix).
3. **Remove** the duplicated `fit_intercept` mutation from all 4 experiment classes.

## Why this approach

- **Clone, don't raise**: raising an error would break existing code that passes `fit_intercept=True` (sklearn's default). Cloning + warning is non-breaking but transparent.
- **`sklearn.clone` preferred over `deepcopy`**: `clone()` is the idiomatic sklearn way to copy estimators — it's lightweight and creates a fresh instance from hyperparameters. `deepcopy` is the fallback for non-standard regressors (e.g. custom `RegressorMixin` subclasses without `get_params()`).
- **Centralised**: one fix in `BaseExperiment` instead of patching 4 experiment classes independently.

## Tests

All existing OLS/sklearn tests pass (38 passed, 0 failed). The existing tests for models without `fit_intercept` (`KNeighborsRegressor`, `MinimalRegressor`) continue to work via the `deepcopy` fallback.

Closes #664

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/causalpy/experiments/base.py
+++ b/causalpy/experiments/base.py
@@ -18,6 +18,8 @@ Base class for quasi experimental designs.
 from __future__ import annotations
 
 import contextlib
+import copy
+import warnings
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Literal
@@ -25,7 +27,7 @@ from typing import Any, Literal
 import arviz as az
 import matplotlib.pyplot as plt
 import pandas as pd
-from sklearn.base import RegressorMixin
+from sklearn.base import RegressorMixin, clone
 
 from causalpy.maketables_adapters import get_maketables_adapter
 from causalpy.pymc_models import PyMCModel
@@ -118,7 +120,27 @@ class BaseExperiment(ABC):
         # Ensure we've made any provided Scikit Learn model (as identified as being type
         # RegressorMixin) compatible with CausalPy by appending our custom methods.
         if isinstance(model, RegressorMixin):
+            # Clone to avoid mutating the caller's estimator instance (#664).
+            try:
+                model = clone(model)
+            except TypeError:
+                # Models without get_params() can't be sklearn-cloned;
+                # fall back to a deep copy.
+                model = copy.deepcopy(model)
             model = create_causalpy_compatible_class(model)
+            # Patsy includes the intercept as a design-matrix column, so
+            # sklearn models must not add their own intercept.
+            if getattr(model, "fit_intercept", False):
+                warnings.warn(
+                    f"{type(model).__name__} had fit_intercept=True, but CausalPy "
+                    "requires fit_intercept=False because the intercept is already "
+                    "included in the design matrix by patsy. A cloned copy of the "
+                    "model with fit_intercept=False will be used; the original "
+                    "instance is unchanged.",
+                    UserWarning,
+                    stacklevel=2,
+                )
+                model.fit_intercept = False
 
         if model is None and self._default_model_class is not None:
             model = self._default_model_class()
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -160,11 +160,6 @@ class DifferenceInDifferences(BaseExperiment):
             }
             self.model.fit(X=self.X, y=self.y, coords=COORDS)
         elif isinstance(self.model, RegressorMixin):
-            # Ensure the intercept is part of the coefficients array rather than
-            # a separate intercept_ attribute.  See #664 / PR #693 for
-            # centralising this in BaseExperiment.
-            if hasattr(self.model, "fit_intercept"):
-                self.model.fit_intercept = False
             self.model.fit(X=self.X, y=self.y)
         else:
             raise ValueError("Model type not recognized")
--- a/causalpy/experiments/panel_regression.py
+++ b/causalpy/experiments/panel_regression.py
@@ -305,11 +305,6 @@ class PanelRegression(BaseExperiment):
             }
             self.model.fit(X=self.X, y=self.y, coords=COORDS)  # type: ignore[arg-type]
         elif isinstance(self.model, RegressorMixin):
-            # For scikit-learn models, set fit_intercept=False so that the
-            # patsy intercept column is included in the coefficients array.
-            # TODO: later, this should be handled in ScikitLearnAdaptor itself
-            if hasattr(self.model, "fit_intercept"):
-                self.model.fit_intercept = False
             self.model.fit(X=self.X, y=self.y)
 
     def _demean_transform(self, data: pd.DataFrame, group_var: str) -> pd.DataFrame:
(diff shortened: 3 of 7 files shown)


aahammer-eu:
https://www.linkedin.com/in/andreas-a-431a9413a

drbenvincent (changes requested):
Welcome, @aahammer-eu — thanks for your first contribution to CausalPy! This is a thoughtful fix for a real footgun (#664), and the PR is easy to review. A few things landed really well:

**What worked well**

- **Right place to fix it** — Centralizing clone + `fit_intercept` handling in `BaseExperiment.__init__` is the clean approach. Every experiment that calls `super().__init__(model=...)` gets the fix automatically, including `PiecewiseITS`, which fits in `__init__` after `super()`.
- **Complete cleanup** — You removed the duplicated `fit_intercept` mutation from all four experiment classes (`DifferenceInDifferences`, `PanelRegression`, `PiecewiseITS`, `StaggeredDifferenceInDifferences`), not just DiD.
- **Sensible non-breaking strategy** — Clone + `UserWarning` is a good balance: callers who pass `fit_intercept=True` (sklearn's default) keep working, but the original estimator is no longer mutated.
- **Robust fallback** — `sklearn.base.clone()` with `copy.deepcopy` for custom regressors without `get_params()` matches how the codebase already handles edge cases in tests.
- **Clear warning message** — The explanation about patsy already including an intercept in the design matrix will help users understand *why* CausalPy overrides this.

**Requested changes before merge**

1. **Add regression tests for #664** — Existing tests pass, but nothing new asserts the behavior this PR is meant to fix. Could you add at least one test (ideally covering one or two experiment types) that:
   - passes a sklearn model with `fit_intercept=True`
   - checks the **original** instance is unchanged (`assert model.fit_intercept is True`)
   - checks the experiment's internal model has `fit_intercept=False`
   - uses `pytest.warns(UserWarning, match="fit_intercept")` where appropriate

   [PR #693](https://github.com/pymc-labs/CausalPy/pull/693) has a similar test pattern in `test_integration_skl_examples.py` if that's a useful reference — your PR is broader in scope, so similar coverage here would be great.

2. **Consider warning verbosity** — `LinearRegression()` defaults to `fit_intercept=True`, so most sklearn experiment usage will emit this warning on every construction. The fix is correct; we may want `warnings.filterwarnings("once", ...)` for this message, or a maintainer decision on whether silent correction is better for the default case. Happy to discuss — not necessarily a blocker if you have a preference.

**Optional nit**

- After cloning, `cloned.set_params(fit_intercept=False)` is slightly more idiomatic than `model.fit_intercept = False`, but either works for standard sklearn estimators.

Overall this is on the right track — thanks again for picking up #664 and for the clear PR description. Looking forward to the test additions!

drbenvincent:
> https://www.linkedin.com/in/andreas-a-431a9413a

Link doesn't work? Or is it my wifi on the train?

aahammer-eu:
Thanks for the review! Pushed updates addressing all three points:

**1. Regression test added** — `test_sklearn_clone_fit_intercept.py` with one test asserting: original model unchanged, internal clone has `fit_intercept=False`, and `UserWarning` emitted. Also adjusted `test_staggered_did_all_identified_with_never_treated` to pass `fit_intercept=False` explicitly, since its `simplefilter("error")` block was catching the new (unrelated) warning.

**2. Warning verbosity** — Kept the warning with no module-level `filterwarnings`. Python's default `"default"` action already suppresses repeats from the same call location ([docs](https://docs.python.org/3/library/warnings.html#the-warnings-filter)), so the user sees it at most once per session. Libraries should not mutate global warning filters at import time — that's the caller's prerogative.

**3. `set_params()` nit** — Kept direct attribute assignment because `set_params()` relies on `get_params()`, which custom regressors using the `copy.deepcopy` fallback may not have. Direct assignment works for both paths.

Full suite passes: 1002 passed, 5 skipped, 97% coverage.

aahammer-eu:
@drbenvincent It's my link. That's what happens if you do things without AI...
https://www.linkedin.com/in/andreas-adlichhammer-431a9413a/
Contributing to open source is new to me. I'm still learning how to live in this new AI-enhanced world.

drbenvincent (approved):
Thanks for addressing the review feedback — the regression test in `test_sklearn_clone_fit_intercept.py` covers the core #664 contract (original unchanged, internal clone corrected, warning emitted).

Approving merge once CI is green. I closed #693 in favour of this PR: the `BaseExperiment.__init__` centralisation fixes all four experiment classes, including `PiecewiseITS` which fits in `__init__` and wouldn't be covered by the deferred fit-time approach in #693.

Optional follow-up (not blocking): `model.set_params(fit_intercept=False)` is slightly more idiomatic than direct attribute assignment for sklearn estimators. Happy to merge either way.

PR #943, opened 2026-06-06
ci(docs-preview): cancel when not in defined scopes of git diff

### LINKED ISSUE #941: Skip remote readthedocs build when unnecessary
So this is just an idea, I'm not sure if it's possible... But it would be nice if we only trigger remote read the docs builds when something has actually changed about the docs.

Because we rely on doc notebooks being executed locally and not remote docs building takes place, then presumably we can define a set of conditions where we do or do not need to trigger remote docs build.

This could save a lot of remote compute and time spent waiting for remote checks to complete

RussellSB:
### Description

In read the docs there is a special exit code 183 that can be triggered to cancel the build docs job.

This is the approach I took for disabling PR builds when unnecessary. I also linked read the docs to my forked repo to test 3 scenarios for expected behaviour.

**Main Change**: 
Skips PR preview builds when no docs-relevant files changed.  Closes #941 
Only builds on PR when changes detected under:
- `docs/`
- `.readthedocs.yaml`
- `causalpy` (exluding sub-folders: `tests/` , `data/`, and `skills/`

### Question

Are you happy with the files/folders specified above @drbenvincent ? 

Let me know if their are any inclusions/exclusions you also want to be made explicit. 
Also note that this only applies to PR preview builds. When merged to main, docs build will trigger regardless.

### Testing

Tested the expected behaviour on my fork repo RussellSB/CausalPy, which I linked to ReadTheDocs:

| # | Scenario | Expected Behavior | Observed Result | Evidence PR(s) | Status |
|---|---|---|---|---|---|
| 1 | Merge to `main` after implementing RTD skip logic and follow-up fixes | Docs build on `main` should still run normally and complete successfully | Docs build on `main` completed successfully after merge | [#1](https://github.com/RussellSB/CausalPy/pull/1), [#4](https://github.com/RussellSB/CausalPy/pull/4), [#5](https://github.com/RussellSB/CausalPy/pull/5) | Pass |
| 2 | PR with no docs-relevant changes | RTD preview build should be cancelled/skipped | Preview was skipped as intended | [#2](https://github.com/RussellSB/CausalPy/pull/2) | Pass |
| 3 | PR with docs-relevant changes | RTD preview build should run | Preview build ran as expected | [#3](https://github.com/RussellSB/CausalPy/pull/3) | Pass |

### Screenshots

Inspecting from the ReadTheDocs builds page `1`, `2`, `3` follows successful expected behaviour described above:

image

And clicking on `2`:

image

----

### Resources

Implemented following read the docs guidelines
- https://docs.readthedocs.com/platform/stable/guides/build/skip-build.html
- https://docs.readthedocs.com/platform/stable/build-customization.html
- https://docs.readthedocs.com/platform/stable/config-file/v2.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -14,6 +14,22 @@ build:
     # nodejs: "16"
     # rust: "1.55"
     # golang: "1.17"
+  jobs:
+    post_checkout:
+      # Skip PR preview builds when no docs-relevant files changed.
+      # exit 183 cancels the build (shown as green on the PR, not a failure).
+      # Pushes to main are unaffected — live docs always rebuild.
+      # Ref: https://docs.readthedocs.com/platform/stable/guides/build/skip-build.html
+      - git fetch origin main --depth=1
+      - |
+        if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && \
+           git diff --quiet origin/main -- docs/ .readthedocs.yaml causalpy/ \
+             ':(exclude)causalpy/tests/' \
+             ':(exclude)causalpy/data/' \
+             ':(exclude)causalpy/skills/';
+        then
+          exit 183;
+        fi
 
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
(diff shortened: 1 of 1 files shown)


drbenvincent (approved):
Excellent! Let's see how it goes. Approving...

PR #949, opened 2026-06-08
Warn when staggered DiD ATTs lack untreated-period support

### LINKED ISSUE #938: Warn on non-identified ATT in StaggeredDiD when periods lack untreated support (no never-treated units)
## Summary

`StaggeredDifferenceInDifferences` can silently produce biased, non-identified treatment effects when a calendar period has no untreated observations (i.e. every unit is already treated in that period). The most common trigger is a dataset with **no never-treated units**, which makes the **last-treated cohort's** post-treatment ATTs non-identified. The estimator currently runs without any warning in this case. We should detect this condition and warn (or error) clearly.

## Background

The imputation estimator (Borusyak, Jaravel & Spiess, 2024) fits a two-way fixed-effects model of the untreated potential outcome, `Y_it(0) = alpha_i + gamma_t`, on untreated observations only, then imputes counterfactuals for treated cells. Imputing a treated cell `(i, t)` requires both:

- `alpha_i` — unit `i`'s fixed effect, identified from its own pre-treatment periods.
- `gamma_t` — the time effect for calendar period `t`, identified from any units that are untreated at `t`.

If a calendar period `t` has **zero untreated observations** (everyone already treated), `gamma_t` is not identified, so the counterfactual and ATT for every cell in that period are not identified.

For a cohort `g` to have its post-treatment effect at time `t >= g` identified, there must exist a unit still untreated at `t` (treatment time `> t`). Early cohorts are fine because later-adopting units provide untreated support. The **last-treated cohort** `G_max` needs units with treatment time `> G_max`, which can only be **never-treated** units. Hence, without never-treated units, the last cohort's post-treatment ATTs are not identified. This is a well-known limitation of imputation / event-study designs: the final period (or last cohort) is not identified without never-treated units or additional restrictions.

## Current behaviour

The only guard is a global check that *some* untreated observation exists:

```python
n_untreated = self.data["_is_untreated"].sum()
if n_untreated == 0:
    raise DataException(...)
```

There is no per-period support check. When a calendar period is fully treated, its `C(time)` column is all-zeros in the training design matrix:

- OLS (`lstsq`) silently returns ~0 for that coefficient.
- The PyMC `LinearRegression` falls back to the prior for that coefficient.

Either way the user gets a biased ATT (typically for the last cohort / final periods) with **no warning**.

## Proposed changes

1. After identifying untreated observations, check for calendar periods (and, by extension, group-time cells) that have no untreated support.
2. Emit a clear warning (or raise, TBD) identifying which periods/cohorts are not identified — e.g. "No untreated observations in period(s) {...}; treatment effects for cohort(s) {...} are not identified. Provide never-treated units or restrict the event window."
3. Consider excluding or flagging non-identified `ATT(g, t)` / `ATT(e)` cells in the output tables rather than reporting misleading point estimates.
4. Document the never-treated-unit requirement and this limitation in the docstring and the example notebook.

## Acceptance criteria

- A dataset with no never-treated units (last cohort fully treated through the final period) triggers a clear, informative warning/error rather than silently returning biased estimates.
- Non-identified cells are clearly distinguished in outputs.
- Tests cover the fully-treated-period detection for both PyMC and OLS paths.
- Docstring and notebook explain the identification requirement.

drbenvincent:
## Summary

Detect calendar periods without untreated support in `StaggeredDifferenceInDifferences`, warn users when post-treatment ATTs are not identified (e.g. no never-treated units), and mark affected `ATT(g, t)` / `ATT(e)` cells with `identified=False` and `NaN` estimates instead of silently returning biased values.

Fixes #938

## Root cause

The imputation estimator fits time fixed effects `γ_t` using untreated observations only. When every unit is already treated in a calendar period, `γ_t` is not identified and counterfactual imputation for that period is invalid. The only existing guard checked that *some* untreated observation exists globally, not per-period support.

## Changes

- `causalpy/experiments/staggered_did.py`: Add per-period identification checks, `UserWarning`, `non_identified_periods_` / `non_identified_cohorts_` attributes, and `identified` column on ATT output tables with masked estimates.
- `causalpy/reporting.py`: Exclude non-identified cells from staggered DiD effect-summary prose averages.
- `causalpy/tests/test_staggered_did.py`: Parametrized tests for OLS and PyMC paths (warning + masking; no warning when never-treated units present).
- `docs/source/notebooks/staggered_did_pymc.ipynb`: Document untreated-support assumption and warning behaviour.

## Testing

- [x] `pytest causalpy/tests/test_staggered_did.py --no-cov` (64 passed)
- [x] `prek run` on changed files (all hooks passed)

## Checklist

- [x] Warns on datasets without never-treated units when last cohort post-treatment ATTs are non-identified
- [x] Non-identified cells distinguished in `att_group_time_` and `att_event_time_`
- [x] Tests cover OLS and PyMC paths
- [x] Docstring and notebook updated

### DIFF
--- a/causalpy/experiments/staggered_did.py
+++ b/causalpy/experiments/staggered_did.py
@@ -19,6 +19,7 @@ the approach of Borusyak, Jaravel, and Spiess (2024). It handles settings where
 different units receive treatment at different times.
 """
 
+import warnings
 from typing import Any, Literal
 
 import numpy as np
@@ -58,6 +59,13 @@ class StaggeredDifferenceInDifferences(BaseExperiment):
        units would have followed parallel outcome trajectories.
     3. **No anticipation**: Units do not change their behavior in anticipation
        of future treatment.
+    4. **Untreated support at each calendar period**: The time fixed effect
+       :math:`\\gamma_t` for calendar period :math:`t` is identified only if at
+       least one unit is untreated in that period. Without never-treated units,
+       post-treatment effects for the last-treated cohort (and any calendar
+       periods where every unit is already treated) are not identified. CausalPy
+       warns when this condition fails and marks the affected ``ATT(g, t)`` and
+       ``ATT(e)`` cells as non-identified in the output tables.
 
     Parameters
     ----------
@@ -94,8 +102,14 @@ class StaggeredDifferenceInDifferences(BaseExperiment):
         and tau_hat (treatment effect) columns.
     att_group_time_ : pd.DataFrame
         Group-time ATT estimates: ATT(g, t) for each cohort g and calendar time t.
+        Includes an ``identified`` column; non-identified cells have ``NaN`` estimates.
     att_event_time_ : pd.DataFrame
         Event-time ATT estimates: ATT(e) for each event-time e = t - G.
+        Includes an ``identified`` column; non-identified cells have ``NaN`` estimates.
+    non_identified_periods_ : set
+        Calendar periods with no untreated observations.
+    non_identified_cohorts_ : set
+        Treatment cohorts with at least one non-identified post-treatment ATT(g, t).
 
     Notes
     -----
@@ -184,6 +198,9 @@ class StaggeredDifferenceInDifferences(BaseExperiment):
 
         # Step 3: Identify untreated observations (training set)
         self._identify_untreated_observations()
+
+        # Step 3b: Check calendar-period identification support
+        self._check_att_identification()
 
         # Step 4: Build design matrices
         self._build_design_matrices()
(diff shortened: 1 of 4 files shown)


drbenvincent:
Requesting changes before merge:

`get_plot_data_bayesian(hdi_prob=...)` can still return non-identified event-time estimates when the requested HDI probability differs from the stored aggregation value. The default/stored path returns `self.att_event_time_`, which now has `identified=False` and `NaN` values for non-identified cells, but the recompute path builds a fresh `pd.DataFrame(att_et_rows)` and returns it directly without passing through `_mark_non_identified_att_rows()`.

That means a user asking for, say, `get_plot_data_bayesian(hdi_prob=0.8)` can bypass the new masking contract and receive numeric estimates for cells this PR otherwise marks as not identified. Please apply the same identification marking/masking before returning the recomputed DataFrame, and add a regression assertion for the non-default HDI path.

drbenvincent:
Good catch — the non-default `hdi_prob` recompute path in `get_plot_data_bayesian()` was returning a fresh `DataFrame` without passing through `_mark_non_identified_att_rows()`.

**Fix:** the recompute path now returns `self._mark_non_identified_att_rows(pd.DataFrame(att_et_rows))`, matching the stored/default path.

**Test:** added `test_staggered_did_get_plot_data_bayesian_masks_non_identified_on_recompute`, which fits a no-never-treated panel and calls `get_plot_data_bayesian(hdi_prob=0.80)` to assert post-treatment non-identified rows have `identified=False` and `NaN` `att` / interval columns.

Pushed in 89fa3dd2.

PR #972, opened 2026-06-22
LLM code smells refactoring

Raposones:
For easier code understaing, i refactored the following code smells:

- too-many-statements
- too-many-instance-attributes
- too-many-branches

All code smells were found by Pylint and refactored by LLM model DeepSeek V4 Flash Free via OpenCode desktop application.


drbenvincent:
Hi @Raposones, thanks for diving in. The smells you're pointing at are genuinely there.

I'm going to close this one, and I want to explain why so it's useful rather than discouraging:

It overlaps an in-flight refactor. #966 is an open PR that restructures these same experiment classes, including the too-many-instance-attributes god-object problem this PR targets. Two large refactors over the same files can't both land. Our contributing guide asks contributors to check open PRs for overlap before starting — that's the step that would've caught this.

Scope + method. A 5,500-line LLM-generated rewrite touching 21 core files is more than we can responsibly review or verify line-by-line. Our guide is explicit that agent-generated code is welcome, but not code the author hasn't personally vetted. Re-writing tests is a big no-no on a massive refactor - the tests ensure unchanged behaviour, so that should be separate.

No agreed approach. For changes this structural we settle the design in an issue first, so effort isn't spent on an approach that conflicts with the direction the maintainers are already taking.

None of this means don't contribute - we'd genuinely welcome you. The best on-ramp: pick an existing issue, maybe one tagged as 'good first issue'. If you want, comment on an existing issue or open one describing the cleanup you'd like to tackle and we'll point you at something that won't collide with current work. Thanks again for the interest, hope to see something in the future.

PR #985, opened 2026-07-08
Refactor plot_xY to plot_ribbon

### LINKED ISSUE #983: Rename `plot_xY` to `plot_ribbon`
## Summary

Rename `plot_xY` (in `causalpy/plot_utils.py`) to `plot_ribbon`.

`plot_xY` is un-Pythonic (mixed-case name in a snake_case codebase) and the capitalization doesn't communicate anything meaningful. `plot_ribbon` describes what the function actually draws (a mean/median line plus a shaded uncertainty ribbon), matches naming conventions used elsewhere in the Bayesian visualization ecosystem (e.g. `tidybayes::geom_lineribbon()`, `bayesplot::mcmc_ribbon()`), and — unlike an HDI-specific name — doesn't bake in the current implementation detail that the ribbon happens to be a highest-density interval. That leaves room to support other interval types (e.g. equal-tailed/quantile intervals) under the same name later, without a second rename.

This is a pure rename with no behavior change — no new logic, no signature change.

## Scope

`plot_xY` is internal (not re-exported from `causalpy/__init__.py`, not referenced in `docs/source/api` or any notebook), so this is not a public API break. Rename the identifier everywhere it appears:

- `causalpy/plot_utils.py` — the function definition
- Call sites and imports in `causalpy/experiments/`: `interrupted_time_series.py`, `synthetic_control.py`, `diff_in_diff.py`, `regression_discontinuity.py`, `regression_kink.py`, `prepostnegd.py`, `piecewise_its.py`, `synthetic_difference_in_differences.py`
- Docstring/comment cross-references, e.g. `:func:`~causalpy.plot_utils.plot_xY`` in `causalpy/experiments/base.py` and comments in `interrupted_time_series.py`
- `causalpy/tests/test_hdi_prob_wiring.py` — this file references the function as **string literals** passed to `mock.patch(...)` (e.g. `"causalpy.experiments.interrupted_time_series.plot_xY"`). These must be updated too, or the patches will silently no-op since the string just won't resolve to anything meaningful post-rename. Also update docstrings/comments in this file that mention `plot_xY`.

A plain project-wide find-and-replace of `plot_xY` → `plot_ribbon` should catch all of the above (identifiers, imports, docstrings, and the mock-patch string literals), but please grep for `plot_xY` afterwards to confirm zero remaining hits before opening a PR.

## Out of scope

Generalizing the function to support interval types other than HDI (e.g. renaming `hdi_prob`/`plot_hdi_kwargs` params or adding a `kind="hdi"|"eti"` option) is a separate, larger follow-up and is **not** part of this issue. HOWEVER, pay attention to what changes were made in #680 (soon to be merged, if not already done so).

## Acceptance criteria

- [ ] `plot_xY` no longer appears anywhere in the codebase (`grep -rn plot_xY causalpy` returns nothing)
- [ ] `plot_ribbon` is defined in `causalpy/plot_utils.py` with the same signature and behavior as the old `plot_xY`
- [ ] All experiment modules and tests updated to use the new name
- [ ] `prek run --all-files` passes
- [ ] `python -m pytest causalpy/tests/test_hdi_prob_wiring.py` passes

FarbodDaneshjoo:
Hi

I have prepared a small Bash script to automate this refactoring across all .py files. And also I added one line in ".pre-commit-config" file. You can easily run it by following these steps in your terminal:

1. Save the code below as rename_functions.sh in your project root.
2. Run the script using the command:    bash Terminal.sh

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
           - --author=The PyMC Labs Developers
           - --exclude=docs/
           - --exclude=scripts/
+          - --exclude=Terminal.sh
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v6.0.0
     hooks:
--- a/Terminal.sh
+++ b/Terminal.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+grep -rn "plot_xY" .
+find . -type f -name "*.py" ! -name "Terminal.sh" -exec sed -i '' 's/plot_xY/plot_ribbon/g' {} +
+grep -rn "plot_xY" .


drbenvincent:
Thanks for this. Though we don't need to commit the script because it's only for one-off use - the PR should run the script so that the PR is addressed

drbenvincent:
Thanks for taking a look at this. We're not going to merge this PR — it adds a one-off script rather than the rename in the codebase, and we've decided not to pursue the `plot_xY` → `plot_ribbon` rename (see discussion on #983). Appreciate the interest in contributing.

PR #993, opened 2026-07-11
Require a valid group*post_treatment interaction term in DiD formulas

Mari1988:
## Summary

`DifferenceInDifferences` previously accepted formulas with no interaction term at all, or with an interaction term between unrelated variables, and only failed later when `causal_impact` was accessed:

- Bayesian models: `AttributeError` (`causal_impact` was never assigned in that code path)
- OLS models: silently returned `causal_impact = None`, with no exception or warning

This PR adds validation in `_validate_formula_interaction_terms` requiring exactly one interaction term that involves both the group and post-treatment variables, raising a clear `FormulaException` at construction time instead.

## Changes

- `causalpy/experiments/diff_in_diff.py`: extend `_validate_formula_interaction_terms` to reject a missing interaction term, or one not between the group/post-treatment variables.
- `causalpy/tests/test_input_validation.py`: add regression tests for both new rejection cases, plus a test confirming a reversed-order interaction term (`post_treatment*group`) still constructs successfully (this fix is scoped to the missing-validation bug; a separate, order-sensitivity bug in the OLS causal_impact lookup is being addressed in a follow-up PR).

## Test plan

- [x] `make doctest` — 36 passed, 13 skipped
- [x] `make test-patch-cov` — full suite 1149 passed, 5 skipped (pre-existing/unrelated); patch coverage 100% on both changed files
- [x] `prek run` on changed files — all hooks pass (ruff, mypy, numpydoc-validation, etc.)
- [x] Manually verified: formula with no interaction term now raises `FormulaException` (previously constructed silently); formula with unrelated interaction term now raises; normal and reversed-order valid formulas still construct successfully

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 __pycache__
 _build
+.venv/
 .ipynb_checkpoints
 *_checkpoints
 *.DS_Store
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -280,8 +280,14 @@ class DifferenceInDifferences(BaseExperiment):
 
     def _validate_formula_interaction_terms(self) -> None:
         """
-        Validate that the formula contains at most one interaction term and no three-way or higher-order interactions.
-        Raises FormulaException if more than one interaction term is found or if any interaction term has more than 2 variables.
+        Validate that the formula contains exactly one interaction term, that it
+        is between the group and post-treatment variables, and that it is not a
+        three-way or higher-order interaction.
+
+        Raises FormulaException if no interaction term is found, if more than one
+        interaction term is found, if any interaction term has more than 2
+        variables, or if the single interaction term does not involve both the
+        group and post-treatment variables.
         """
         # Define interaction indicators
         INTERACTION_INDICATORS = ["*", ":"]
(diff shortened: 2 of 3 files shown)


daimon-pymclabs:
Thanks for this fix — the root-cause analysis is spot on and the new error message is clear and actionable. 🎉

One optional, non-blocking observation (a pre-existing pattern, not something this PR introduces):

The new check uses a raw substring test:

```python
self.group_variable_name in interaction_terms[0]
and self.post_treatment_variable_name in interaction_terms[0]
```

Because `in` is a substring match on the term string, it could in principle misfire if a group or post-treatment variable name happens to be a substring of another variable in the interaction (e.g. a group variable named `g` would match an interaction term `group*post_treatment`). This mirrors the existing pattern already used in the causal-impact read-back (the Bayesian loop and the OLS lookup do the same `in label` matching), so it's consistent with the surrounding code and totally fine to leave as-is here.

Flagging it only so it's on record: if we ever tighten this up, splitting the interaction term on `*`/`:` and comparing components exactly would make all three sites more robust. Not a blocker for this PR — just a note for a future cleanup.

daimon-pymclabs:
On reflection, the substring-matching point I raised above is really a separate, pre-existing concern rather than anything about this PR — so I've split it out into its own tracking issue: #995 (labelled `bug`).

Nothing to action here. This PR stands on its own and the new validation follows the existing pattern in the module. The issue just captures the broader cleanup (exact component matching across all three call sites) for a future change. Thanks again for the clean, well-scoped contribution! 🎉

drbenvincent:
Great work @Mari1988. Agent picked up on an existing bug, but a spun that out into a new issue. Approving!

Thanks

drbenvincent (approved):

Mari1988:
thank you. https://www.linkedin.com/in/mariln/

PR #994, opened 2026-07-11
Make OLS causal_impact lookup order-independent for the DiD interaction term

Mari1988:
## Summary

`DifferenceInDifferences.algorithm()` looked up the OLS interaction coefficient by checking whether a single concatenated string (`"group:post_treatment"`) appeared in each coefficient label. `patsy` names the interaction column with whichever variable is written first in the formula, so writing `post_treatment*group` instead of `group*post_treatment` produced an identically-fitted model but a coefficient labeled `"post_treatment[T.True]:group"` — which the concatenated-string check silently failed to match, leaving `causal_impact` as `None` with no error or warning.

The Bayesian branch immediately above already matches by checking both variable names as independent substrings, which is order-independent. This PR makes the OLS branch do the same.

This is a follow-up to #993, which fixed a related but separate bug (missing validation when no interaction term exists at all). That PR is scoped to validation at construction time; this one fixes a silent incorrect result in the OLS causal-impact computation itself.

## Changes

- `causalpy/experiments/diff_in_diff.py`: OLS branch of `algorithm()` now matches the interaction coefficient by checking both the group and post-treatment variable names independently, instead of a single concatenated substring.
- `causalpy/tests/test_integration_skl_examples.py`: new regression test confirming `group*post_treatment` and `post_treatment*group` produce identical, non-`None` `causal_impact` values.

## Test plan

- [x] `make doctest` — 36 passed, 13 skipped
- [x] `make test-patch-cov` — full suite 1149 passed, 5 skipped (pre-existing/unrelated); patch coverage 100% on both changed files
- [x] `prek run` on changed files — all hooks pass (ruff, mypy, numpydoc-validation, etc.)
- [x] Manually verified: `post_treatment*group` (OLS) now returns the same `causal_impact` value as `group*post_treatment`, matching the Bayesian estimate for the same formula

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 __pycache__
 _build
+.venv/
 .ipynb_checkpoints
 *_checkpoints
 *.DS_Store
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -236,16 +236,24 @@ class DifferenceInDifferences(BaseExperiment):
                         {"coeffs": i}
                     )
         elif self._model_backend.is_ols:
-            # This is the coefficient on the interaction term
-            # Store the coefficient into dictionary {intercept:value}
+            # This is the coefficient on the interaction term. Match by checking
+            # both variable names independently (as the Bayesian branch above
+            # does), rather than a single concatenated substring, since patsy
+            # may name the interaction term with either variable first (e.g.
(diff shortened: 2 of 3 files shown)


drbenvincent (approved):

drbenvincent:
Thanks @Mari1988 🤩

Mari1988:
Thank you! https://www.linkedin.com/in/mariln/

PR #1007, opened 2026-07-11
Treat datetime formula predictors as continuous

### LINKED ISSUE #1002: Treat bare datetime predictors as continuous in Patsy formulas
## Description

Patsy encodes a bare pandas datetime formula factor as categorical levels. CausalPy should instead interpret bare datetime predictors as continuous elapsed time, while retaining `C(date)` as the explicit syntax for date fixed effects. This policy should apply consistently to every CausalPy experiment that accepts a Patsy formula, not only ITS.

This is especially visible in PiecewiseITS, whose documentation currently shows `date` as a baseline term. The documented formula fits date dummies instead of a continuous baseline trend. In InterruptedTimeSeries, the same formula fails when predicting post-treatment observations because their datetime levels were not present in the pre-treatment design matrix.

SyntheticControl and SyntheticDifferenceInDifferences are out of scope: they use `DatetimeIndex` and `treatment_time`, not Patsy formulas.

## Steps to Reproduce

1. Create monthly data with a datetime `date` column and an outcome `y`.
2. Fit `PiecewiseITS` with `formula="y ~ 1 + date + step(date, '2020-05-01') + ramp(date, '2020-05-01')"` and inspect `result.labels`.
3. Fit `InterruptedTimeSeries` on the same data indexed by date with `formula="y ~ 1 + date"` and a treatment date before the final observation.

## Expected Behavior

- A bare datetime predictor such as `date` is represented by one continuous column, measured as elapsed days from a stable training-data origin.
- `C(date)` retains Patsy's normal categorical expansion for explicit date fixed effects.
- The stored origin is reused for out-of-sample design matrices, so ITS pre/post features remain compatible.
- `step(date, ...)` and `ramp(date, ...)` retain their existing datetime behavior.

## Actual Behavior

Patsy expands `date` into one dummy column per timestamp. PiecewiseITS can therefore fit a near-saturated date-level model instead of a baseline time trend. InterruptedTimeSeries builds the pre-period categories first and then raises `PatsyError` for unseen post-period dates.

The custom `step(date, ...)` and `ramp(date, ...)` transforms are not affected: they handle datetimes correctly, with `ramp` expressed in elapsed days.

## Scope

Apply the policy to all formula-based experiments: InterruptedTimeSeries, PiecewiseITS, DifferenceInDifferences, StaggeredDifferenceInDifferences, PanelRegression, PrePostNEGD, RegressionDiscontinuity, RegressionKink, InversePropensityWeighting, and InstrumentalVariable. Existing intentional fixed-effect usages such as `C(time)` remain categorical by explicit user choice.

## Environment

- CausalPy source: current main worktree
- Python: 3.14
- OS: macOS 26 (Darwin 25.5.0)

## Proposed Solution

Add shared formula preprocessing/design-matrix support that recognizes bare datetime factors and encodes them as elapsed days from a persisted origin. Preserve the original datetime values where they are passed to datetime-aware transforms such as `step()` and `ramp()`, and do not change explicitly categorical expressions such as `C(date)`.

Update the PiecewiseITS documentation to show the new supported behavior, add release notes for the intentional behavior change, and test continuous encoding, explicit `C(date)` encoding, and ITS out-of-sample prediction with monthly and irregular datetime series.

## Additional Context

The documented arXiv example already avoids the problem by constructing numeric `t` and passing `date` only to `ramp()`. The requested behavior would make a direct `date` term safe and unsurprising while preserving an explicit categorical escape hatch.

drbenvincent:
## Summary
- Encode bare datetime formula predictors as continuous elapsed days across Patsy-based experiments.
- Preserve `C(date)` for explicit categorical date effects and retain datetime-aware `step()` / `ramp()` behavior.
- Add regression coverage for datetime encoding and ITS post-treatment prediction.

Fixes #1002

## Testing
- `prek run --all-files`
- `make test-patch-cov`

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -75,7 +75,7 @@ Instantiation fits eagerly in `__init__`: `_build_design_matrices()` → `_prepa
 
 | Topic | Detail |
 |-------|--------|
-| **Formulas** | Patsy `dmatrices()` for design matrices; `build_design_matrices()` for counterfactual prediction. `PiecewiseITS` uses `step()`/`ramp()` stateful transforms. |
+| **Formulas** | Patsy `dmatrices()` for design matrices; `build_design_matrices()` for counterfactual prediction. Bare datetime predictors are encoded as continuous elapsed days from the fitted origin; use `C(date)` for date fixed effects. `PiecewiseITS` uses `step()`/`ramp()` stateful transforms. |
 | **obs_ind** | All experiments set `data.index.name = "obs_ind"`. Canonical xarray/PyMC dimension name. |
 | **treated_units always 2D** | Even single-unit experiments use `treated_units=["unit_0"]`. Never pass 1D y to PyMC. |
 | **Impact uses mu, not y_hat** | `calculate_impact()` subtracts posterior `mu` (expected value), not `y_hat` (with observation noise). |
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -22,7 +22,7 @@ import pandas as pd
 import seaborn as sns
 import xarray as xr
 from matplotlib import pyplot as plt
-from patsy import ModelDesc, build_design_matrices, dmatrices
+from patsy import ModelDesc, build_design_matrices
 from sklearn.base import RegressorMixin
 
 from causalpy.constants import HDI_PROB, LEGEND_FONT_SIZE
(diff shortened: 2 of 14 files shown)


drbenvincent:
## Automated triage

**Why:**
- Cross-cutting bug fix touching 11 experiment files plus new `formula_utils.py` and `transforms.py` — affects how datetime predictors are handled across all Patsy-based experiments.
- Public API behaviour change: bare datetime formula predictors are now encoded as continuous elapsed days instead of raising or silently misbehaving.
- All CI checks pass, but the PR is blocked (merge conflict).

**Review focus:**
1. Verify datetime encoding edge cases (timezones, missing dates, single-row data).
2. Confirm that `C(date)` and `step()`/`ramp()` still work as documented.
3. Check that the new `formula_utils.py` module is covered by the regression tests.

**Confidence:** high

PR #1014, opened 2026-07-14
Add pymc_forecast model-provider adapter for InterruptedTimeSeries

### LINKED ISSUE #1013: Add a pymc_forecast model-provider adapter (ITS/SC counterfactual backend)
## Summary

Add a thin adapter that lets a `pymc_forecast` forecasting model act as a **model provider** behind CausalPy's existing experiment API, in the same spirit as the scikit-learn wrapper (`causalpy/skl_models.py` + `experiments/model_adapter.py`). CausalPy keeps identification, counterfactual construction, and placebo methods; `pymc_forecast` provides the fitted forecasting model.

This is the outcome of a design discussion with the `pymc_forecast` maintainers. The five upstream API changes CausalPy needs are implemented, reviewed, and shipped (see below).

> **Package naming:** the PyPI distribution is **`pymc-forecast`** (hyphen); the import name is `pymc_forecast` (underscore). Install with `pip install pymc-forecast>=0.1.0`.

## Scope — first pass is ITS-only

**In (this issue):** `pymc_forecast` as a selectable time-series **model backend** for `InterruptedTimeSeries`.

**Deferred (follow-up):** `SyntheticControl` as a second backend target. Left out of the first pass to keep the adapter minimal and prove the boundary on one experiment before generalizing. Tracked as a follow-up once ITS lands.

**Out (both passes):** no outsourcing of placebo-in-time, power analysis, or power curves. Placebo-in-time is a cross-method quasi-experimental primitive and stays in CausalPy; back-tests/cross-validation are not causal tests and must not be reframed as such. Keep the boundary at model-provider + forecasting metrics.

## Adapter mapping

The ITS experiment (`experiments/interrupted_time_series.py`) drives a backend through `fit(X, y, coords)`, `predict(X)` (in-sample, pre-period), `predict(X, out_of_sample=True)` (the counterfactual, post-period), and `score(X, y)`. Mapping onto `pymc_forecast`:

- `fit(X, y)` → construct + fit the forecasting model on the pre-period.
- `predict(X)` (pre) → `predict_in_sample(...)`; take the `posterior_predictive` group.
- `predict(X, out_of_sample=True)` (post, "as if untreated") → `.forecast(future_covariates=post_X)` when the model has exogenous regressors, or `.forecast(future_index=post_index)` for a covariate-free trend/seasonal counterfactual.
- Extract draw-level samples via `prediction_samples(result)`, then rename the documented dims onto CausalPy coords: `time_future` → `obs_ind`, `series` → `treated_units`.
- Priors flow through the model object (pymc-extras `Prior`), so users keep the transparent-prior ethos.

## Upstream dependencies (pymc-labs/pymc_forecast) — shipped in 0.1.0

- future exogenous regressors on `.forecast(future_covariates=)` — pymc-labs/pymc_forecast#26 (closes #19)
- full draw-level posterior-predictive samples + `prediction_samples()` — pymc-labs/pymc_forecast#27 (closes #20)
- documented, test-locked output schema + exported name constants — pymc-labs/pymc_forecast#28 (closes #21)
- horizon-agnostic `.forecast(future_index=)` — pymc-labs/pymc_forecast#25 (closes #22)
- user-injectable priors via the pymc-extras `Prior` API — pymc-labs/pymc_forecast#34 (closes #23)

## Tasks (ITS-only first pass)

- [x] Upstream API shipped in a release: `pymc-forecast==0.1.0`.
- [ ] Add `pymc-forecast>=0.1.0` as an **optional extra** (not a hard dep), mirroring the pymc-marketing extra.
- [ ] Implement the adapter (target: comparable in size to the sklearn wrapper).
- [ ] Wire it as a selectable backend for `InterruptedTimeSeries`; no change to the public API/UX ethos.
- [ ] Round-trip test: fit pre / forecast post-as-untreated / `calculate_impact` on draw-level samples, checked against the existing pymc BSTS path.
- [ ] Docs note on when to reach for this backend vs the existing `PyMCModel` classes (ties into #992 method-routing).
- [ ] Follow-up (separate issue): extend the adapter to `SyntheticControl` once ITS is proven.

## Caveats

- `pymc-forecast` is young (0.1.0); introduce behind an optional extra and a version pin, mirroring how we already manage pymc/pytensor pins. Watch for API movement across 0.x minors — the output schema is the one surface upstream has committed to keeping stable (contract-tested, changelog-gated).
- The interval-calibration-under-input-scaling concern (raised externally re: synthetic control) stays a CausalPy-side responsibility; the upstream prior injectability is the mechanism to plumb it, not a fix.

## Related

Feeds/relates to #982 (graduate a BSTS default), #980 (exogenous regressors on `StateSpaceTimeSeries`), and the temporal-validation idea in #367.

twiecki:
Closes #1013.

Implements the ITS-only first pass: a thin adapter that lets a `pymc_forecast` forecasting model act as a **model provider** behind CausalPy's existing experiment API, in the same spirit as the sklearn wrapper. CausalPy keeps identification, counterfactual construction, and placebo methods; `pymc_forecast` provides the fitted forecasting model.

## What's in here

- **`causalpy/pymc_forecast_models.py`** — `PyMCForecastModel`, a thin wrapper (comparable in size to `skl_models.py`) around a `pymc_forecast` model function / `ForecastingModel` plus a forecaster class (NUTS `HMCForecaster` by default, matching CausalPy's native backends; ADVI/Pathfinder are explicit opt-ins). Maps the experiment protocol onto the upstream 0.1.0 API exactly as scoped in the issue:
  - `fit(X, y)` → construct + fit the forecaster on the pre-period; the patsy design matrix rides along as covariates (`obs_ind → time`, `coeffs → covariate`).
  - `predict(X)` (pre) → `predict_in_sample()`, `posterior_predictive` group.
  - `predict(X, out_of_sample=True)` (post, "as if untreated") → `forecast(future_covariates=post_X)` with exogenous regressors, or `forecast(future_index=post_index)` for a covariate-free trend model.
  - Draw-level samples come out of `prediction_samples(result)` and the documented schema dims are renamed onto CausalPy coords (`time_future → obs_ind`, `series → treated_units`).
  - Priors flow through the model object (pymc-extras `Prior`), keeping the transparent-prior ethos.
- **`PyMCForecastAdapter`** in `experiments/model_adapter.py`, gated by a new `supports_pymc_forecast` flag — enabled for `InterruptedTimeSeries` only (`SyntheticControl` is a tracked follow-up). No change to the public API/UX: users just pass `model=PyMCForecastModel(...)` to `cp.InterruptedTimeSeries`.
- **Optional extra, not a hard dep**: `pip install causalpy[forecast]` → `pymc-forecast>=0.1.0,<0.2` (minor pinned while upstream is 0.x; the output schema is the surface upstream contract-tests). Also added to the `test` extra so CI exercises the adapter; the test module `importorskip`s it.
- **Round-trip tests** (`test_pymc_forecast_adapter.py`, 12 tests): fit pre / forecast post-as-untreated / `calculate_impact` on draw-level samples, with the output contract and recovered effect checked against the existing PyMC `LinearRegression` path on the same synthetic data; plus the covariate-free `future_index` path, three-period designs, plot/`summary()`/`effect_summary()`/`get_plot_data_bayesian()` smoke, single-unit validation, and adapter resolution/gating.
- **Docs**: module docstring carries the "when to reach for this backend vs the existing `PyMCModel` classes" guidance (referenced from the ITS docstring), and the module is added to the API reference index.

## Out of scope (per the issue)

No outsourcing of placebo-in-time, power analysis, or power curves — the boundary stays at model-provider + forecasting metrics.

## Caveats

Coefficient printing reports posterior summaries of the model's scalar parameters instead of design-matrix coefficients.

~~`pymc_forecast` exposes the draw-level posterior predictive of the observed variable rather than a separate noise-free `mu` ...~~ **Resolved with pymc-forecast 0.2**: `mu` now carries the upstream noise-free latent predictor (`mu`/`mu_future`), so causal impact excludes observation noise exactly like the native backends, one posterior subsample is shared by the pre-period fit and the counterfactual (draw coherence via `posterior=`), and the forecaster uses the deferred-fit lifecycle. Pinned `pymc-forecast>=0.2,<0.3`.

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,6 +1,6 @@
 # ARCHITECTURE
 
-CausalPy implements 10+ quasi-experimental causal inference methods over two statistical backends (PyMC and scikit-learn). This document orients agents and contributors to where things live, how the pieces compose, and the non-obvious conventions.
+CausalPy implements 10+ quasi-experimental causal inference methods over two core statistical backends (PyMC and scikit-learn), plus an optional third backend (`pymc-forecast`, currently `InterruptedTimeSeries` only). This document orients agents and contributors to where things live, how the pieces compose, and the non-obvious conventions.
 
 ## Module Map
 
@@ -10,6 +10,7 @@ CausalPy implements 10+ quasi-experimental causal inference methods over two sta
 | `causalpy/experiments/` | One experiment class per file; `base.py` holds `BaseExperiment` |
 | `causalpy/pymc_models.py` | All `PyMCModel` subclasses (Bayesian backend) |
 | `causalpy/skl_models.py` | `ScikitLearnAdaptor` mixin, `create_causalpy_compatible_class()` |
+| `causalpy/pymc_forecast_models.py` | `PyMCForecastModel` wrapper (optional `pymc-forecast` backend) |
 | `causalpy/reporting.py` | `EffectSummary`, statistics and prose for both backends |
 | `causalpy/pipeline.py` | `Pipeline`, `PipelineContext`, `PipelineResult`, `Step` protocol |
 | `causalpy/steps/` | `EstimateEffect`, `SensitivityAnalysis`, `GenerateReport` |
@@ -21,9 +22,9 @@ CausalPy implements 10+ quasi-experimental causal inference methods over two sta
 | `docs/source/notebooks/` | How-to notebooks (`{method}_{backend}.ipynb`) |
 | `docs/source/knowledgebase/` | Educational content (glossary, reporting explainers) |
 
-## Two-Backend Model
+## Backend Model
 
-Backend dispatch is centralized in `causalpy/experiments/model_adapter.py`. `BaseExperiment.__init__` calls `make_model_adapter()`, which handles sklearn coercion (`clone`/`deepcopy`, `create_causalpy_compatible_class()`, `fit_intercept=False` warning), default-model instantiation, and `supports_bayes`/`supports_ols` validation. Each experiment stores `self._model_backend` (private) and keeps `self.model` as the public handle.
+Backend dispatch is centralized in `causalpy/experiments/model_adapter.py`. `BaseExperiment.__init__` calls `make_model_adapter()`, which handles sklearn coercion (`clone`/`deepcopy`, `create_causalpy_compatible_class()`, `fit_intercept=False` warning), default-model instantiation, and `supports_bayes`/`supports_ols`/`supports_pymc_forecast` validation. Each experiment stores `self._model_backend` (private) and keeps `self.model` as the public handle.
 
 Standard regression experiments call `self._model_backend.fit(X, y, coords=build_coords(...))` unconditionally. `build_coords()` assembles the PyMC `coeffs` / `obs_ind` / `treated_units` dict; sklearn backends ignore `coords`. `SklearnModelAdapter` normalizes inputs before delegating to sklearn: xarray `DataArray` values become numpy arrays, and a single-column `treated_units` outcome is squeezed to 1D so call sites do not need per-experiment `.isel(treated_units=0)` branches.
 
@@ -40,6 +41,8 @@ self._model_backend.fit(
 Experiments with non-standard fit signatures bypass this path and call `self.model.fit(...)` directly with custom arguments: `InstrumentalVariable` (two-stage IV), `InversePropensityWeighting` (propensity `fit(X, t, coords)`), and `SyntheticDifferenceInDifferences` (dict-shaped weight-fitter inputs). Those models are not forced through `build_coords` or sklearn y-normalization.
 
 `PyMCModel` extends `pymc.Model` with a sklearn-like `fit` / `predict` / `score` / `calculate_impact` interface. `ScikitLearnAdaptor` is a mixin patched onto any `RegressorMixin` via `create_causalpy_compatible_class()` during adapter construction. Every experiment declares `supports_ols` and `supports_bayes`; validation runs in `make_model_adapter()`. When `model=None`, `_default_model_class` is instantiated (always Bayesian; `PanelRegression` requires an explicit model).
+
+The optional third backend, `PyMCForecastModel` (`causalpy/pymc_forecast_models.py`), wraps a `pymc_forecast` forecasting model behind the same protocol and is wired through `PyMCForecastAdapter`. It reports as Bayesian (`is_bayesian` is true for both `"pymc"` and `"pymc-forecast"` adapter kinds), so experiments and checks that branch on Bayesian-vs-OLS treat it like a PyMC backend. Experiments opt in via `supports_pymc_forecast` (currently `InterruptedTimeSeries` only); the dependency ships as the `causalpy[forecast]` extra, pinned to one upstream minor while `pymc-forecast` is 0.x.
 
 ## Experiment Lifecycle
 
@@ -88,7 +91,7 @@ Instantiation fits eagerly in `__init__`: `_build_design_matrices()` → `_prepa
 
 Copy the closest existing experiment or model and follow the `BaseExperiment` contract:
 
-- Declare `supports_ols` / `supports_bayes`; implement `_bayesian_plot()` / `_ols_plot()` only for supported backends
+- Declare `supports_ols` / `supports_bayes` (and `supports_pymc_forecast` to opt into the optional pymc-forecast backend); implement `_bayesian_plot()` / `_ols_plot()` only for supported backends
 - `algorithm()` with the fit/predict/impact flow; `effect_summary()` via helpers in `causalpy.reporting`
 - Public `plot(*, ...)` with a kwarg-only signature that delegates to `_render_plot()` — bare `*args` / `**kwargs` are forbidden on the public surface (enforced by `causalpy/tests/test_public_plot_signatures.py`). For experiments without a unified plot view (e.g. `InversePropensityWeighting`, `InstrumentalVariable`), declare an explicit `plot()` stub that raises `NotImplementedError`. For `hdi_prob` defaults, use ``Defaults to :data:`~causalpy.constants.HDI_PROB` (currently 0.94).`` in the docstring.
 - Raise `FormulaException`, `DataException`, or `BadIndexException` from `causalpy.custom_exceptions` for formula, data, and index errors
--- a/causalpy/__init__.py
+++ b/causalpy/__init__.py
@@ -14,6 +14,7 @@
 """CausalPy: causal inference for quasi-experiments in Python."""
 
 import causalpy.checks as checks  # noqa: E402
+import causalpy.pymc_forecast_models as pymc_forecast_models
 import causalpy.pymc_models as pymc_models
 import causalpy.skl_models as skl_models
 import causalpy.variable_selection_priors as variable_selection_priors
@@ -64,6 +65,7 @@ __all__ = [
     "PanelRegression",
     "plot_correlations",
     "PrePostNEGD",
+    "pymc_forecast_models",
     "pymc_models",
     "ramp",
     "RegressionDiscontinuity",
--- a/causalpy/checks/placebo_in_time.py
+++ b/causalpy/checks/placebo_in_time.py
@@ -29,8 +29,8 @@ Supports two fold-selection strategies:
   fraction, minimum gap between folds, and optional period exclusion.
 
 Supports experiments with a ``treatment_time`` parameter
-(InterruptedTimeSeries, SyntheticControl).  Requires a PyMC model
-for posterior extraction.
+(InterruptedTimeSeries, SyntheticControl).  Requires a Bayesian model
+backend (PyMC or pymc-forecast) for posterior extraction.
 """
 
 from __future__ import annotations
@@ -50,7 +50,6 @@ from causalpy.experiments.base import BaseExperiment
 from causalpy.experiments.interrupted_time_series import InterruptedTimeSeries
 from causalpy.experiments.synthetic_control import SyntheticControl
 from causalpy.pipeline import PipelineContext
-from causalpy.pymc_models import PyMCModel
 
 logger = logging.getLogger(__name__)
 
@@ -300,8 +299,8 @@ class PlaceboInTime:
         Raises
         ------
         TypeError
-            If the experiment lacks ``treatment_time`` or uses a non-PyMC
-            model.
+            If the experiment lacks ``treatment_time`` or uses a
+            non-Bayesian model backend.
         """
         if not hasattr(experiment, "treatment_time"):
             raise TypeError(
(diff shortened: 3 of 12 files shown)


drbenvincent:
## Automated triage

**Why:**
- **Public API + behaviour change**: Adds a new first-class causal backend (`PyMCForecastModel`) exposed via `causalpy.__init__` and integrated into `InterruptedTimeSeries` (9 files, ~820 lines). This is a new integration surface, not a contained internal refactor.
- **Dependency change**: Introduces `pymc-forecast>=0.1.0,<0.2` as an optional extra in `pyproject.toml` and `environment.yml`, changing the install/CI-surface for the project.
- **CI failure**: The `test (3.11)` job failed and `test (3.14)` was cancelled. The cause is not obvious from the metadata — needs confirmation whether this is flaky or a genuine regression introduced by the new adapter.

**Review focus:**
1. Confirm the failing CI (`test (3.11)`) — is it related to the new adapter or pre-existing?
2. Verify that the adapter boundary (fit/predict schema mapping) correctly isolates CausalPy identification logic from `pymc_forecast` model fitting, especially the `mu`/observation-noise caveat documented in the module docstring.
3. Check that the optional-extra dependency pin (`>=0.1.0,<0.2`) is appropriate for the upstream's pre-1.0 API stability.

**Confidence:** high

twiecki:
The upstream changes this adapter requested are now in: pymc-labs/pymc-forecast#44 implemented all four feature requests (pymc-labs/pymc-forecast#36 noise-free `mu`/`mu_future` in the output schema, pymc-labs/pymc-forecast#37 `posterior=` passthrough for draw-coherent predictions, pymc-labs/pymc-forecast#38 ADVI non-convergence warning, pymc-labs/pymc-forecast#39 uniform `progressbar=` + deferred fit) and is merged; the 0.2.0 release PR (pymc-labs/pymc-forecast#45) is green and awaiting merge + PyPI publish.

This PR is updated accordingly (latest commit):

- **Noise-free impact**: `mu` is now the upstream latent predictor, `y_hat` the posterior predictive — the documented observation-noise caveat is gone, and impact HDIs match the native backends' convention. A new test locks `std(mu) < std(y_hat)` and `std(impact) == std(mu)`.
- **Draw coherence**: one posterior subsample is drawn at fit time and passed to both `predict_in_sample` and `forecast` via `posterior=`; a new test verifies pre/post `mu` reproduce `X @ beta` from the shared draws draw-for-draw and that repeated `predict()` calls are bit-identical.
- **Deferred fit**: `PyMCForecastModel` now constructs the (unfitted) forecaster at construction and calls `fit(data, covariates)`, dropping the carried `(class, kwargs)` recipe; invalid options fail early. The default forecaster resolves to `StatespaceForecaster` for `StatespaceModel` definitions.
- Pins bumped to `pymc-forecast>=0.2,<0.3`.

Note: CI will stay red on the dependency resolve until `pymc-forecast==0.2.0` is published to PyPI (release PR pymc-labs/pymc-forecast#45 is ready). All 14 adapter tests pass locally against 0.2.0.

drbenvincent (commented):
Thanks for this — the core direction looks right. `pymc-forecast` should remain a distinct, experiment-gated backend rather than inherit from `PyMCModel`: it owns a forecaster with a different fit/forecast lifecycle, coefficient contract, and inference-result shape. Once its outputs are normalized, treating it as Bayesian through `is_bayesian` is appropriate. `supports_pymc_forecast` is a reasonable incremental gate for the ITS-only first pass; before adding further providers, we should replace the parallel booleans with an explicit set of supported backend kinds.

Three areas need addressing before merge:

## 1. Make the existing CausalPy placebo-in-time path work

The PR/issue says that placebo methods remain CausalPy's responsibility, but `PlaceboInTime.validate()` currently rejects `PyMCForecastModel` because it checks `isinstance(model, PyMCModel)`. Refitting also fails because `PyMCForecastModel` has no `_clone()` and its fallback `deepcopy()` raises `TypeError: cannot pickle 'module' object`.

This is a small CausalPy-only fix, not a new upstream feature: add an `_clone()` that reconstructs an unfitted wrapper from its stored configuration, validate against the backend's Bayesian capability, and add coverage that a forecast-backed ITS can be cloned and accepted/refitted by `PlaceboInTime`. Track broader support for other sensitivity checks separately.

## 2. Do not alias noisy statespace predictions to `mu`

For `StatespaceForecaster`, the adapter currently falls back from a missing noise-free expected observation to `y_hat`, then exposes that as `mu`. ITS consequently computes causal impact with observation noise included, contrary to CausalPy's `mu`-based impact contract.

This requires a new upstream `pymc-forecast` capability: statespace prediction currently exposes latent-state draws and noisy observed draws, but not the expected observation in observation space with measurement error removed. Please open and link an upstream issue requesting draw-level in-sample and future expected-observation outputs that:

- apply the state-space observation equation rather than returning the latent state vector;
- exclude measurement error;
- remain conditioned on the same supplied `posterior=` draws;
- preserve matching chain/draw/time/series coordinates; and
- align with the core forecaster's `mu` / `mu_future` schema where possible.

Until that capability exists, please reject or stop advertising the statespace path here rather than silently substituting `y_hat` for `mu`. Core `pymc-forecast` models built with `predict()` already provide proper `mu` / `mu_future` and can remain supported.

## 3. Complete the new backend's central contracts

- Update `ARCHITECTURE.md`; it still describes two backends and its maintenance rule explicitly requires updates when backend dispatch changes.
- Update the central `BaseExperiment` and `ModelAdapter.model` annotations/docstrings, which still exclude `PyMCForecastModel`.
- Align the optional dependency with the documented API. Upstream places statespace/Pathfinder support behind `pymc-forecast[extras]` with `pymc-extras>=0.10`, while this PR installs bare `pymc-forecast` and CausalPy permits `pymc-extras>=0.3.0`.
- Make inference diagnostics discoverable. The full HMC/Pathfinder fit remains available at `model.forecaster.idata` while `result.idata` contains the draw-coherent prediction subset; document that distinction or expose a direct `fit_idata`/`fit_result` property. No `result.idata` redesign is required in this PR.

Maketables support can be a separate CausalPy issue. Forecasting parameters do not necessarily map to Patsy coefficients, so that needs an explicit table-semantics decision rather than just another dispatch branch. For this PR, documenting the limitation and raising an intentional error is enough.

The distribution/import naming (`pymc-forecast` / `pymc_forecast`), optional minor pin, draw-coherence work, output normalization, and ITS integration tests all look appropriate. CI is green. With the placebo/refit path, statespace semantic boundary, and central architecture contracts addressed, this will establish a sound direction for future forecast-backed experiments.

twiecki:
Thanks for the review @drbenvincent — all four points addressed in 9f09035:

1. **PlaceboInTime**: `validate()` now checks the backend capability (`_model_backend.is_bayesian`) instead of `isinstance(model, PyMCModel)`, and `PyMCForecastModel` gained `_clone()` so `clone_model()` reconstructs an unfitted wrapper from its configuration. Covered by a real 2-fold `PlaceboInTime.run()` test asserting each fold received a fresh, refitted `PyMCForecastModel`.

2. **Statespace `mu` aliasing**: went with reject-plus-upstream-issue rather than silently substituting the noisy predictive. `StatespaceModel` / `StatespaceForecaster` now raise `NotImplementedError` at construction, the `mu`/`mu_future` lookups are strict (a missing latent fails loudly), and the request for draw-level expected observations excluding measurement error is filed as pymc-labs/pymc-forecast#50 — once that ships, the rejection lifts.

3. **Docs/annotations**: ARCHITECTURE.md now covers the third backend (intro, module map, a dedicated backend paragraph, and the `supports_pymc_forecast` note in the authoring checklist); `BaseExperiment` / `ModelAdapter.model` annotations and docstrings include `PyMCForecastModel`. Dependency pins stand at `pymc-forecast>=0.2,<0.3` in both the `forecast` and `test` extras, matching the API the module documents.

4. **Diagnostics discoverability**: added a `fit_idata` property exposing the full forecaster fit result (complete NUTS `InferenceData` with sample stats), documented against `idata` (the thinned, draw-coherent posterior subsample used for prediction). Raises informatively when unfitted or when the forecaster doesn't retain an `InferenceData` (variational fits point to `.forecaster.approx` / `.losses`).

118 tests green locally across the adapter, placebo-in-time, model-adapter, panel-regression, and guard-script suites, plus doctests and the full hook set (mypy, numpydoc, ruff, architecture/exports).

drbenvincent:
Thanks @twiecki — I rechecked the follow-up commit and the substantive review points are addressed: placebo refitting is covered, statespace now fails safely with pymc-forecast#50 tracking the missing capability, and the architecture, typing, diagnostics, tests, and current CI look good. The upstream statespace issue does not need to block this PR.

The one remaining item is the optional dependency contract: the adapter advertises Pathfinder, but the extras still install bare `pymc-forecast` while allowing `pymc-extras>=0.3`; upstream declares these integrations through `pymc-forecast[extras]` with `pymc-extras>=0.10`. Please either use `pymc-forecast[extras]>=0.2,<0.3` in the forecast/test extras or confirm and document why the lower floor is supported. Once that is resolved and the updated CI remains green, I’m happy to approve this for merge.

twiecki:
@drbenvincent done.

drbenvincent:
One final scope clarification before approval, following a broader look at CausalPy's prediction semantics.

**Current behavior:** CausalPy's `calculate_impact()` interprets `mu` as the conditional expected outcome in observed outcome units and computes `observed - mu`. Most existing built-in models satisfy that convention because they use Gaussian identity links. However, this is currently an implicit convention rather than an enforced backend contract.

**Boundary exposed by this PR:** `pymc-forecast` documents `mu` / `mu_future` as the latent passed to `predict()`, which may be a link-scale linear predictor for a GLM. The Gaussian identity-link path implemented and tested here is correct. But for a linked model, directly consuming a link-scale `mu` would mix units—for example, observed counts minus a log-rate.

No adapter redesign or upstream release should block this PR. Please add a short note documenting the current requirement: models using a link function must pass the inverse-linked conditional expectation in observed outcome units to `pymc_forecast.predict()`. Once that scope is explicit and CI remains green, I am happy to approve.

We have opened follow-up issues so the temporary convention and the intended long-term design are historically connected:

- pymc-labs/CausalPy#1016 — define outcome-scale prediction semantics across CausalPy backends, audit current behavior, and add contract tests.
- pymc-labs/CausalPy#1017 — add a g-computation tutorial showing correct and incorrect causal contrasts with link functions.
- pymc-labs/pymc-forecast#52 — expose an explicit outcome-scale expected observation separately from the generic latent predictor.
- pymc-labs/CausalPy#1018 — require sorted, unique ITS indexes; this is independent of the link-function contract but was exposed while reviewing the forecasting integration.

The existing statespace limitation remains tracked in pymc-labs/pymc-forecast#50.

drbenvincent:
FYI: No action needed from @twiecki. I'm going to get an agent to blast through this and the issues it generated

Issue #1016, opened 2026-07-18
Define outcome-scale prediction semantics across model backends


drbenvincent:
# Define outcome-scale prediction semantics across model backends

## Problem

CausalPy currently calculates Bayesian impact as `observed - posterior_predictive["mu"]`. The `PyMCModel.calculate_impact()` documentation defines `mu` as the expected outcome in observed outcome units, excluding observation noise. Most built-in models satisfy this implicitly because they use Gaussian likelihoods with identity links, where the linear predictor and conditional expected outcome coincide.

That meaning is not represented by a formal backend contract or enforced by shared tests. A custom model can instead expose a link-scale linear predictor as `mu`, and some state-space paths alias `mu` to `y_hat`. PR #1014 exposes the ambiguity at a package boundary: `pymc-forecast` deliberately defines `mu` / `mu_future` as the latent passed to `predict()`, which may be a GLM linear predictor rather than the expected observation.

For example, with a Poisson log-link model, subtracting `eta = X @ beta` from an observed count mixes log-rate and count units. CausalPy needs `exp(eta) = E[Y | ...]` for its current additive outcome-scale impact.

## Current behavior

- Bayesian impact is calculated from a variable named `mu`.
- CausalPy documents `mu` as the expected outcome without observation noise.
- Built-in Gaussian identity-link models generally satisfy that convention.
- The convention is not enforced across native PyMC, sklearn, state-space, custom, or provider-backed models.
- CausalPy currently reports additive effects in observed outcome units.

## Desired behavior

CausalPy should define backend-independent semantics for at least:

- the linear predictor or latent state;
- posterior draws of the conditional expected outcome in observed units;
- posterior-predictive draws that include observation noise.

Impact calculations should consume the explicitly defined expected-outcome quantity, not infer its meaning from the name `mu`. Link transformations must be applied draw by draw before contrasts or posterior summaries are calculated.

## Proposed work

- Decide the canonical prediction schema and whether the public name remains `mu` or becomes an explicit name such as `expected_outcome`.
- Centralize or otherwise enforce impact calculation against the outcome-scale conditional expectation.
- Audit built-in PyMC models, sklearn adapters, `StateSpaceTimeSeries`, and `PyMCForecastAdapter`.
- Add shared backend contract tests for dimensions, coordinates, exclusion of observation noise, draw coherence, and outcome-scale semantics under a non-identity link.
- Define a backwards-compatibility plan for existing `mu` consumers.
[...]

PR #1024, opened 2026-07-18
Add built-in GeneralizedLinearRegression for count and binary outcomes (#1022)

### LINKED ISSUE #1022: Add GeneralizedLinearRegression with curated family and link support
## Problem

CausalPy's native PyMC regression models expose a configurable `y_hat` likelihood through `pymc_extras.prior.Prior`, but the built-in `LinearRegression` graph always constructs `mu = Xβ` with an identity link. Replacing the likelihood alone is therefore insufficient for generalized linear regression: a Poisson `Prior("Poisson")` on `y_hat` still feeds an unconstrained linear predictor into `mu`, which violates both the distribution's support and CausalPy's outcome-scale prediction contract.

The prediction contract (#1016, implemented in #1020) requires that `posterior_predictive["mu"]` be the draw-level conditional expectation `E[Y | parameters, covariates]` on the **response scale**, while `y_hat` remains the noisy likelihood / posterior-predictive variable. Experiments such as `InterruptedTimeSeries` compute causal impact as `y - mu` via `PyMCModel.calculate_impact()`. For count or binary outcomes, `mu` must be the inverse-linked expectation (for example `exp(η)` for Poisson), not the link-scale linear predictor `η`.

Contract tests in `causalpy/tests/test_prediction_contract.py` encode the intended semantics using custom `PyMCModel` subclasses, but there is no public built-in API that wires family, link, and likelihood together. Issue #1017 (customer-support ticket counts ITS tutorial) is blocked on this gap.

## Goal

Add first-class generalized linear regression support for native PyMC regression models through a new `GeneralizedLinearRegression` class, while keeping `LinearRegression` backward compatible as the Gaussian / identity specialization.

Users should be able to fit quasi-experimental models with non-Gaussian outcomes without writing a custom `PyMCModel` subclass, and CausalPy-generated impact, plots, and summaries must remain on the response scale (ticket counts, probabilities, rates).

## Proposed public API

Introduce `causalpy.pymc_models.GeneralizedLinearRegression` with a curated family / link surface:

```python
import causalpy as cp
from causalpy.pymc_models import GeneralizedLinearRegression
from pymc_extras.prior import Prior

model = GeneralizedLinearRegression(
    family="poisson",          # default link: "log"
    # link="log",              # optional override when valid for the family
  sample_kwargs={"random_seed": 42, "progressbar": False},
    priors={
        "beta": Prior("Normal", mu=0, sigma=2.5, dims=["treated_units", "coeffs"]),
        # family-specific auxiliary priors as needed, e.g. "alpha" for NegativeBinomial
    },
)

result = cp.InterruptedTimeSeries(
    data,
    treatment_time,
    formula="tickets ~ 1 + t + C(dow) + active_users",
    model=model,
)
```

### Curated families (initial release)

| `family` | Default `link` | Response parameter passed to likelihood | Notes |
|----------|----------------|----------------------------------------|-------|
| `"gaussian"` | `"identity"` | `mu` | Equivalent to current `LinearRegression` behavior; supports existing `y_hat` `Prior("Normal", sigma=...)` customization |
| `"poisson"` | `"log"` | `mu` | Count outcomes; `mu = exp(eta)` must be strictly positive |
| `"negative_binomial"` | `"log"` | `mu` | Overdispersed counts; expose dispersion via `priors["alpha"]` or equivalent documented default |
| `"bernoulli"` | `"logit"` | `p` | Binary outcomes; graph still exposes outcome-scale `mu` as the probability for impact / contract compliance |

Supported `link` literals for v1: `"identity"`, `"log"`, `"logit"`. Validate family / link pairings at construction time (reject invalid combinations rather than failing silently at sampling). Default `link=None` means use the family canonical link.

`LinearRegression` remains the existing public class and default backend for experiments that do not pass an explicit model. Internally it should delegate to the shared generalized regression implementation (or subclass it) so behavior does not drift.

### Required model graph

For all supported families, `build_model()` must follow the prediction contract:

1. Compute the linear predictor `eta = X @ beta` (optionally exposed as a `Deterministic` named `eta`).
2. Apply the configured inverse link draw by draw to obtain the response-scale expectation.
3. Register that quantity as `pm.Deterministic("mu", ..., dims=["obs_ind", "treated_units"])`.
4. Construct the family likelihood as `y_hat`, using `Prior.create_likelihood_variable` when the distribution accepts a `mu` parameter and a documented branch when it requires `p` (Bernoulli).

`PyMCModel.predict()` must continue to return both `posterior_predictive["mu"]` and `posterior_predictive["y_hat"]` with dims `(chain, draw, obs_ind, treated_units)`.

### Shared regression head

Extract a small internal helper used by `GeneralizedLinearRegression` (and reused by `LinearRegression`) rather than duplicating graph construction. Suggested responsibilities:

- `inverse_link(eta, link)` for `"identity"`, `"log"`, and `"logit"` using PyMC / PyTensor ops (`pt.exp`, `pm.math.sigmoid` / `invlogit`).
- `build_regression_head(X, beta, link) -> (eta, mu)` returning link-scale and response-scale tensors with correct dims.
- `create_family_likelihood(family, mu_or_p, y_obs, priors)` handling `mu`-parameterized and `p`-parameterized distributions.

Keep the surface minimal: no Bambi dependency, no public custom `Family` objects in v1, no statsmodels GLM wrapper.

### Prior customization

Preserve the existing `priors=` dict pattern:

- `"beta"` (and any other coefficient priors) as today.
- `"y_hat"` overrides for Gaussian ancillary parameters (`sigma`) where applicable.
- Documented auxiliary keys per family (for example `"alpha"` for NegativeBinomial).

### Clone / refit

Sensitivity checks (`PlaceboInTime`, `PriorSensitivity`, etc.) clone models via `_clone()`. `GeneralizedLinearRegression._clone()` must copy `family`, `link`, `sample_kwargs`, and `_user_priors` so refits preserve the configured GLM.

### `score()` behavior

`PyMCModel.score()` currently computes Bayesian R² from `mu` and is meaningful for continuous Gaussian outcomes. For non-Gaussian families, either document the limitation, emit a targeted warning, or gate / skip scoring. Do not return silently misleading R² values for Bernoulli or count models.

### In scope

- `GeneralizedLinearRegression` with curated families and links listed above.
- Refactor `LinearRegression` to share the generalized implementation without changing default behavior.
- Public export from `causalpy.pymc_models` and `causalpy` (subject to existing export checks).
- Knowledgebase / API documentation cross-linked from `prediction-contract.md`.
- Fast pytest coverage:
  - Unit tests for link helper and family / link validation.
  - Contract tests mirroring `test_prediction_contract.py` for each curated family on the **built-in** class (not only test subclasses).
  - At least one `InterruptedTimeSeries` integration test with a count outcome showing outcome-scale impact recovery within posterior uncertainty.
- Unblocks #1017 once merged.

### Out of scope (explicit exclusions)

- Adding `family=` / `link=` kwargs directly on experiment classes (`InterruptedTimeSeries`, `DifferenceInDifferences`, etc.). Users configure the model object; experiments stay agnostic.
- Synthetic-control weight models (`WeightedSumFitter`, `SoftmaxWeightedSumFitter`): donor-weighted sums are already outcome-scale counterfactuals; a global outcome link would change the estimand.
- `SyntheticDifferenceInDifferencesWeightFitter`, `PropensityScore`, `InstrumentalVariableRegression`: different graph structures and estimands.
- `BayesianBasisExpansionTimeSeries`, `StateSpaceTimeSeries`: Gaussian observation models with component-specific structure; separate follow-up if non-Gaussian observation models are needed.
- `PyMCForecastModel` / `pymc-forecast` backend: link must be applied inside the user `model_fn`; coordinate via docs and upstream issues (`pymc-forecast#52`, `#50`).
- scikit-learn backends: no `eta` / `mu` / `y_hat` contract; separate design if sklearn GLM support is ever desired.
- Public custom `Family` / `Link` objects, probit / cloglog / softplus links, and zero-inflated families.
- Bambi or statsmodels as dependencies.

## Acceptance criteria

- [ ] `GeneralizedLinearRegression(family="poisson")` builds a graph with `eta`, inverse-linked outcome-scale `mu`, and Poisson `y_hat`.
- [ ] `GeneralizedLinearRegression(family="bernoulli")` exposes probability-scale `mu` and Bernoulli `y_hat` via `p`.
- [ ] `LinearRegression` behavior is unchanged for existing callers (identity Gaussian default).
- [ ] `predict()` returns both `mu` and `y_hat` with dims `(chain, draw, obs_ind, treated_units)`.
- [ ] `calculate_impact()` produces response-scale contrasts (`y - mu`), verified by tests for each curated family.
- [ ] Invalid family / link combinations raise clear errors at model construction.
- [ ] `_clone()` preserves `family` and `link` for sensitivity-check refits.
- [ ] `InterruptedTimeSeries` integration test with a simulated count outcome recovers known effects on the count scale.
- [ ] Documentation updated; `prek run --all-files` passes.

## Related

- #1016 — prediction contract documentation
- #1020 — prediction contract implementation (merged)
- #1017 — customer-support ticket counts ITS tutorial (blocked on this issue)
- `causalpy/tests/test_prediction_contract.py` — semantic ground truth for `eta`, `mu`, and `y_hat`
- `docs/source/knowledgebase/prediction-contract.md` — author-facing contract

## Suggested implementation order

1. Internal `inverse_link` helper + family / link validation module.
2. `GeneralizedLinearRegression.build_model()` for Poisson (proves the full contract path).
3. Refactor `LinearRegression` to share the regression head (identity Gaussian).
4. Add NegativeBinomial and Bernoulli branches.
5. Integration test via `InterruptedTimeSeries`.
6. Documentation and export wiring.
7. Resume #1017 notebook work as the first end-to-end consumer.

drbenvincent:
## Summary

Closes #1022 

- Adds `GeneralizedLinearRegression` with curated families (Gaussian, Poisson, Negative Binomial, Bernoulli) and canonical links, exposing response-scale `mu` for the prediction contract (#1016).
- Refactors `LinearRegression` as a Gaussian/identity specialization of the new GLM backend.
- Routes non-identity DiD and PrePostNEGD ATT through g-computation on treated units; RD/RK plots skip R² when `score()` is unavailable.
- Documents the GLM API in the prediction contract and estimands knowledgebase pages.

## Test plan

- [x] `prek run --all-files`
- [x] `make test-patch-cov` (1252 passed, 96% patch coverage)
- [x] CI green on this PR


drbenvincent (commented):
Reviewer: **GPT-5.6 Sol High**

The core `eta → inverse link → response-scale mu → y_hat` design is sound, and keeping `LinearRegression` as a thin Gaussian/identity specialization is the right compatibility choice. However, I recommend changes before merge because the estimator dispatch and test evidence do not yet support the breadth claimed by this major feature.

## Must-fix

### 1. Use g-computation consistently for Bayesian causal estimands

In `causalpy/pymc_models.py::model_uses_identity_link`, models without an explicit capability flag default to identity-link behavior. `DifferenceInDifferences.algorithm` and `PrePostNEGD.algorithm` therefore send contract-compliant custom linked models down the coefficient path, producing a link-scale parameter rather than a response-scale causal effect. The identity branch in PrePostNEGD is also not generally an ATT when group–covariate interactions are present.

The simpler and safer rule is estimand-based: all Bayesian DiD and PrePostNEGD models already promise response-scale `mu`, so always contrast factual and counterfactual `mu` draws. Keep coefficient extraction only for OLS. This also removes `model_uses_identity_link` and makes Gaussian additive models an equivalent special case.

### 2. Add tests that establish causal correctness, not only execution

`causalpy/tests/test_glm.py::test_poisson_its_recovers_simulated_effect` uses mocked sampling and only checks that cumulative impact is finite; it never compares the estimate with the known effect. The DiD and PrePostNEGD tests assert dimensions but not direction or magnitude. RD, RK, PiecewiseITS, and Staggered DiD are primarily smoke tests.

There is also a concrete false-positive in `test_poisson_piecewise_its_smoke`: `t` is normalized to `[0, 1]`, while the formula keeps `step(t, 40)` and `ramp(t, 40)`, so the interruption terms are always zero.

Please add deterministic numerical assertions for the g-computation contrasts and one lightweight unmocked ITS recovery test whose posterior interval contains the known daily/cumulative truth. The PiecewiseITS interruption must lie inside the modeled time range.

### 3. Preserve the existing `default_priors` extension point

`GeneralizedLinearRegression.__init__` calls `PyMCModel.__init__`, which assembles priors from `self.default_priors`, and then immediately replaces `self.priors` with `_default_priors_for_family(...)`. Existing `LinearRegression` subclasses that override `default_priors` are therefore silently ignored. Please retain one authoritative prior source while preserving this released subclass contract.

## Should-fix before the public API freezes

- `PrePostNEGD._glr_att_from_g_computation` performs two posterior-predictive calls per treated row and also samples unused `y_hat`. Build treated/control design matrices in bulk and make two prediction calls total.
- `ARCHITECTURE.md` says GLR can be used in every `LinearRegression` experiment, including `PanelRegression`. Non-Gaussian outcomes are invalid with `fe_method="demeaned"`, because demeaning creates negative/non-integral responses. Reject that combination or narrow the support claim to dummy fixed effects.
- The public `link` argument currently adds no capability because every non-canonical pairing is rejected. For v1, `family` alone with a read-only resolved link is a smaller API.
- A custom `priors["y_hat"]` can contradict the declared family without validation. Either validate the distribution or reserve this override for the Gaussian compatibility path.
- Document the identification scale for nonlinear DiD: Poisson/log imposes untreated parallel trends on the log-mean scale, while Bernoulli/logit imposes it on the log-odds scale, even though the reported ATT is transformed to an additive response-scale contrast.
- Clarify that the current DiD ATT weights treated-post rows, not treated units, and reconcile the detailed estimand text with the quick-reference table, which still says DiD is coefficient-based.
- Remove or qualify “rate” support until there is an exposure/offset API. The current implementation supports counts conditional on covariates, not a fixed-exposure rate model.

## What worked well

- `mu` is inverse-linked draw by draw before effects are computed.
- The family likelihood heads and Negative Binomial `alpha` customization are compact and understandable.
- `_clone()` preserves family, link, priors, and sampling configuration.
- Returning `None` rather than a misleading non-Gaussian Bayesian R² is conservative.
- Keeping `LinearRegression` preserves imports, default experiment wiring, type identity, and the common Gaussian path.

If these issues are resolved, this genuinely unlocks count and binary outcomes across ITS, Piecewise ITS, DiD, PrePostNEGD, RD, RKD, Staggered DiD, and appropriately configured panel regressions, with effects reported in counts or probability points rather than link-scale coefficients.

**Merge readiness:** changes recommended. The branch is current and mergeable, and remote CI is green; the blockers are statistical-contract and public-API issues rather than mechanical failures.

drbenvincent:
Thanks @drbenvincent — pushed `31c7d4c0` addressing the review. Summary of what changed, how, and why, plus what I deliberately left for later.

## Must-fix

### 1. Uniform Bayesian g-computation for DiD / PrePostNEGD
**What:** Removed `model_uses_identity_link` and the identity/coefficient dispatch entirely. Bayesian `DifferenceInDifferences` and `PrePostNEGD` now always compute ATT from factual minus counterfactual response-scale `mu` draws. OLS DiD still uses the interaction coefficient.
**How:** `_att_from_g_computation()` in both experiments calls `predict(..., var_names=["mu"])` so we do not sample unused `y_hat` on the estimand path. PrePost builds treated/control design matrices in bulk (two prediction calls total, not per-row).
**Why:** Every compliant `PyMCModel` already exposes outcome-scale `mu`; estimand logic should not guess link behaviour from model type. This also fixes custom linked models and PrePost cases with group×covariate interactions where a main-effect coefficient is not the ATT.

### 2. Correctness tests (not just smoke)
**What:**
- Deterministic `FixedMuRegression` test double with fixed log-link `mu`; exact DiD and PrePost ATT assertions, including a group×`pre` interaction where coefficient extraction would be wrong. PrePost test records that exactly two bulk `mu`-only predictions are used for the effect.
- Fixed `test_poisson_piecewise_its_smoke`: interruption threshold now matches normalized `t` (step/ramp columns asserted nonzero).
- New `test_glm_recovery.py::test_poisson_its_posterior_recovers_level_shift` uses real MCMC via a function-scoped `real_pymc_sample` fixture (session prior mock preserved elsewhere). Deterministic count DGP on log-mean scale; HDI checks for daily and cumulative impact.
**Why:** Prior-sampling mocks are fine for plumbing but cannot validate estimands. One isolated real-posterior test gives genuine coverage without slowing the whole suite.

### 3. `default_priors` subclass contract
**What:** `GeneralizedLinearRegression.default_priors` is now a property merging `_default_priors_for_family(family)` with subclass `default_priors` overrides in the MRO. `PyMCModel.__init__` remains the single merge point (`defaults → user priors`); we no longer overwrite after `super().__init__`.
**Why:** Restores the released extension point for `LinearRegression` subclasses without a second prior assembly path.

## Should-fix (also done in this push)

| Item | Change |
|------|--------|
| PrePost vectorization | Two bulk `mu` predictions over all treated rows |
| Panel + demeaned + non-Gaussian GLR | Construction-time `ValueError`; docs/ARCHITECTURE qualified |
| Public `link` arg | Removed; `family` only, read-only `link` property |
| Custom `priors["y_hat"]` on non-Gaussian families | Rejected at construction with a clear error |
| Nonlinear identification scale | Documented in `estimands.md` (log-mean / log-odds parallel trends; reported effects remain response-scale contrasts) |
| DiD weighting + quick-reference | Documented row-weighted treated-post averaging; table distinguishes OLS vs Bayesian |
| “Rate” claims | Removed from `prediction-contract.md`; count means only until offset/exposure API exists |
| `custom_pymc_models.ipynb` | Likelihood override narrowed to Gaussian `LinearRegression`; count/binary → `GeneralizedLinearRegression` |

## Validation
- Targeted GLM / recovery / panel / input-validation tests pass (mocked + real sampler in one run).
- `make doctest`, `make html`, `make test-patch-cov` (97% patch coverage), `prek run --all-files` all green locally.

## Not fixed yet (intentionally deferred)
- **Deeper correctness tests for RD, RK, Staggered DiD** — still smoke/plumbing coverage; same pattern as the new DiD/PrePost deterministic tests could be added in a follow-up if useful.
- **Offset / exposure (“rate”) models** — out of scope for this PR; docs no longer claim rate support.
- **Stale API autodoc stub** for removed `model_uses_identity_link` may linger until docs are regenerated on RTD (harmless warning locally).

Happy to iterate on any of the deferred items or tighten the recovery test budget if CI runtime is a concern.

drbenvincent (commented):
Reviewer: **GPT-5.6 Sol High**, reviewing on behalf of @drbenvincent.

I support the core direction here. A curated GLR with canonical families, draw-wise inverse linking, response-scale `mu`, and `LinearRegression` retained as the Gaussian specialization is the right basic design. The second commit also materially improved the DiD/PrePost estimands.

I still think this needs changes before merge, primarily to make the model/experiment boundary principled and to fix two concrete public-API bugs.

## Must-fix

### 1. Subclass prior overrides still drop required family priors

In `causalpy/pymc_models.py:382,784-794`, the new `default_priors` property is shadowed whenever a subclass defines the existing class-level `default_priors` extension point. For `_CustomPoissonGLR`, `PyMCModel.__init__` therefore sees only the subclass’s `beta`; the family’s `y_hat` is absent and fitting fails with `KeyError: "y_hat"`.

The tests at `causalpy/tests/test_glm.py:252-274` do not catch this: they inspect the override or invoke the base descriptor manually, but never assert that `model.priors` retains `y_hat` and never fit the subclass.

Please use one non-shadowable prior-merging path and add a subclass fit test. The implementation should not require callers to invoke a base descriptor manually.

### 2. `predict()` breaks for natural integer count/binary outcomes

`PyMCModel._data_setter()` (`causalpy/pymc_models.py:456-458`) replaces `y` with a float64 zero array. If a Poisson, Negative Binomial, or Bernoulli model was fit using an integer `y`, PyMC registered integer mutable data and prediction then fails with a dtype error.

Please preserve the fitted data node’s dtype when creating the prediction placeholder and add an integer-outcome `fit()` → `predict(var_names=["mu"])` regression test.

The GLR should also validate response support at the fit boundary: Bernoulli values must be 0/1; Poisson/NB values must be finite, non-negative, and integer-like. Otherwise users receive opaque PyMC sampling errors.

### 3. State the estimand boundary independently of backend type

The current implementation is moving in the right direction, but “Bayesian versus OLS” is being used as a proxy for estimand computation in `DifferenceInDifferences.algorithm()` (`diff_in_diff.py:223-235`). That is not the durable boundary: sklearn also supplies outcome-scale expected predictions, while some Bayesian models do not satisfy the response-scale `mu` contract.

Please make the ownership rule explicit:

- The model/backend owns `E[Y | X, θ]` on the response scale and the uncertainty representation.
- The experiment owns the intervention/counterfactual construction, evaluation population, and aggregation.
- “Counterfactual prediction/contrast” is the broad computational primitive. G-computation is the treatment-standardization subset, not a synonym for every prediction-based estimator.

For DiD, retaining the OLS interaction coefficient is fine as an algebraic shortcut under the currently validated additive design, but document and test its equivalence to the experiment’s ATT rather than defining the estimand by backend.

A compact matrix in `ARCHITECTURE.md` (probably not `estimands.md` because #1033 is moving specifics out of docs into docstrings, to avoid drift between code and docs) should identify, for each experiment, its counterfactual operation and required model capability:

- observed minus counterfactual expectation: ITS, Synthetic Control, Staggered DiD;
- factual minus counterfactual expected-outcome contrast: DiD, PrePostNEGD;
- local expected-outcome geometry: RD and RK;
- specialized/non-portable estimators: IV, IPW, SDiD;
- PanelRegression: model fitting and link-scale coefficient inference only until it has an explicit treatment contrast and target-population API.

I would not add a generic `g_computation.py` abstraction in this PR: the interventions are genuinely experiment-specific. Establish the semantic contract first; extract a helper later only if repeated code has the same semantics.

### 4. Correct the claimed model/experiment compatibility

The committed `custom_pymc_models.ipynb` says implementing one `build_model()` makes a custom model work with all CausalPy experiment classes, plots, summaries, and diagnostics. That is false for experiments with specialized model signatures or estimators (IV, IPW, SDiD), and optional methods such as scoring/printing are not uniformly supported.

Please narrow this to the standard regression experiments that actually satisfy the contract and connect it to the support/estimand matrix above. Similarly, describe Panel + non-Gaussian GLR as coefficient-level support, not a completed response-scale causal estimand.

## Required behavioral evidence

Please keep the existing deterministic intervention tests, but add the smallest tests that pin the actual contracts:

1. A GLR subclass with partial default priors can fit and retains the family likelihood.
2. Integer Poisson/Bernoulli data can fit and predict.
3. Standard Gaussian DiD and PrePostNEGD remain equivalent to their former coefficient estimands where that equivalence is expected.
4. At least one built-in GLR experiment test asserts a numerical expected-outcome contrast, rather than only type/dimension checks.
5. The Poisson panel test uses actual non-negative integer count data; it currently reuses a continuous Gaussian fixture containing invalid Poisson outcomes under mocked sampling.

## Should-fix

- `DifferenceInDifferences` plots collapse treated rows with `groupby(time).first()` while the reported ATT averages all treated-post rows (`diff_in_diff.py:197-209` versus `310-336`). Either align the populations or label the plot as one representative trajectory.
- Decide whether score-less models are unsupported in Synthetic Control and reject them early, or handle `score is None` in `synthetic_control.py:919-932`. Patching individual plots after they fail is evidence that backend kind is too coarse a capability boundary.
- Document that Regression Kink with a nonlinear link reports a change in the slope of expected counts/probabilities, not a link-scale coefficient.
- Split commits `16eb78bb` and `03e25b80` (Codecov/Makefile/maintainer workflow policy) into a separate PR. They are repo-wide process changes unrelated to the GLR feature and complicate review and revert.

## What is already strong

- `eta → inverse link → mu → y_hat` is the correct prediction contract.
- The family-only public API is appropriately small; this does not need a Bambi dependency.
- Bayesian DiD and PrePostNEGD now contrast response-scale expectations, including interaction-aware PrePost effects.
- The new ATT calculations are vectorized and request only `mu`.
- The docs correctly distinguish link-scale identification assumptions from response-scale reported effects.

Merge readiness: changes requested. The branch is current, mergeable, and CI is green; the remaining blockers are API correctness and the conceptual support contract, not mechanical CI failures.

drbenvincent:
Pushed cca16beb and 36eb87e6 addressing the second review.

## Must-fix

1. **Subclass prior overrides**: removed the shadowable `default_priors` property. `GeneralizedLinearRegression.__init__` now merges the family's required priors *underneath* whatever `PyMCModel.__init__` assembled, so a partial class-level `default_priors` override can never drop the family `y_hat`. No base-descriptor invocation needed. New tests: `test_glr_subclass_partial_default_priors_retain_family_priors` and `test_glr_subclass_partial_default_priors_can_fit` (the subclass actually fits and keeps the Poisson likelihood).
2. **Integer outcomes**: `_data_setter` now creates the prediction placeholder with the fitted `y` node's dtype. `GeneralizedLinearRegression.fit` validates response support at the boundary (Bernoulli 0/1; Poisson/NB finite, non-negative, integer-valued) with clear `ValueError`s. New tests: `test_glr_integer_outcome_fit_and_predict` (int64 fit → `predict(var_names=["mu"])`) and `test_glr_fit_rejects_out_of_support_outcomes`.
3. **Estimand boundary**: added an estimand/capability matrix to `ARCHITECTURE.md` stating the ownership rule (experiment owns intervention, population, aggregation; model owns response-scale `E[Y|X,θ]` and uncertainty) and classifying each experiment's counterfactual operation, including PanelRegression as coefficient-inference-only. `DifferenceInDifferences.algorithm` now documents the OLS interaction coefficient as a tested algebraic shortcut rather than a backend-defined estimand, pinned by `test_ols_did_coefficient_equals_prediction_contrast`. No generic `g_computation.py` abstraction added, per the review.
4. **Compatibility claims**: `custom_pymc_models.ipynb` now scopes custom-model support to the standard regression experiments, points at the ARCHITECTURE matrix, names IV/IPW/SDiD as requiring dedicated model classes, and describes Panel + non-Gaussian GLR as coefficient-level support.

## Behavioral evidence

1. Subclass with partial default priors fits and retains the family likelihood ✔ (see must-fix 1)
2. Integer Poisson/Bernoulli fit and predict ✔ (see must-fix 2)
3. `test_gaussian_did_att_matches_interaction_coefficient` and `test_gaussian_prepostnegd_att_matches_group_coefficient` assert draw-wise equality of the g-computation ATT with the former coefficient estimands under identity link.
4. `test_poisson_glr_did_att_matches_posterior_beta_contrast` reconstructs `mean(exp(X_f·β) − exp(X_c·β))` from the built-in GLR's own posterior draws and asserts numerical equality with `causal_impact`.
5. `test_panel_regression_poisson_dummies` now generates genuine Poisson counts; the old continuous fixture is rejected by the new fit-boundary validation.

## Should-fix

- DiD plot vs ATT population: documented in the `plot()` docstring that trajectories use one representative covariate row per time point while the reported ATT averages all treated-post rows.
- Synthetic Control now handles `score is None` gracefully in `_get_score_title` instead of crashing, consistent with the RD/RK handling.
- Regression Kink docstring now states that `gradient_change` with a non-identity link is a change in the slope of expected counts/probabilities, not a link-scale coefficient.
- Commits `16eb78bb`/`03e25b80` process changes split out to #1034 and reverted here (the GLR tests from `16eb78bb` stay in this PR).

`prek run --all-files` passes and `make test-patch-cov` reports 98% patch coverage (1273 passed).

drbenvincent:
## Status summary

For clarity on where this PR stands after commits `cca16beb` and `36eb87e6` (details in the comment above):

**Resolved — all blockers from the second review:**

| Review item | Status | How |
|---|---|---|
| Must-fix 1: subclass `default_priors` drops family `y_hat` | ✅ Fixed | Shadowable property removed; family priors merged non-shadowably underneath in `__init__`; subclass now fitted in tests |
| Must-fix 2: `predict()` breaks on integer outcomes | ✅ Fixed | `_data_setter` preserves the fitted `y` dtype; fit boundary validates family support (0/1, non-negative integer counts) |
| Must-fix 3: estimand defined by backend type | ✅ Fixed | Ownership rule + estimand/capability matrix in `ARCHITECTURE.md`; OLS coefficient documented and tested as an algebraic shortcut; no generic g-computation abstraction added |
| Must-fix 4: overstated custom-model compatibility | ✅ Fixed | `custom_pymc_models.ipynb` scoped to standard regression experiments; IV/IPW/SDiD and Panel caveats stated |
| Behavioral evidence 1–5 | ✅ Added | Subclass fit, integer fit/predict, Gaussian DiD/PrePostNEGD draw-wise equivalence, built-in GLR numeric contrast, Poisson panel with real counts |
| Should-fixes (DiD plot label, SC `score=None`, RK link docs, commit split) | ✅ Done | Process changes extracted to #1034 and reverted here |

**Verification:** `prek run --all-files` clean, `make test-patch-cov` at 98% patch coverage (1273 passed), all CI checks green.

**Remaining, not blocking merge of this PR:**

- Human review of the docs that now state project policy: the estimand/capability matrix in `ARCHITECTURE.md`, the narrowed claims in `custom_pymc_models.ipynb`, and the estimands.md wording.
- #1034 (patch-coverage process changes split from this PR) merges independently.
- The reworked g-computation tutorial notebook (#1021/#1017) is deliberately **not** in this PR; it is parked locally as separate follow-up work.

No known open issues from either review round remain unaddressed in code.

drbenvincent:
## MERGE ORDER

* Ideally we merge #1033 @ErikRingen / @NathanielF soon
* Then we can update docstrings as appropriate in this PR and merge

drbenvincent:
Heads-up for whoever updates this branch from `main`: #1036 is about to merge and it reshapes the prediction boundary this PR builds on. Guidance for the rebase:

- **`ModelAdapter.predict()` is now the single prediction method.** It returns response-scale `mu` as an `xr.DataArray` with canonical dims `(chain, draw, obs_ind, treated_units)` on every backend (sklearn returns singleton `chain`/`draw`). `predict_mu()` no longer exists — if a conflict resolution reintroduces it, that's a wrong merge.
- **`calculate_impact()` / `calculate_cumulative_impact()` were deleted** from `PyMCModel`, `ScikitLearnAdaptor`, and `PyMCForecastModel`. Experiments now compute impact as plain xarray subtraction `y - predict(X)` and `.cumsum(dim="obs_ind")`. Don't resurrect the methods when resolving conflicts in the model classes.
- **Nothing about GLM semantics needs to change here.** This PR already registers `mu = inverse_link(eta)` on the response scale, which is exactly what the adapter's `predict()` extracts (via `_extract_mu` in `model_adapter.py`), so impact/effect summaries stay in outcome units automatically. The g-computation ATT paths should consume `self._model_backend.predict(...)` and get `mu` draws directly — no `posterior_predictive["mu"]` digging needed.
- **`reporting.py` now assumes canonical containers.** `_extract_window` / `_extract_counterfactual` no longer duck-type across InferenceData/dict/ndarray; the duplicated Bayesian/OLS `effect_summary` bodies in ITS/SC/PiecewiseITS collapsed onto `_effect_summary_timeseries`. If this branch touched those, prefer the `main` side and re-apply the GLM-specific bits on top.
- If this PR ends up needing link-scale `eta` or noise-inclusive `y_hat` downstream of the adapter, the agreed design is a small `var_name="mu"` parameter on `predict()` at the `_extract_mu` seam — not a new family of `predict_*` methods.

The contract is pinned by `causalpy/tests/test_prediction_contract.py` (now asserted through the adapter), so a correct merge should leave those tests green.

drbenvincent:
## TODO (note to self)

- [x] Confirm the slow CI jobs from the main-sync push went green: test (3.11/3.14), pymc + mixed notebook jobs, Read the Docs — all green as of 2026-07-22 15:55 UTC+1
- [ ] Human read-through of the policy-stating docs: estimand/capability matrix in `ARCHITECTURE.md`, narrowed claims in `custom_pymc_models.ipynb`, `estimands.md` wording, and the updated authoring checklist in `prediction-contract.md`
- [ ] Decide on merge order vs #1033: either merge #1033 first and align docstrings here, or drop that ordering and merge this directly
- [ ] Approve + merge (no unresolved review threads; no known code work left)

Context: branch was synced with main on 2026-07-22 — resolved conflicts against the unified `_plot` refactor (#1053) and canonical prediction container (#1036), non-Gaussian GLM plots now fall back gracefully when `score()` is skipped. Full suite green locally, 99% patch coverage.

ErikRingen:
Am I right in my reading that one has to set a single prior on all coefficients, so same for intercept and slopes? This PR made me think about the trickiness of setting good default priors for GLMs. Whether N(0,1) is a good intercept prior seems entirely dependent on the magnitude of the count, with the current prior assuming a range of only  [0.14, 7.4]. If you had a larger count you would want something bigger, but then you wouldn't also want to scale up the predictor betas because it is multiplicative.

So two suggestions:

(1) allow different spec of intercept and slope priors. PyMC marketing does this for loglink models: https://github.com/pymc-labs/pymc-marketing/blob/07b1a2046bdb7e0397cd8e488145553eed32ee28/pymc_marketing/mmm/link.py#L240

(2) add the sample mean as an offset for count models, so that a generic prior makes sense regardless of count magnitude, i.e., log(mu) = log(y_bar) + b0 + b)

b0 ~ N(0,1)

ErikRingen:
For the `_plot()` method, we might want to add a parameter to plot on the response or link scale. For example, given that log-link diff-in-diff makes the parallel assumptions trend on the log scale, that would be the appropriate scale to visually interrogate that assumption.

PR #1025, opened 2026-07-20
Fix biased SE denominator in OLS DiD effect_summary (SSR/n -> SSR/(n-p))

Mari1988:
## Summary
- `_compute_statistics_did_ols` (used by `DifferenceInDifferences.effect_summary()` for OLS-fitted models) estimated the residual variance as `mse = np.mean(residuals**2)` (divides by `n`), while the degrees of freedom used a few lines later for the t-distribution critical value is `df = n - p`.
- The standard unbiased OLS variance estimator is `SSR / (n - p)`, not `SSR / n`, so the two were internally inconsistent and the reported SE/CI were biased downward (understated) by a factor of `sqrt((n-p)/n)`.
- Fixes `mse` to `np.sum(residuals**2) / df`, matching the `df` already used for the t-critical value.

Note: a separate, more serious bug in how the residuals themselves are computed (an `(n, n)` broadcasting issue, ~70x SE inflation) was found during this investigation and is fixed independently in #1026 so each fix stays small and reviewable. The regression test added here deliberately isolates the denominator-convention change from that separate issue.

## Test plan
- [x] Added `test_effect_summary_ols_did_se_uses_unbiased_denominator` in `causalpy/tests/test_reporting.py`
- [x] `pytest causalpy/tests/test_reporting.py -k did` — 11 passed
- [x] Independently verified the corrected formula against `statsmodels` OLS (matches to 1e-10 once residuals are correctly shaped)

### DIFF
--- a/causalpy/reporting.py
+++ b/causalpy/reporting.py
@@ -1533,10 +1533,12 @@ def _compute_statistics_did_ols(
 
     # Calculate standard error from model residuals
     residuals = _point_residuals(result)
-    mse = np.mean(residuals**2)
     X_da = result.design["X"]
     n, p = X_da.shape
     df = n - p
+    # Unbiased estimator of the residual variance: SSR / (n - p), consistent
+    # with the degrees of freedom used below for the t-distribution.
+    mse = np.sum(residuals**2) / df
 
     # Find the interaction term coefficient index
     interaction_term = (
--- a/causalpy/tests/test_reporting.py
+++ b/causalpy/tests/test_reporting.py
@@ -335,17 +335,19 @@ def test_effect_summary_ols_did(mock_pymc_sample, did_data):
 @pytest.mark.integration
 def test_effect_summary_ols_did_residuals_are_per_observation(did_data):
     """``_compute_statistics_did_ols`` must compute one residual per
-    observation, not an (n, n) array.
+    observation, not an (n, n) array, and use the unbiased SSR/(n-p)
+    estimator of the residual variance (matching the ``df = n - p`` already
+    used for the t-distribution critical value).
 
     ``y_da`` (shape ``(n, 1)``, dims ``obs_ind x treated_units``) minus a bare
     ``(n,)`` ``y_pred`` array used to broadcast positionally against the
     *last* axis (``treated_units``, size 1) instead of ``obs_ind``, producing
     an ``(n, n)`` array of every observation's y minus every *other*
-    observation's prediction. That inflated the reported SE by roughly 70x
-    on this dataset. This test checks the residual shape directly and checks
-    the reported SE against an independent statsmodels fit, up to the
-    still-present (separately tracked) mean/(n-p) denominator convention
-    difference, which is out of scope for this fix.
+    observation's prediction; that inflated the reported SE by roughly 70x
+    on this dataset. Separately, the residual variance was estimated as
+    SSR/n (biased) rather than SSR/(n-p) (unbiased). With both bugs fixed,
+    the reported SE/CI should match an independent statsmodels OLS fit
+    almost exactly, not just up to some remaining conversion factor.
     """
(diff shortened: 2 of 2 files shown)


drbenvincent:
## Automated triage

**Why:**
- Fixes a numerical bug in `causalpy/reporting.py` (`_compute_statistics_did_ols`): SE denominator was `SSR/n` instead of `SSR/(n-p)`, understating reported SE.
- Changes numerical output of the public `DifferenceInDifferences.effect_summary()` method.
- Adds ~59 lines of regression tests. All CI checks pass (100% patch coverage).

**Review focus:**
1. Verify the fix (SSR/(n-p) with the df already used for t-critical values) is correct.
2. Check that the regression test is robust and isolates the denominator fix from the residual-shape bug (handled in #1026).
3. Coordinate merge with sibling PRs #1026 (DiD residual fix) and #1027/#1028 (RD equivalents).

**Confidence:** high

Mari1988:
Rebased onto `main` now that #1026 merged (that's why the diff/conflict changed).

Since #1026 already fixed the residual-shape bug, the two remaining changes here are internally consistent again:
- `mse` now uses `sum(residuals**2) / df` instead of `mean(residuals**2)`, matching the `df = n - p` already used for the t-critical value.
- Updated `test_effect_summary_ols_did_residuals_are_per_observation` (already on `main` from #1026): with both bugs now fixed, the reported SE/CI/p-value match an independent `statsmodels` OLS fit to ~1e-16, not just up to the previously-noted `sqrt((n-p)/n)` gap. Strengthened the assertion accordingly and kept a regression guard against the old biased-denominator SE.

Full suite (1225 tests) passes, 100% patch coverage against the new `main`.

drbenvincent:
Hi @Mari1988. Can you update from `main` (for all these PRs), we just merged a refactor, so we want to make sure everything is up to date.

Mari1988:
Updated from main. This picked up the #1057 refactor (\`_point_residuals()\`), which already handles the residual-shape fix for both DiD and RD — so this PR's diff is now purely the denominator fix (\`mse = sum(residuals**2)/df\` instead of \`mean\`), adapted to call \`_point_residuals()\` for the residuals themselves rather than computing them inline.

Full suite: 1233 passed, 5 skipped. Patch coverage: 100%.

Mari1988:
Hi @drbenvincent , I think this is ready to be merged whenever you can. Thank you.

drbenvincent:
Thanks for the reminder @Mari1988. Just need to find some hours in the day for open source 😂. Very soon hopefully 🤞

daimon-pymclabs (commented):
## Review — looks correct, no blockers

Verified the fix against the OLS variance estimator. `mse = np.sum(residuals**2) / df` with `df = n - p` is the standard unbiased residual-variance estimate, and `se = sqrt(mse * (XᵀX)⁻¹_jj)` is then exactly the textbook coefficient SE. This restores internal consistency with the `df = n - p` already used for the t-critical value, and the diff correctly sources the residuals from `_point_residuals()` (post-#1057), so the earlier `(n, n)` broadcasting issue can't creep back in.

**Test:** the strengthened `test_effect_summary_ols_did_residuals_are_per_observation` is a good regression guard — it now asserts SE/CI match an independent statsmodels OLS fit to `rel=1e-8` and keeps an explicit guard against regressing to the biased `SSR/n` denominator (and the old ~70x shape bug). Clean.

### One coordination note (non-blocking)
The identical `mse = np.mean(residuals**2)` pattern still lives in `_compute_statistics_rd_ols` (`causalpy/reporting.py:1667`). That's the RD twin of this bug, and it's tracked in the still-open #1027 — just flagging so #1027 lands before the next release, otherwise DiD and RD will report SEs on different conventions.

**Recommendation:** merge. Correct, minimal, well-tested.

drbenvincent (approved):

PR #1026, opened 2026-07-20
Fix residual-shape broadcasting bug in OLS DiD effect_summary (~70x SE inflation)

Mari1988:
## Summary
- `_compute_statistics_did_ols` computed `residuals = y_da - y_pred`, where `y_da` is an xarray `DataArray` of shape `(n, 1)` with dims `(obs_ind, treated_units)` and `y_pred` is a bare `(n,)` numpy array with no dimension names.
- xarray aligned `y_pred` positionally against the *last* axis (`treated_units`, size 1) instead of `obs_ind`, producing an `(n, n)` array of every observation's `y` minus every *other* observation's prediction instead of `n` proper residuals.
- On the built-in `did` dataset this inflated the reported SE by roughly 70x (SE ≈ 4.16 instead of ≈ 0.06), making the reported CI `[-7.98, 8.89]` instead of the correct `[0.34, 0.58]` — not a rounding nuance, essentially uninformative.
- Fixes it by flattening both operands to 1-D before differencing so they align by position on `obs_ind`.

This is independent of #1025 (which fixes a separate denominator-convention bug a couple of lines below) — both touch adjacent but non-overlapping lines and can merge in either order without conflict. This PR deliberately leaves the `mean`-based denominator as-is; #1025 handles that.

## Test plan
- [x] Added `test_effect_summary_ols_did_residuals_are_per_observation` in `causalpy/tests/test_reporting.py`, asserting residuals have shape `(n,)` and the reported SE matches the expected value up to the still-present (separately tracked) denominator convention
- [x] `pytest causalpy/tests/test_reporting.py -k did` — 11 passed
- [x] Independently verified against `statsmodels` OLS: reported SE now matches `unbiased_se * sqrt((n-p)/n)` exactly, confirming the `(n,n)` broadcast is gone

### DIFF
--- a/causalpy/reporting.py
+++ b/causalpy/reporting.py
@@ -1494,7 +1494,14 @@ def _compute_statistics_did_ols(
     X_da = result.design["X"]
     y_da = result.design["y"]
     y_pred = result.model.predict(X_da)
-    residuals = y_da - y_pred
+    # y_da has shape (n, 1) with dims (obs_ind, treated_units); y_pred is a
+    # bare (n,) array with no dimension names. Subtracting them directly lets
+    # xarray align y_pred positionally against the *last* axis (treated_units,
+    # size 1) instead of obs_ind, producing an (n, n) array of every
+    # observation's y minus every other observation's prediction instead of
+    # per-observation residuals. Flatten both to 1-D first so they align by
+    # position on obs_ind as intended.
+    residuals = np.asarray(y_da).reshape(-1) - np.asarray(y_pred).reshape(-1)
     mse = np.mean(residuals**2)
     n, p = X_da.shape
     df = n - p
--- a/causalpy/tests/test_reporting.py
+++ b/causalpy/tests/test_reporting.py
@@ -333,6 +333,67 @@ def test_effect_summary_ols_did(mock_pymc_sample, did_data):
 
 
 @pytest.mark.integration
+def test_effect_summary_ols_did_residuals_are_per_observation(did_data):
+    """``_compute_statistics_did_ols`` must compute one residual per
+    observation, not an (n, n) array.
+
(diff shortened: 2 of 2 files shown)


drbenvincent:
## Automated triage

**Why:**
- Fixes a numerical bug in `causalpy/reporting.py` (`_compute_statistics_did_ols`) producing `(n, n)` residuals instead of `(n,)` — reported SE inflated ~70x.
- Changes numerical output of the public `DifferenceInDifferences.effect_summary()` method.
- Adds ~61 lines of regression tests. All CI checks pass.

**Review focus:**
1. Confirm the flattening fix correctly resolves the xarray alignment issue (identical root cause to #1028 for RD).
2. Verify the regression test captures the correct before/after SE values.
3. Coordinate merge order with sibling PRs #1025 (DiD denominator fix) and #1028/#1027 (RD equivalents).

**Confidence:** high

drbenvincent (approved):
Thank you for tracing this carefully. I verified the root cause: the canonical single-outcome `y` is shaped `(n, 1)`, while sklearn predicts `(n,)`, so the old subtraction produced an `(n, n)` residual matrix through trailing-axis broadcasting. Flattening both operands before subtraction gives the intended one-residual-per-observation calculation, and the end-to-end comparison with statsmodels guards the public CI behavior. The focused test and the full remote checks pass.

We intend to merge this residual-shape fix before #1025, then return to the denominator correction afterward. One coordination note: the two PRs are not actually merge-order-independent at the test level—this test intentionally pins the current `RSS / n` convention, while #1025 changes it. That does not block this PR given the chosen order; #1025 will need to be rebased and its test updated to assert the combined final behavior after this lands.

The branch is three unrelated maintenance commits behind current `main`, but GitHub reports it mergeable. This is a good, minimal fix for the severe DiD error. Thanks again for the clear diagnosis and regression coverage.

PR #1027, opened 2026-07-20
Fix biased SE denominator in OLS RD effect_summary (SSR/n -> SSR/(n-p))

Mari1988:
## Summary
- `_compute_statistics_rd_ols` (used by `RegressionDiscontinuity.effect_summary()` for OLS-fitted models) estimated the residual variance as `mse = np.mean(residuals**2)` (divides by `n`), while the degrees of freedom used a few lines later for the t-distribution critical value is `df = n - p`.
- The standard unbiased OLS variance estimator is `SSR / (n - p)`, not `SSR / n`, so the two were internally inconsistent and the reported SE/CI were biased downward (understated) by a factor of `sqrt((n-p)/n)`.
- Fixes `mse` to `np.sum(residuals**2) / df`, matching the `df` already used for the t-critical value.
- This is the identical bug and fix already applied to `_compute_statistics_did_ols` (Difference-in-Differences) in #1025 — same root cause, same fix, different experiment class.

Note: a separate, more serious bug in how the residuals themselves are computed (an `(n, n)` broadcasting issue) also exists in this function and is fixed independently in #1028 so each fix stays small and reviewable, mirroring the DiD pair (#1025/#1026). The regression test added here deliberately isolates the denominator-convention change from that separate issue.

## Test plan
- [x] Added `test_effect_summary_ols_rd_se_uses_unbiased_denominator` in `causalpy/tests/test_reporting.py`
- [x] `pytest causalpy/tests/test_reporting.py -k rd` — 5 passed
- [x] Independently verified the corrected formula: reported SE matches `sum(residuals**2)/df` exactly (21.404), not the biased `mean` version (2.097), reproducing the function's residuals computation directly

### DIFF
--- a/causalpy/reporting.py
+++ b/causalpy/reporting.py
@@ -1664,10 +1664,12 @@ def _compute_statistics_rd_ols(result, alpha=0.05):
 
     # Calculate standard error from model residuals
     residuals = _point_residuals(result)
-    mse = np.mean(residuals**2)
     X_da = result.design["X"]
     n, p = X_da.shape
     df = n - p
+    # Unbiased estimator of the residual variance: SSR / (n - p), consistent
+    # with the degrees of freedom used below for the t-distribution.
+    mse = np.sum(residuals**2) / df
 
     # Find the treated coefficient index
     coeff_idx = None
--- a/causalpy/tests/test_reporting.py
+++ b/causalpy/tests/test_reporting.py
@@ -514,7 +514,9 @@ def test_effect_summary_ols_rd_residuals_are_per_observation(rd_data):
     # Guard against regressing to either the (n, n) broadcast bug (~3x
     # inflation) or the biased SSR/n denominator understatement.
     biased_mse = np.mean(residuals**2)
-    XtX_inv = np.linalg.inv(np.asarray(result.design["X"]).T @ np.asarray(result.design["X"]))
+    XtX_inv = np.linalg.inv(
+        np.asarray(result.design["X"]).T @ np.asarray(result.design["X"])
+    )
     coeff_idx = next(
         i
         for i, label in enumerate(result.labels)
--- a/causalpy/tests/test_reporting.py
+++ b/causalpy/tests/test_reporting.py
@@ -474,8 +474,12 @@ def test_effect_summary_ols_rd_residuals_are_per_observation(rd_data):
     ``(n, n)`` residual matrix (every observation's y minus every *other*
     observation's prediction) and inflating the MSE ~9x on this dataset, so
     the reported standard errors were ~3x too wide. ``_point_residuals``
-    fixed that; this test pins the corrected residual shape and SE against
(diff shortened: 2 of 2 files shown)


drbenvincent:
## Automated triage

**Why:**
- Fixes a numerical bug in `causalpy/reporting.py` (`_compute_statistics_rd_ols`): SE denominator was `SSR/n` instead of `SSR/(n-p)`, understating reported SE.
- Changes numerical output of the public `RegressionDiscontinuity.effect_summary()` method.
- Adds ~57 lines of regression tests. CI shows the `test (3.11)` job was cancelled (not failed) — other checks pass.

**Review focus:**
1. Verify the fix is correct (SSR/(n-p) with the df already used for t-critical values).
2. Confirm the cancelled `test (3.11)` job is a transient infrastructure issue, not a real test failure.
3. Coordinate merge with sibling PRs #1028 (RD residual shape) and #1025/#1026 (DiD equivalents).

**Confidence:** high

Mari1988:
Updated from main. Same situation as #1025 — this PR's diff is now purely the RD denominator fix, adapted to call the new \`_point_residuals()\` helper from #1057 instead of computing residuals inline.

Full suite: 1233 passed, 5 skipped. Patch coverage: 100%.

daimon-pymclabs (commented):
Reviewed. Clean, merge it — same fix as #1025, just the RD twin.

**The fix is correct.** `mse = np.sum(residuals**2) / df` with `df = n - p` is the unbiased OLS residual-variance estimate, so `se = sqrt(mse · (XᵀX)⁻¹_jj)` now matches the textbook coefficient SE and is internally consistent with the `df` already feeding the t-critical value. Identical in form to the `_compute_statistics_did_ols` fix already merged in #1025. Residuals come from `_point_residuals()`, so the corrected per-observation residuals are already in place on `main` and the SE lands on statsmodels exactly — the strengthened test asserts `rel=1e-8` and guards against regressing to either the `SSR/n` understatement or the old `(n, n)` broadcast inflation.

**Does it need updating from main?** Not for correctness. The branch is 2 ahead / 1 behind `main` (the 1 being the #1025 merge), but GitHub reports it MERGEABLE — #1025 touched a disjoint function (`_compute_statistics_did_ols`) and a disjoint test (`..._did_residuals_are_per_observation`), so there's no textual or logical overlap with this diff. The `BLOCKED` state is `REVIEW_REQUIRED` (no approval yet), not out-of-date. A sync to re-run CI against post-#1025 `main` is nice-to-have but optional; all checks are already green.

**Merge or concerns?** Merge. No blocking concerns. One doc nit: the PR body says the residual `(n, n)` broadcast bug "also exists in this function" and is deferred to #1028 — that's stale, `_point_residuals()` already fixed it on `main`, which is why this test can assert an exact statsmodels match rather than a conversion-factor approximation. Worth a one-line edit to the description, not a code change.

drbenvincent (approved):

PR #1028, opened 2026-07-20
Fix residual-shape broadcasting bug in OLS RD effect_summary

Mari1988:
## Summary
- `_compute_statistics_rd_ols` computed `residuals = y_da - y_pred`, where `y_da` is an xarray `DataArray` of shape `(n, 1)` with dims `(obs_ind, treated_units)` and `y_pred` is a bare `(n,)` numpy array with no dimension names.
- xarray aligned `y_pred` positionally against the *last* axis (`treated_units`, size 1) instead of `obs_ind`, producing an `(n, n)` array of every observation's `y` minus every *other* observation's prediction instead of `n` proper residuals.
- On the built-in `rd` test dataset this pushed the reported SE well away from the correct value; after this fix the reported SE matches an independent statsmodels fit up to the small, separately-tracked denominator convention difference (ratio `sqrt((n-p)/n)`), rather than being off by an order of magnitude.
- Fixes it by flattening both operands to 1-D before differencing so they align by position on `obs_ind`.
- This is the identical bug and fix already applied to `_compute_statistics_did_ols` (Difference-in-Differences) in #1026 — same root cause, same fix, different experiment class.

Independent of #1027 (which fixes a separate denominator-convention bug a couple of lines below) — both touch adjacent but non-overlapping lines and can merge in either order without conflict. This PR deliberately leaves the `mean`-based denominator as-is; #1027 handles that.

## Test plan
- [x] Added `test_effect_summary_ols_rd_residuals_are_per_observation` in `causalpy/tests/test_reporting.py`, asserting residuals have shape `(n,)` and the reported SE matches the expected value up to the still-present (separately tracked) denominator convention
- [x] `pytest causalpy/tests/test_reporting.py -k rd` — 5 passed
- [x] Independently verified against `statsmodels` OLS: reported SE / unbiased SE ratio = 0.979796, matching `sqrt((n-p)/n) = sqrt(96/100)` to 12+ significant digits, confirming the `(n,n)` broadcast is gone

### DIFF
--- a/causalpy/experiments/regression_discontinuity.py
+++ b/causalpy/experiments/regression_discontinuity.py
@@ -233,6 +233,12 @@ class RegressionDiscontinuity(BaseExperiment):
             }
         )
         (new_x,) = build_design_matrices([self._x_design_info], self.x_discon)
+        # Store the raw design rows used to build the discontinuity contrast
+        # (row 0 = just below the threshold, row 1 = just above) so that
+        # reporting code can compute the standard error of
+        # discontinuity_at_threshold via the same linear contrast
+        # (row 1 - row 0) rather than re-deriving it from coefficient names.
+        self.x_discon_design = np.asarray(new_x)
         self.pred_discon = self._model_backend.predict(X=np.asarray(new_x))
         self.discontinuity_at_threshold = self.pred_discon.isel(
             obs_ind=1, treated_units=0
--- a/causalpy/reporting.py
+++ b/causalpy/reporting.py
@@ -1669,26 +1669,26 @@ def _compute_statistics_rd_ols(result, alpha=0.05):
     n, p = X_da.shape
     df = n - p
 
-    # Find the treated coefficient index
-    coeff_idx = None
-    for i, label in enumerate(result.labels):
-        if "treated" in label.lower() and ":" in label:
-            coeff_idx = i
-            break
-
-    if coeff_idx is None:
+    # discontinuity_at_threshold is not a single coefficient: it is the
+    # difference between the model's predictions just above and just below
+    # the threshold (see RegressionDiscontinuity.algorithm()), i.e. a linear
+    # contrast c @ beta where c is that pair of design rows subtracted. Its
+    # standard error must use the covariance of that same contrast,
+    # sqrt(mse * c @ inv(X'X) @ c), not the variance of any single named
+    # coefficient (e.g. picking out "treated" or "x:treated" alone silently
+    # drops the other coefficients' contributions and their covariances).
+    X = X_da
+    try:
+        if hasattr(X, "values"):
+            X = X.values
+        elif hasattr(X, "data"):
+            X = X.data
+        X = np.asarray(X)
+        XtX_inv = np.linalg.inv(X.T @ X)
+        c = result.x_discon_design[1] - result.x_discon_design[0]
+        se = np.sqrt(mse * c @ XtX_inv @ c)
+    except (np.linalg.LinAlgError, AttributeError):
         se = np.std(residuals) / np.sqrt(n)
-    else:
-        X = X_da
-        try:
-            if hasattr(X, "values"):
-                X = X.values
-            elif hasattr(X, "data"):
-                X = X.data
-            XtX_inv = np.linalg.inv(X.T @ X)
-            se = np.sqrt(mse * XtX_inv[coeff_idx, coeff_idx])
-        except (np.linalg.LinAlgError, AttributeError):
-            se = np.std(residuals) / np.sqrt(n)
 
     # t-critical value
     t_critical = t.ppf(1 - alpha / 2, df=df)
--- a/causalpy/tests/test_reporting.py
+++ b/causalpy/tests/test_reporting.py
@@ -1684,6 +1684,54 @@ def test_compute_statistics_rd_ols_fallback_path(mock_pymc_sample, rd_data):
     assert "p_value" in stats
 
 
+@pytest.mark.integration
+def test_compute_statistics_rd_ols_design_with_data_attribute_only(rd_data):
+    """``_compute_statistics_rd_ols`` must extract the raw array from design
+    matrices that expose only a ``.data`` attribute (no ``.values``), not
+    just the ``xarray``/``pandas``-style objects normally returned by
+    ``result.design["X"]``. Wrap the real design matrix in such a minimal
+    stand-in and check the SE matches the un-wrapped computation exactly,
+    so this isn't just checking that *a* number came out.
+    """
+    from sklearn.linear_model import LinearRegression
+
+    from causalpy.reporting import _compute_statistics_rd_ols
+
+    class _DataOnly:
+        """Minimal array-like exposing `.data`/`.shape` but not `.values`."""
+
+        def __init__(self, array):
+            self._array = np.asarray(array)
+            self.shape = self._array.shape
+
+        @property
+        def data(self):
+            return self._array
+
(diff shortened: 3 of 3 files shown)


drbenvincent:
## Automated triage

**Why:**
- Fixes a numerical bug in `causalpy/reporting.py` (`_compute_statistics_rd_ols`) that was producing `(n, n)` residual arrays instead of `(n,)` — reported SE was inflated by an order of magnitude.
- Changes numerical output behaviour of the public `RegressionDiscontinuity.effect_summary()` method.
- Adds ~60 lines of regression tests in `causalpy/tests/test_reporting.py`. All CI checks pass (including codecov/patch).

**Review focus:**
1. Confirm the flattening fix correctly resolves the xarray alignment issue without introducing edge cases for other callers.
2. Verify the regression test adequately covers the before/after behaviour.
3. Coordinate merge order with sibling PRs #1025, #1026, #1027 (author notes they touch adjacent but non-overlapping lines).

**Confidence:** high

drbenvincent (changes requested):
Thank you for finding and clearly explaining the broadcasting bug. The flattening itself is correct: it changes the RD residuals from `(n, n)` to one residual per observation, and the focused tests and remote CI pass.

I found one blocking correctness issue in the uncertainty calculation that the new regression test currently codifies. In `causalpy/reporting.py`, `_compute_statistics_rd_ols` reports `result.discontinuity_at_threshold` as its mean. That quantity is the difference between predictions immediately above and below the threshold, constructed in `RegressionDiscontinuity.algorithm()`. Its standard error therefore needs to use the covariance of that same linear prediction contrast: `sqrt(mse * c.T @ inv(X.T @ X) @ c)`, where `c` is the above-threshold design row minus the below-threshold design row.

The current implementation and new test instead use only the `x:treated` interaction coefficient's standard error. For the test formula, the contrast includes the treated main effect, the small change in `x`, and the interaction term; it is not the interaction coefficient alone. On the fixture, the PR validates an SE of about 0.695, while the matching threshold-contrast SE is about 0.201 with the current denominator, or 0.205 after #1027. Consequently, the CI and p-value remain attached to a different estimand from the reported mean.

Please update the production calculation and regression test so the mean and SE describe the same threshold contrast. We still intend to merge this residual/contrast correction before #1027, then rebase and adapt #1027 afterward. As with the DiD pair, the current sibling tests pin opposite intermediate states, so #1027 will need to test the combined final behavior once this PR lands.

I appreciate the investigation here—the residual-shape diagnosis is exactly right, and it exposed this second RD-specific issue before we locked the wrong interval into a regression test.

Mari1988:
Thanks for the detailed catch — you're right, and I've pushed a fix that implements exactly what you described.

**What changed:**
- `RegressionDiscontinuity.algorithm()` now stores the design rows used to build the threshold contrast as `self.x_discon_design` (a `(2, p)` array; row 0 = just below the threshold, row 1 = just above), right where those rows are already built for the point estimate.
- `_compute_statistics_rd_ols` no longer searches `result.labels` for a coefficient name at all. It computes `c = result.x_discon_design[1] - result.x_discon_design[0]` and `se = sqrt(mse * c @ inv(X'X) @ c)`, matching your prescription directly.

This also fixes a second, related bug the old name-search had: any RD formula *without* an interaction term (e.g. `y ~ 1 + x + treated`, a legitimate parallel-slopes specification already used elsewhere in the test suite) never matched the `"treated" in label and ":" in label` search, so it silently fell back to a crude `std(residuals)/sqrt(n)` approximation — understating the SE by ~3.4x on this fixture. The contrast-based formula handles that case correctly too (it naturally collapses to `Var(treated)` when there's no interaction term), with no special-casing needed.

**Test plan:**
- Replaced the old test with `test_effect_summary_ols_rd_se_uses_discontinuity_contrast`, which independently verifies the contrast SE against `statsmodels`' own `OLSResults.t_test(contrast)` — not by reproducing CausalPy's own formula — for the interaction-term formula. Scaled by `sqrt((n-p)/n)` to isolate this fix from the separately-tracked denominator bug (#1027), so it doesn't overclaim correctness of that piece.
- Added `test_effect_summary_ols_rd_se_without_interaction_term`, same independent verification for the no-interaction case, with a regression guard against the old ~3.4x-too-small fallback.
- Updated `test_compute_statistics_rd_ols_fallback_path`: since the SE calculation no longer reads `result.labels`, the fallback is now triggered by removing `result.x_discon_design` rather than corrupting labels (the old trigger no longer does anything, since that attribute isn't read by the new code).
- Full suite: 1227 passed, 5 skipped. `make test-patch-cov`: 97% patch coverage (the 2 uncovered lines are a pre-existing, practically-unreachable `elif hasattr(X, "data")` fallback inherited unchanged from the original code).

Independently confirmed the numbers match your report: with the interaction formula, correct SE ≈ 0.201 with the current (still-biased) denominator, ≈ 0.205 once #1027 lands — both match what you reported.

Mari1988:
Rebased onto the latest main (now that #1026 merged) — no conflicts, since that touched only the unrelated DiD function. Full suite still passes (1228 tests), 97% patch coverage.

Mari1988:
Updated from main. Since #1057's \`_point_residuals()\` already fixes the residual-shape bug for RD (independently of this PR), I dropped that half of this PR's diff as redundant. What's left is exactly the contrast-vector SE fix from the review above: \`RegressionDiscontinuity.algorithm()\` stores \`x_discon_design\`, and \`_compute_statistics_rd_ols\` computes \`se = sqrt(mse * c @ inv(X'X) @ c)\` using it, unchanged from what was already reviewed.

Also updated the pre-existing \`test_effect_summary_ols_rd_residuals_are_per_observation\` (now upstream via #1057-adjacent work) since it asserted the old single-coefficient SE value, which this PR's contrast fix correctly changes — trimmed it back to just the residual-shape assertion, since SE correctness is now covered in full by \`test_effect_summary_ols_rd_se_uses_discontinuity_contrast\`.

Full suite: 1235 passed, 5 skipped. Patch coverage: 97% (gate is 96%; the 2 uncovered lines are the same pre-existing, practically-unreachable \`elif hasattr(X, "data")\` fallback noted before).

daimon-pymclabs:
Thanks — the contrast-SE fix and the no-interaction fallback fix both look right, and they compose cleanly with the now-unbiased denominator. Two things before merge:

1. **Update from main.** There's now a conflict in `_compute_statistics_rd_ols` and `test_reporting.py` since #1025/#1027 landed.

2. **Adapt the test expectations to the post-#1027 world.** #1027 merged ahead of this PR, so the unbiased `SSR/(n-p)` denominator is now on main. Your SE tests still scale the statsmodels contrast SE by `sqrt((n-p)/n)` (`biased_contrast_se`) to match the old convention, so they'll now assert the wrong value. After the rebase, drop that factor and compare against statsmodels' unbiased contrast SE directly (`unbiased_contrast_se` from `sm_fit.t_test(contrast)`) in both `test_effect_summary_ols_rd_se_uses_discontinuity_contrast` and `test_effect_summary_ols_rd_se_without_interaction_term`. The production code needs no change — only the test expectations.

Once CI is green on that, this is good to merge.

cetagostini:
Superseded by #1129, which landed the reviewed threshold-contrast standard error directly on the `pymc6_and_pymcmarketing1_migration` transition branch. Thank you @Mari1988 — the diagnosis and the contrast formula are exactly what shipped, including the parallel-slopes case that the old label search mishandled, and the test oracle now compares against `statsmodels`' unbiased contrast SE without rescaling because #1027's denominator fix is already in place.

PR #1036, opened 2026-07-22
Unify predict() to a canonical prediction container and delete the parallel legacy paths

### LINKED ISSUE #1035: Normalize prediction containers at the adapter boundary to remove backend branches from estimand code
# Normalize prediction containers at the adapter boundary to remove backend branches from estimand code

## Problem

Experiments that support both backends repeatedly branch on `is_bayesian` for a purely mechanical reason: PyMC backends return predictions as `az.InferenceData` (posterior draws of `mu`), while sklearn backends return bare numpy arrays. The branches do no statistical work — they only unwrap containers. Examples:

- `RegressionDiscontinuity.algorithm()` carries a `THIS IS SUBOPTIMAL AT THE MOMENT` banner around exactly this branch (draws contrast vs `np.squeeze` contrast).
- `StaggeredDifferenceInDifferences._predict_counterfactuals()` branches to compute `y_hat0` as mean-of-draws vs squeezed array.
- `InterruptedTimeSeries.algorithm()` branches around `calculate_impact` for a `treated_units` squeeze.
- `PiecewiseITS` has parallel Bayesian/OLS blocks with the same shape-juggling.

This duplication is accidental complexity: PR #1024 established that the experiment owns the estimand and the backend supplies response-scale expected predictions (see the estimand/capability matrix in `ARCHITECTURE.md`), but each experiment still re-implements the "what shape did the backend hand me" dance.

## Proposal

Add one method to the existing model adapter layer (`causalpy/experiments/model_adapter.py`, already the home for backend normalization — it normalizes fit inputs and squeezes sklearn `y` today):

```python
def predict_mu(self, X) -> xr.DataArray:
    """Response-scale expected outcome with canonical dims
    ("chain", "draw", "obs_ind", "treated_units") on every backend."""
```

- **PyMC adapters**: extract `posterior_predictive["mu"]` from `predict(X, var_names=["mu"])`.
- **sklearn adapter**: wrap point predictions in the same `xr.DataArray` with singleton `chain=1, draw=1` — a point estimate is a posterior with one atom.
- **pymc-forecast adapter**: same extraction as PyMC.

Estimand code then has one path on all backends, with xarray doing the lifting:

```python
# RD: works for both backends; OLS result is a 1x1-draw DataArray
disc = mu.isel(obs_ind=1) - mu.isel(obs_ind=0)

# Staggered DiD: mean over draws is a no-op for OLS
y_hat0 = self._model_backend.predict_mu(X_full).mean(dim=("chain", "draw"))
```

Backend awareness remains only where it is irreducible — reporting and plotting (HDIs and tail probabilities vs coefficient SEs/CIs, `_bayesian_plot` vs `_ols_plot`). Those branches must stay: computing an HDI from a singleton draw would be silently meaningless, so interval logic keeps guarding on `is_bayesian` (or an equivalent `has_posterior` adapter property).

`predict()` remains unchanged for callers that need `y_hat` (posterior predictive noise) or full `InferenceData` for plots.

## Non-goals / rejected alternatives

- **A `point_prediction(pred)` helper**: smallest diff, but two container shapes keep flowing through experiment code; it relocates the wart instead of removing it.
- **A `Prediction` result-wrapper class** with `.point` / `.draws` / operator overloading: reimplements xarray's arithmetic and broadcasting; over-engineering.
- **Changing summary/plot dispatch**: uncertainty representation genuinely differs by backend and stays branched.

## Scope

Candidate call sites to migrate (each deletes an `is_bayesian` branch or a parallel OLS block): `RegressionDiscontinuity.algorithm`, `StaggeredDifferenceInDifferences._predict_counterfactuals`, `InterruptedTimeSeries.algorithm`, `PiecewiseITS` prediction/impact blocks, and optionally the DiD/PrePostNEGD g-computation helpers (they already request `var_names=["mu"]` and would just call `predict_mu`). Migration can be incremental — one experiment per PR, RD first since it is self-flagged.

## Acceptance criteria

- `predict_mu` on all three adapter kinds returns identical dims/coords structure; property-style test asserting shape parity across backends.
- RD's `THIS IS SUBOPTIMAL` block is gone; `discontinuity_at_threshold` computed by one expression on both backends.
- No behavior change: existing estimand values and summaries are numerically identical (pin with equivalence tests).
- Interval/summary code paths still guard on backend capability so singleton-draw intervals can never be reported.

drbenvincent:
## Summary

This PR eliminates the Bayesian-vs-OLS `if/else` dispatching that was scattered across experiments and reporting, by making every model backend speak one language at the prediction boundary.

**One `predict()`, one container.** `ModelAdapter.predict()` now returns the same thing on every backend: response-scale expected outcomes (`mu`) as an `xr.DataArray` with canonical dims `(chain, draw, obs_ind, treated_units)`. Sklearn backends return singleton `chain`/`draw` dimensions — a point estimate is just a posterior with one atom. The interim `predict_mu()` method is gone; there is no parallel legacy path left.

**Impact is now plain arithmetic.** Because observed `y` and predictions live in the same labeled container, causal impact is `y - predict(X)` and cumulative impact is `.cumsum(dim="obs_ind")` — native xarray broadcasting, no helper methods, no backend branching. This made `calculate_impact()` / `calculate_cumulative_impact()` dead code on all three model classes (`PyMCModel`, `ScikitLearnAdaptor`, `PyMCForecastModel`), so they are deleted. The mu-not-y_hat contract they encoded is unchanged and now enforced at the adapter seam (`test_prediction_contract.py` asserts it through `ModelAdapter.predict()`).

**Reporting stopped sniffing containers.** `_extract_window` and `_extract_counterfactual` no longer duck-type across InferenceData / dict / ndarray / DataArray — they slice the canonical container and are a few lines each. `_select_treated_unit_numpy` is deleted. The three near-identical Bayesian/OLS `effect_summary` bodies in ITS, SyntheticControl, and PiecewiseITS collapsed onto a single shared `_effect_summary_timeseries` helper.

**SDiD's fake-InferenceData shim is gone.** It now builds the canonical prediction DataArray directly.

**Latent bug fixed along the way.** The OLS paths in ITS `_comparison_period_summary` and `analyze_persistence` fed `(1, 1, n)`-shaped arrays into the t-statistic helper, so `len(impact)` evaluated to 1 and degrees of freedom collapsed to zero. They now ravel to 1-D first.

**What deliberately survives.** Exactly one backend branch remains, inside `_effect_summary_timeseries` and the plot methods: Bayesian backends summarize posterior draws (HDI, tail probabilities, ROPE) while OLS backends report frequentist t-based intervals, because an HDI computed from a singleton draw would be silently meaningless. Everything upstream of that split is backend-blind.

## Change size

Net **−372 lines** vs `main` (646 insertions, 1,018 deletions), almost all of it deleted dispatching, container-juggling, and duplicated summary/plot bodies.

Closes #1035

## Testing

- [x] full suite: 1,225 passed, 5 skipped
- [x] `make test-patch-cov`: 99% patch coverage
- [x] contract tests reworked to assert mu-not-y_hat through the adapter

## Why this refactor is so effective (a note for contributors)

The pattern here generalizes well beyond `predict()`, so it is worth naming what actually happened:

- **Variation moved from the point of use to the point of origin.** With E experiments and B backends, dispatching at each use site costs E×B conditionals, and every new feature (summary, plot, report) gets written once per backend. Normalizing at one seam costs E+B: each backend pays once to emit the canonical container, and each consumer is written once against it. That arithmetic is where the −372 lines came from — and why the deletion compounds with every future feature.
- **The special case embeds in the general case.** The unlock was refusing the intuition that point estimates and posteriors are different kinds of things: *a point estimate is a posterior with one atom* (`chain=1, draw=1`). Once the special case is embedded in the general representation, the type distinction downstream evaporates. Labeled xarray dimensions are what make the embedding free — singleton dims cost nothing and broadcasting is by name, not position.
- **Parse, don't validate.** The mu-not-y_hat invariant used to be a convention every consumer had to independently honor; now it is enforced in one function (`_extract_mu`) and asserted by one test module (`test_prediction_contract.py`). Everything downstream gets to *assume* rather than *check*.
- **Polymorphism through data beats polymorphism through code.** Classic OO polymorphism unifies the call site but still needs a method body per type. Because the *output* is one canonical container, downstream code needs no dispatch at all — `impact = y - predict(X)` is not Bayesian code or OLS code, just arithmetic that is correct for both. The best branch is the one with nothing left to branch on.
- **Canonical shapes flush latent bugs.** The zero-degrees-of-freedom t-statistic bug survived for so long precisely because ambiguous shapes fail silently; named dims make wrong shapes loud.
- **Contracts also tell you what *not* to unify.** Exactly one branch deliberately survives (HDI/ROPE vs t-intervals), because an HDI from a singleton draw is meaningless. The goal is not pretending everything is the same — it is shrinking the genuinely different parts to their smallest possible surface.

### Further reading

- Bertrand Meyer, ["Applying 'Design by Contract'"](https://se.inf.ethz.ch/~meyer/publications/computer/contract.pdf), *IEEE Computer* 25(10), 1992 — the original articulation of interface contracts as enforceable obligations.
- Barbara Liskov & Jeannette Wing, ["A Behavioral Notion of Subtyping"](https://www.cs.cmu.edu/~wing/publications/LiskovWing94.pdf), *ACM TOPLAS* 16(6), 1994 — why a substitute must honor the contract, not just the signature (the sklearn adapter's singleton-draw container is Liskov substitution in action).
- Alexis King, ["Parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), 2019 — normalize into a truth-carrying representation at the boundary once, instead of re-checking everywhere.
- [Adapter pattern](https://refactoring.guru/design-patterns/adapter) (Gamma et al., *Design Patterns*, 1994) — though note the potent move here goes beyond the classic adapter: it is canonicalizing the *data*, not just unifying method signatures, that let downstream conditionals vanish rather than merely relocate.

### Best YouTube picks (applied, not academic)
1. [Validate, Parse, Guard — Ardalis (Steve Smith)](https://www.youtube.com/watch?v=EILI9EOWaJo)
Best starting point. Short, practical, no monads. Core idea: stop scattering validation/if checks through the codebase; normalize once at the boundary into types you can trust. This is the same instinct as _extract_mu() + contract tests — just in C# instead of Python/xarray.

2. [Anti-corruption Layer for mapping between Boundaries](https://www.youtube.com/watch?v=Dok2Ikcjaro)
~10 min. One external system speaks one language, your domain speaks another; a thin translation layer keeps the mess out. That's your ModelAdapter — PyMC returns InferenceData, sklearn returns ndarray, consumers see one xr.DataArray.

3. [Modularizing a 10-Year Monolith — Victor Lyuboslavsky (ACL section ~25:59)](https://www.youtube.com/watch?v=RHrPkuTbIVw)
Real production refactor. ACL as "the only place that understands both worlds." Good for the why when you're staring at a codebase full of is_bayesian and wondering if the refactor is worth it.

4. [Ports and Adapters Architecture in practice with .NET](https://www.youtube.com/watch?v=ld6epROxDcA)
Adapters must not know about each other; wiring lives in one place. Less about data shape, more about seam placement — still relevant to make_model_adapter().

Skip unless you want FP depth: [Monadic Parsers at the Input Boundary](https://www.youtube.com/watch?v=LLkbzt4ms6M) — directly cites Alexis King's essay but assumes monads; more theory than you need.

Watch order: Ardalis → ACL short video → monolith talk (ACL bit only). ~45 min total, maps cleanly onto what you already lived through in CausalPy.

### Follow-up issues applying the same pattern

- #1037 — unify `_bayesian_plot` / `_ols_plot` pairs
- #1049 — dispatch remaining branches on posterior size, not backend identity
- #1050 — canonical coefficients contract
- #1051 — canonical score contract
- #1052 — finish the adapter seam (`idata` LSP wart, maketables `isinstance` dispatch)


daimon-pymclabs:
Final review notes (non-blocking) — good to merge otherwise. Two things worth flagging:

**1. The "impact = y − predict(X)" subtraction contract is now implicit.** It's correct today only because ITS and SyntheticControl feed design `DataArray`s carrying real `obs_ind` coords (`pre_design["X"]` / `post_design["control"]`), so pred and observed `y` align on `obs_ind`. If a future code path ever fed a bare ndarray `X` into ITS/SC prediction, the sklearn adapter would stamp `obs_ind = arange(n)` and the subtraction against a datetime-indexed `y` would silently produce NaNs rather than raise. A one-line assert that pred and y `obs_ind` match (or a comment pinning the assumption) would harden the seam cheaply — symmetric with the `treated_units` column-count guard the adapter already does on predict.

**2. Behaviour change for RD.** The Bayesian `discontinuity_at_threshold` now drops the `treated_units` dim (`isel(obs_ind=…, treated_units=0)`) where before it retained it. Intentional and covered by the updated `test_rd`, just worth noting for anyone reading the posterior downstream.

drbenvincent:
Thanks for the review. Addressed in 91bb267b:

1. **obs_ind alignment guard** — added one-line asserts in ITS and SyntheticControl immediately before the impact subtraction, checking that predicted and observed `obs_ind` coords match exactly. This pins the alignment assumption cheaply, symmetric with the adapter's existing `treated_units` guard.

2. **RD `discontinuity_at_threshold` dropping `treated_units`** — confirmed intentional. RD is inherently single-unit, so the posterior over the discontinuity is cleaner as `(chain, draw)`; covered by the updated `test_rd`.

PR #990, opened 2026-07-09
[plotting] Migrate plotting from matplotlib/xarray to tidydraws + plotnine (#988)

drbenvincent:
## Summary

Migrates CausalPy experiment plotting from imperative matplotlib/ArviZ composition to a declarative [tidydraws](https://drbenvincent.github.io/tidydraws/) + [plotnine](https://plotnine.org/) stack (#988).

The public contract remains stable: experiment `plot()` methods return `(matplotlib.figure.Figure, matplotlib.axes.Axes | numpy.ndarray)`. Internally, Bayesian plot builders return `PlotSpec`, which contains a declarative `ggplot` and an optional narrow post-draw matplotlib overlay.

## Architecture

```
model output
    ↓
prediction_draws / dataarray_draws       # canonical long-form posterior draws
    ↓
summarize_draws / spaghetti_draws         # reusable posterior transformations
    ↓
posterior_kind_layers                     # ribbon / histogram / spaghetti geoms
    ↓
PlotSpec                                  # declarative plot + compatibility overlay
    ↓
BaseExperiment._finalize_plot             # stable public (Figure, Axes) return
```

ITS, Synthetic Control, Synthetic DiD, and Piecewise ITS additionally share a semantic `CausalPanelData` → `build_causal_panel_plot()` path for their common three-panel outcome/effect/cumulative-effect view.

Plotnine is imported consistently through the `p9` namespace. Matplotlib overlays remain only for compatibility behavior that is awkward to express declaratively, including date-axis formatting, custom legends required by `legend_kwargs`, and selected post-draw axis adjustments.

## What changed

- Added `plotnine>=0.15.7`, `tidydraws==0.4.3`, and direct `polars` dependencies while retaining Python 3.11 support.
- Migrated Bayesian ribbon, histogram, and spaghetti uncertainty views to shared tidy-draw transformations and Plotnine geoms.
- Migrated RD, RK, DiD, PrePostNEGD, ITS, PiecewiseITS, Synthetic Control, Synthetic DiD, Staggered DiD, and PanelRegression plotting surfaces.
- Replaced ArviZ plotting calls with explicit summaries and declarative layers; production experiment plotting no longer calls `az.plot_*`.
- Preserved explicit keyword-only public plot signatures and stable matplotlib return values.
- Added centralized validation for `kind`, `ci_kind`, and `num_samples`.
- Ensured spaghetti paths are isolated by panel, series, chain, and draw identity.
[...]

PR #1053, opened 2026-07-22
Unify `_bayesian_plot` / `_ols_plot` into a single backend-agnostic `_plot`

### LINKED ISSUE #1037: Unify _bayesian_plot / _ols_plot now that prediction containers are canonical
## Background

**Depends on #1036 being merged first.** This issue builds directly on the canonical prediction container work in #1035 / #1036.

Historically, every experiment carried paired `_bayesian_plot` / `_ols_plot` methods (plus paired `get_plot_data_bayesian` / `get_plot_data_ols`), with `BaseExperiment.plot()` dispatching on `is_bayesian`. That split was structural: the two backends returned different objects (`az.InferenceData` vs numpy arrays), so the plotting code had to differ.

After #1036, that reason is gone. Both backends emit the same canonical `(chain, draw, obs_ind, treated_units)` `xr.DataArray` from `ModelAdapter.predict()`, and OLS is just the degenerate case with a single chain/draw. What still differs between the paired plot methods is purely presentational:

1. **Uncertainty bands** — HDI ribbons for Bayesian, bare lines for OLS. This reduces to a single guard: `has_uncertainty = Y.sizes["chain"] * Y.sizes["draw"] > 1`, draw the band only when true. The guard belongs in one place (`plot_posterior_over_x` or a thin wrapper in `plot_utils.py`), not per experiment.
2. **Score labels** — Bayesian R² reported as mean ± std, OLS as a point value. A tiny label-formatting helper covers it.
3. **Bayesian-only flourishes** — e.g. the DiD causal-impact arrow annotation, spaghetti/histogram plot kinds. These stay behind the same `has_uncertainty` guard.

## Proposal

Replace each `_bayesian_plot` / `_ols_plot` pair with a single `_plot()` implementation per experiment, consuming the canonical DataArray and using a shared uncertainty guard. Collapse the `is_bayesian` dispatch in `BaseExperiment.plot()`. Unify the paired `get_plot_data_*` methods the same way.

## Scope

Eight experiments carry paired methods, and the pairs are heavily duplicated:

| Experiment | `_bayesian_plot` | `_ols_plot` |
|---|---|---|
| `interrupted_time_series.py` | 246 lines | 104 lines |
| `synthetic_control.py` | 189 | 123 |
| `diff_in_diff.py` | 163 | 92 |
| `piecewise_its.py` | 136 | 74 |
| `regression_discontinuity.py` | 107 | 79 |
| `staggered_did.py` | 133 (+63 group-time) | 128 (+63 group-time) |
| `panel_regression.py` | 19 | 9 |
| `synthetic_difference_in_differences.py` | 164 | raises `NotImplementedError` |

Rough estimate: **600–900 lines deleted**, plus simplification of the base-class plot dispatch.

## Caveats

- **Output should be pixel-equivalent for both backends.** The unified path must not subtly change what OLS plots look like (zero-width bands, legend entries appearing/disappearing). This argues for migrating experiment-by-experiment rather than big-bang, checking rendered docs notebooks along the way.
- **SDID currently raises `NotImplementedError` for `_ols_plot`.** Unification gives OLS SDID plotting for free — a nice side effect, but it should be a conscious decision (and tested), not an accident.
- `plot_posterior_over_x` already accepts the canonical DataArray and handles singleton chain/draw without breaking, so most of the plumbing exists; this is primarily a deletion exercise.

## Tasks

- [ ] Add shared uncertainty guard + score-label helper in `plot_utils.py`
- [ ] Migrate experiments one at a time: ITS, SC, DiD, Piecewise ITS, RD, Staggered DiD, Panel Regression, SDID
- [ ] Unify `get_plot_data_bayesian` / `get_plot_data_ols` pairs
- [ ] Collapse `is_bayesian` dispatch in `BaseExperiment.plot()`
- [ ] Decide on and test OLS SDID plotting (currently `NotImplementedError`)

drbenvincent:
Closes #1037.

## Summary

Now that every backend emits the canonical prediction container (`chain`/`draw`/`obs_ind`/`treated_units` DataArray), the paired `_bayesian_plot` / `_ols_plot` and `get_plot_data_bayesian` / `get_plot_data_ols` methods were dispatch-at-use: near-identical bodies that differed only in container unpacking and uncertainty rendering. This PR merges each pair into one `_plot` / `get_plot_data` per experiment and deletes the backend dispatch from `BaseExperiment`.

Uncertainty rendering now keys on **data properties, not backend identity**: the new `has_posterior_draws()` helper (`chain × draw > 1`) decides whether ribbons/HDI labels are drawn, and `extract_r2_score()` normalizes score-container access. A point estimate is a posterior with one atom, so the point-estimate path is the degenerate case of the general one; where OLS styling deliberately differs (black fit line, no band, plain R² title), the branch is a single `with_uncertainty` guard inside the one method rather than a duplicated method.

## Pixel equivalence

Every migrated experiment was verified pixel-identical (byte-identical PNGs via `cmp`) before/after on both backends using a seeded batch plot-generation script (19 figures: ITS, SC, DiD, PiecewiseITS, RD, StaggeredDiD event-study + group-time, Panel, SDiD, PrePostNEGD, RKink). The script and plot-regression test files will land from a separate feature branch to keep this PR a pure refactor.

## Code simplification record

| Commit | Experiment | Net lines |
|---|---|---|
| ITS | InterruptedTimeSeries (+ helpers, base transition) | −89 |
| SC | SyntheticControl | −121 |
| DiD | DifferenceInDifferences | −36 |
| PITS | PiecewiseITS | −52 |
| RD | RegressionDiscontinuity | −51 |
| Panel | PanelRegression | −65 |
| SDiD | StaggeredDiD | −151 |
| Bayes-only | SDiD/RKink/PrePostNEGD renames | −6 |
| Base | Remove dispatch + stubs, docs | −52 |
| **Total** | | **−623** (986 insertions, 1609 deletions) |

## Surviving backend branches (with justification)

- `PanelRegression.get_plot_data` / `_plot_coefficients_internal`: the experiment stores no canonical in-sample prediction container, so fitted values come from backend-native objects (`idata.posterior["mu"]` vs sklearn `predict`), and `az.plot_forest` genuinely requires `idata`. Both branches are isolated in single methods with `ponytail:` upgrade-path comments.
- `StaggeredDiD._get_group_time_placebo_data`: HDI bounds need posterior draws, sample dispersion needs only point residuals — genuinely different statistics, dispatched on `has_posterior_draws`, isolated to one seam.

## Test plan

- [x] Full suite: 1223 passed, 5 skipped
- [x] `make test-patch-cov`: 99% patch coverage (gate 96%)
- [x] 19/19 baseline-vs-after PNGs byte-identical on both backends

### DIFF
--- a/.agents/skills/review-pr/resources/code-patterns.md
+++ b/.agents/skills/review-pr/resources/code-patterns.md
@@ -7,9 +7,8 @@ Use this resource when reviewing source-code diffs. It collects CausalPy contrac
 All experiment classes in `causalpy/experiments/` inherit from `BaseExperiment`.
 
 - Declare `supports_ols: bool` and `supports_bayes: bool`.
-- If Bayesian is supported, implement `_bayesian_plot()` and `get_plot_data_bayesian()`.
-- If OLS is supported, implement `_ols_plot()` and `get_plot_data_ols()`.
-- Dispatch by model type with `isinstance(self.model, PyMCModel)` and `isinstance(self.model, RegressorMixin)`.
+- Implement a single backend-agnostic `_plot()` (and `get_plot_data()` where applicable) that consumes the canonical prediction container.
+- Key uncertainty rendering on data properties (`has_posterior_draws()` from `causalpy.plot_utils`), not backend identity; any surviving `is_bayesian` branch needs written justification.
 - Keep the data index named `"obs_ind"`.
 - Parse formulas through patsy `dmatrices()`.
 - Use project exceptions from `causalpy.custom_exceptions` for formula, data, and index errors.
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -46,7 +46,7 @@ The optional third backend, `PyMCForecastModel` (`causalpy/pymc_forecast_models.
 
 ## Experiment Lifecycle
 
-Instantiation fits eagerly in `__init__`: `_build_design_matrices()` → `_prepare_data()` → `algorithm()`. There is no separate `.fit()` on the experiment. Each subclass's public `plot(*, ...)` delegates to `_render_plot()`, which dispatches to `_bayesian_plot()` or `_ols_plot()`. `effect_summary()` returns `EffectSummary(table, text)` using helpers from `causalpy.reporting`.
+Instantiation fits eagerly in `__init__`: `_build_design_matrices()` → `_prepare_data()` → `algorithm()`. There is no separate `.fit()` on the experiment. Each subclass's public `plot(*, ...)` delegates to `_render_plot()`, which calls the subclass's backend-agnostic `_plot()`. Uncertainty rendering keys on data properties of the canonical prediction container (`has_posterior_draws()`), not on backend identity. `effect_summary()` returns `EffectSummary(table, text)` using helpers from `causalpy.reporting`.
 
 ## Experiment Inventory
 
@@ -91,7 +91,7 @@ Instantiation fits eagerly in `__init__`: `_build_design_matrices()` → `_prepa
 
 Copy the closest existing experiment or model and follow the `BaseExperiment` contract:
 
-- Declare `supports_ols` / `supports_bayes` (and `supports_pymc_forecast` to opt into the optional pymc-forecast backend); implement `_bayesian_plot()` / `_ols_plot()` only for supported backends
+- Declare `supports_ols` / `supports_bayes` (and `supports_pymc_forecast` to opt into the optional pymc-forecast backend); implement a single backend-agnostic `_plot()` (and `get_plot_data()`) that consumes the canonical prediction container, keying uncertainty rendering on `has_posterior_draws()` rather than backend identity
 - `algorithm()` with the fit/predict/impact flow; `effect_summary()` via helpers in `causalpy.reporting`
 - Public `plot(*, ...)` with a kwarg-only signature that delegates to `_render_plot()` — bare `*args` / `**kwargs` are forbidden on the public surface (enforced by `causalpy/tests/test_public_plot_signatures.py`). For experiments without a unified plot view (e.g. `InversePropensityWeighting`, `InstrumentalVariable`), declare an explicit `plot()` stub that raises `NotImplementedError`. For `hdi_prob` defaults, use ``Defaults to :data:`~causalpy.constants.HDI_PROB` (currently 0.94).`` in the docstring.
 - Raise `FormulaException`, `DataException`, or `BadIndexException` from `causalpy.custom_exceptions` for formula, data, and index errors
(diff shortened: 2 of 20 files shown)


daimon-pymclabs:
Final review notes (none blocking, mirrors the flavor of the #1036 comment):

**1. The strongest safety net for this change lives outside the PR.** This is a pure plotting refactor, and per the PR body the byte-identical PNG regression tests are landing "from a separate feature branch." CI here proves the `_plot` methods run and return figures, not that rendered output is unchanged. The author verified 19/19 figures pixel-identical locally (which I trust), but until those tests land there's no CI guard against a future visual regression on this seam. Worth landing them soon after, or tracking.

**2. Two discriminators for one backend split inside a single method.** In `interrupted_time_series` and `piecewise_its`, the plot body branches on `has_posterior_draws()` (a data property) while the R² title branches on `isinstance(self.score, pd.Series)` (a container type). They agree today because Bayesian gives Series + draws and OLS gives scalar + no-draws. If a future backend ever broke that pairing (a point-estimate backend reporting a Series score, or a bootstrap OLS carrying dispersion), the title and the bands would disagree. `synthetic_control._get_score_title` already keys purely on the data property (`r2_std_val is not None`), so unifying ITS/PITS on the same predicate would make the seam uniform. Cheap, same flavor as the implicit-contract note on #1036.

**3. `has_posterior_draws` defines a point estimate as `chain × draw == 1`,** so a genuine Bayesian fit sampled with a single chain and single draw would render OLS-style with no bands. Not realistic for MCMC, and it matches the helper's stated contract ("a point estimate is a posterior with one atom"), so it's a definitional edge rather than a bug, just worth knowing it's the discriminator.

drbenvincent:
Thanks for the review. Point by point:

**1. Regression harness outside the PR** — agreed. The seeded plot-generation script is written and was used for the 19/19 byte-identical verification here; it lands as a small follow-up PR right after this merges (kept out deliberately so this PR stays a pure refactor with a clean simplification record). It is also the intended eyeball harness for the #990 plotnine migration, where pixel equality will no longer hold.

**2. Two discriminators for one split** — fixed in d8b63cc3. ITS and PiecewiseITS titles now key on `r2_std_val is not None` (same data property `SyntheticControl._get_score_title` uses) instead of `isinstance(score, pd.Series)`, so the title and the bands can no longer disagree if a future backend decouples score container from draw count. Re-verified pixel-identical on both backends.

**3. `chain × draw == 1` edge** — acknowledged as definitional and intended: a one-atom posterior *is* a point estimate, so rendering it without bands is the correct degenerate behavior (an HDI from a single draw would be a zero-width band anyway). The contract is stated in the `has_posterior_draws` docstring.

PR #1057, opened 2026-07-22
Dispatch remaining Bayesian/OLS branches on posterior size, not backend identity

### LINKED ISSUE #1049: Dispatch remaining Bayesian/OLS branches on posterior size, not backend identity
## Background

**Depends on #1036 being merged first.** Follow-up to the canonical prediction container work in #1035 / #1036, sibling of #1037 (which covers the plotting surface).

#1036 deliberately kept exactly one backend branch alive: Bayesian backends summarize posterior draws (HDI, tail probabilities, ROPE) while OLS backends report frequentist t-based intervals, because an HDI computed from a singleton draw would be silently meaningless. That split is statistically legitimate — but the branches currently key off **backend identity** (`self._model_backend.is_bayesian` / `.is_ols`), which is a proxy for the property they actually care about: *does the prediction container carry more than one draw?*

The honest discriminator is the container itself:

```python
has_draws = pred.sizes["chain"] * pred.sizes["draw"] > 1
```

This is the same guard #1037 proposes for plots, applied to the remaining non-plot surfaces.

## Why it matters beyond aesthetics

Type sniffing hardcodes today's backend lineup into the statistics. Dispatching on the container property means any future backend that emits many draws — a bootstrap-resampled sklearn wrapper, a conformal-prediction backend, simulation-based inference — gets HDIs, tail probabilities, and ROPE **for free**, with zero new branches. Conversely, a degenerate Bayesian run would correctly fall back to point summaries instead of producing a meaningless HDI. The contract, not the type, decides.

## Affected surfaces (non-plot; plots are #1037)

- `causalpy/reporting.py` — `is_bayesian` branches at ~463 and ~686
- `_effect_summary_timeseries` (shared ITS / SyntheticControl / PiecewiseITS helper from #1036) — the one surviving branch inside it
- `causalpy/experiments/diff_in_diff.py` — `effect_summary` dispatch (~716)
- `causalpy/experiments/staggered_did.py` — `is_bayesian` branches (~516, ~1282)
- `causalpy/experiments/panel_regression.py` — the most branch-dense file left (~9 `is_bayesian` sites); the summary/reporting subset belongs here, the plot subset to #1037
- `causalpy/experiments/prepostnegd.py`, `synthetic_difference_in_differences.py` — remaining `is_ols` guards worth auditing with the same lens

## Tasks

- [ ] Add a single shared predicate (e.g. `has_posterior_draws(pred)` in a utility module, or a property on the canonical container's accessor) — one definition, not per-file re-derivations
- [ ] Migrate the reporting and effect-summary branches from `is_bayesian` to the container predicate
- [ ] Audit remaining `is_bayesian` / `is_ols` call sites and classify each as: (a) container-property dispatch → migrate, (b) genuinely backend-specific (e.g. `idata` access, prior sensitivity) → leave with a comment
- [ ] `is_bayesian` / `is_ols` should end up rare enough that each remaining use is individually justifiable

## Non-goals

- Constructor-time backend validation (`make_model_adapter` support flags) stays type-based — that is a trust-boundary check, not statistical dispatch.
- Bayesian-only diagnostics (prior sensitivity, placebo checks needing `idata`) are genuinely backend-specific and out of scope.

drbenvincent:
# Dispatch remaining Bayesian/OLS branches on posterior size, not backend identity

## Summary

Replaces the remaining `is_bayesian` / `is_ols` statistical dispatch in reporting and effect-summary code with a data-property check on the canonical prediction container: `has_posterior_draws(pred)` (`chain × draw > 1`). The branch condition is now honest — the statistics care about whether the container carries more than one draw, not which library fitted it. Any future many-draw backend (bootstrap-resampled sklearn, conformal prediction, SBI) gets HDIs, tail probabilities, and ROPE for free, and a degenerate single-draw Bayesian run correctly falls back to point summaries instead of emitting a meaningless HDI.

Fixes #1049

## Changes

- **Shared predicate**: `has_posterior_draws` moved from `plot_utils.py` to `causalpy/utils.py` (neutral home, so `reporting.py` does not import the plotting module), with a re-export in `plot_utils` so existing importers are unchanged. One definition, no per-file re-derivations.
- **Migrated dispatch sites** (backend flag → container property):
  - `reporting.py`: `_effect_summary_rd` keys on `discontinuity_at_threshold`; `_effect_summary_timeseries` keys on `windowed_impact` (its now-unused `result` parameter removed, ITS/SC/Piecewise callers updated)
  - `diff_in_diff.py` `effect_summary` keys on `y_pred_control`
  - `staggered_did.py` `_aggregate_effects` keys on `y_pred`
  - `interrupted_time_series.py` `_comparison_period_summary` and `analyze_persistence` key on `intervention_impact`
  - `regression_discontinuity.py` `summary()` keys on `discontinuity_at_threshold`
  - `utils.py` `extract_lift_for_mmm` checks `has_posterior_draws(post_impact)` instead of `isinstance(model, PyMCModel)` (also rejects degenerate runs that would silently produce `sigma = 0`)
- **Backend-agnostic point-summary fallback**: new `_point_residuals` helper computes in-sample residuals through the adapter's canonical `predict` instead of sklearn-native `result.model.predict`, and the DiD point path uses `_as_scalar(causal_impact)`. A degenerate single-draw Bayesian run now genuinely falls back to t-intervals.
- **Behavior change / bugfix — RD OLS intervals**: the old `_compute_statistics_rd_ols` subtracted a bare `(n,)` prediction array from the `(n, 1)` y DataArray, broadcasting to an `(n, n)` residual matrix and inflating the MSE ~9x (~3x too-wide standard errors, too-conservative p-values) — the same broadcast bug fixed for DiD OLS in #1026. `_point_residuals` flattens both sides, so RD OLS `effect_summary()` confidence intervals and p-values change (narrower/correct) for existing users. A regression test pins the corrected SE against an independent statsmodels fit. DiD OLS numbers are unchanged (already flattened by #1026).
- **Audited survivors, each with a written justification comment**:
  - Constructor-time capability validation (trust boundary): `synthetic_control`, `prepostnegd`, `synthetic_difference_in_differences`
  - Adapter-seam extraction from backend-native storage: `diff_in_diff` fit-time coefficient extraction
  - `panel_regression.py`: coefficient posteriors and in-sample `mu`/`y_hat` live in backend-native `idata` because PanelRegression stores no canonical container; the existing `ponytail:` note documents the upgrade path (store canonical in-sample predictions at fit time)
  - `placebo_in_time`, `prior_sensitivity`, `outcome_falsification`: Bayesian-only diagnostics needing `idata` (issue non-goal, already documented)

## Testing

- New contract test `test_effect_summary_timeseries_dispatches_on_draws_not_backend`: the same `_effect_summary_timeseries` call produces HDI columns with a 200-draw container and t-interval columns with a singleton container — the container, not the backend flag, decides.
- New regression pin `test_effect_summary_ols_rd_residuals_are_per_observation`: asserts per-observation residual shape and pins the RD OLS `effect_summary()` SE against an independent statsmodels fit (up to the documented RSS/n denominator convention), so the corrected numbers cannot silently move again.
- Full suite green. `prek run --all-files` clean. `make test-patch-cov`: 100% patch coverage (gate: 96%).

## Checklist

- [x] `prek run --all-files` passes
- [x] `make test-patch-cov` passes (100% on changed lines)
- [x] Remaining `is_bayesian` / `is_ols` uses individually justified in comments
- [x] No user-facing behavior change for existing backends, **except RD OLS `effect_summary()` intervals/p-values, which change deliberately** — the previous values were inflated by the `(n, n)` residual-broadcast bug (see above); the fix is pinned by a regression test

### DIFF
--- a/causalpy/reporting.py
+++ b/causalpy/reporting.py
@@ -33,7 +33,7 @@ import xarray as xr
 from scipy.stats import t
 
 from causalpy.constants import HDI_PROB
-from causalpy.utils import _as_scalar
+from causalpy.utils import _as_scalar, has_posterior_draws
 
 
 @dataclass
@@ -460,8 +460,8 @@ def _effect_summary_rd(
     """Generate effect summary for Regression Discontinuity experiments."""
     discontinuity = result.discontinuity_at_threshold
 
-    if result._model_backend.is_bayesian:
-        # PyMC model: use unified scalar functions
+    if has_posterior_draws(discontinuity):
+        # Posterior draws present: use unified scalar functions
         hdi_prob = 1 - alpha
         stats = _compute_statistics_scalar(
             discontinuity, hdi_prob=hdi_prob, direction=direction, min_effect=min_effect
@@ -634,7 +634,6 @@ def _extract_counterfactual(result, window_coords, treated_unit=None):
 
 
 def _effect_summary_timeseries(
-    result,
     windowed_impact: xr.DataArray,
     counterfactual: xr.DataArray,
     window_coords,
@@ -650,15 +649,14 @@ def _effect_summary_timeseries(
     """Build an :class:`EffectSummary` for time-series experiments (ITS, SC,
     Piecewise ITS) from canonical impact/counterfactual containers.
 
-    The single irreducible backend branch lives here: Bayesian backends
-    summarize posterior draws (HDIs, tail probabilities, ROPE), while OLS
-    backends report frequentist t-based intervals — an HDI computed from a
-    singleton draw would be silently meaningless.
+    The single irreducible statistical branch lives here, keyed on the
+    container itself: predictions carrying posterior draws are summarized
+    with HDIs, tail probabilities, and ROPE, while single-draw (point
+    estimate) predictions report frequentist t-based intervals — an HDI
+    computed from a singleton draw would be silently meaningless.
 
     Parameters
     ----------
-    result
-        Experiment result object exposing ``_model_backend``.
     windowed_impact : xr.DataArray
         Causal impact in the analysis window with canonical prediction
         dimensions.
@@ -683,7 +681,7 @@ def _effect_summary_timeseries(
         Experiment tag ("its", "sc", "piecewise_its") for tailored
         assumptions text.
     """
-    if result._model_backend.is_bayesian:
+    if has_posterior_draws(windowed_impact):
         hdi_prob = 1 - alpha
         stats = _compute_statistics(
             windowed_impact,
@@ -1491,6 +1489,26 @@ def _compute_statistics_ols(
     return stats
 
 
+def _point_residuals(result) -> np.ndarray:
+    """In-sample point residuals via the canonical prediction container.
+
+    Uses the model adapter's canonical ``predict`` output (a point estimate
+    is the first — and for point backends only — draw), so the t-based
+    point-summary path works for any backend, including a degenerate
+    single-draw Bayesian run.
+
+    ``y`` may have shape ``(n, 1)`` with dims ``(obs_ind, treated_units)``
+    while the fitted values are conceptually ``(n,)``; both are flattened to
+    1-D so they align positionally on ``obs_ind`` (letting xarray align them
+    would broadcast against ``treated_units`` and produce an ``(n, n)``
+    array).
+    """
+    y = np.asarray(result.design["y"]).reshape(-1)
+    pred = result._model_backend.predict(X=np.asarray(result.design["X"]))
+    y_fitted = np.asarray(pred.isel(chain=0, draw=0)).reshape(-1)
+    return y - y_fitted
+
+
 def _compute_statistics_did_ols(
     result,
     alpha=0.05,
@@ -1509,22 +1527,12 @@ def _compute_statistics_did_ols(
     dict
         Dictionary of statistics
     """
-    causal_impact = result.causal_impact  # scalar
+    causal_impact = _as_scalar(result.causal_impact)
 
     # Calculate standard error from model residuals
-    # Get fitted values and residuals
-    X_da = result.design["X"]
-    y_da = result.design["y"]
-    y_pred = result.model.predict(X_da)
-    # y_da has shape (n, 1) with dims (obs_ind, treated_units); y_pred is a
-    # bare (n,) array with no dimension names. Subtracting them directly lets
-    # xarray align y_pred positionally against the *last* axis (treated_units,
-    # size 1) instead of obs_ind, producing an (n, n) array of every
-    # observation's y minus every other observation's prediction instead of
-    # per-observation residuals. Flatten both to 1-D first so they align by
-    # position on obs_ind as intended.
-    residuals = np.asarray(y_da).reshape(-1) - np.asarray(y_pred).reshape(-1)
+    residuals = _point_residuals(result)
     mse = np.mean(residuals**2)
+    X_da = result.design["X"]
     n, p = X_da.shape
     df = n - p
 
@@ -1652,12 +1660,10 @@ def _compute_statistics_rd_ols(result, alpha=0.05):
     """Compute statistics for RD scalar effect with OLS model."""
     discontinuity = _as_scalar(result.discontinuity_at_threshold)
 
-    # Calculate standard error from model
-    X_da = result.design["X"]
-    y_da = result.design["y"]
-    y_pred = result.model.predict(X_da)
-    residuals = y_da - y_pred
+    # Calculate standard error from model residuals
+    residuals = _point_residuals(result)
     mse = np.mean(residuals**2)
+    X_da = result.design["X"]
     n, p = X_da.shape
     df = n - p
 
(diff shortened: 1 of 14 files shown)


causalpy/reporting.py
```diff
@@ -1652,12 +1660,10 @@ def _compute_statistics_rd_ols(result, alpha=0.05):
     """Compute statistics for RD scalar effect with OLS model."""
     discontinuity = _as_scalar(result.discontinuity_at_threshold)
 
-    # Calculate standard error from model
-    X_da = result.design["X"]
-    y_da = result.design["y"]
-    y_pred = result.model.predict(X_da)
-    residuals = y_da - y_pred
+    # Calculate standard error from model residuals
+    residuals = _point_residuals(result)
```
daimon-pymclabs, line 1666: This is a **behavior change for RD OLS**, not a no-op refactor. The old path here was `residuals = y_da - y_pred`, where `y_da` is a `(n, 1)` DataArray (`obs_ind`, `treated_units`) and `y_pred` is a bare `(n,)` array. That subtraction broadcasts to `(n, n)` instead of computing per-observation residuals, so `mse = np.mean(residuals**2)` was averaging an `(n, n)` matrix. `_point_residuals` flattens both to 1-D and fixes it.

Verified on the branch with the `rd` dataset + `LinearRegression`:
```
old naive resid shape: (100, 100)  old mse = 1.234761
new residuals shape:   (100,)      new mse = 0.135483
```
That's a ~9x MSE inflation, so on `main` the RD OLS standard errors were ~3x too wide and p-values correspondingly too conservative. This PR corrects them, which changes user-facing output for every existing RD OLS user.

Two asks:
1. Reword the "no user-facing behavior change" checklist item to carve out RD OLS as a bugfix (worth a CHANGELOG line too).
2. Add a regression test pinning RD OLS `effect_summary()`/`summary()` numbers. The fact that the full suite still passed 1224 tests means **nothing currently asserts these values** — that's exactly how the old broadcast bug survived, and without a pin the next refactor can silently move them again. (DiD OLS already flattened, so it's genuinely unchanged — I confirmed its numbers didn't move.)

causalpy/reporting.py
```diff
@@ -1491,6 +1489,26 @@ def _compute_statistics_ols(
     return stats
 
 
+def _point_residuals(result) -> np.ndarray:
```
daimon-pymclabs, line 1492: Nit / for the record: for the degenerate single-draw Bayesian fallback, `pred.isel(chain=0, draw=0)` takes the *first* draw as the point estimate rather than a posterior mean. That's fine here because this path is only reached when `has_posterior_draws` is False, i.e. `chain*draw == 1`, so draw 0 is the only draw. No change needed — just flagging that the "first draw" choice is load-bearing on that invariant, so if a `chain=1, draw=1`-but-not-really container ever appears it'd silently pick an arbitrary sample.

daimon-pymclabs (commented):
Reviewed the whole diff and exercised RD/DiD OLS on the branch. The core refactor is sound: dispatching HDI-vs-t on `has_posterior_draws(container)` instead of `is_bayesian` is the honest discriminator, the shared predicate now living in `causalpy.utils` with a `plot_utils` re-export is the right home, and the audited survivors (constructor trust-boundary checks, adapter-seam coefficient extraction, PanelRegression's backend-native `idata`) are each correctly classified and commented. The `_effect_summary_timeseries` signature cleanup (dropping the unused `result` param) is a nice bonus.

One thing the description gets wrong, and it matters: the checklist says **"No user-facing behavior change for existing PyMC/sklearn backends."** That isn't true for RD OLS. The new `_point_residuals` helper silently fixes a latent broadcasting bug in the old `_compute_statistics_rd_ols`, so RD OLS `summary()` / `effect_summary()` intervals and p-values change. See the inline comment. This is a genuine bugfix and a good one, but it should be called out as a behavior change, not folded into a refactor claiming none, and it needs a regression test pinning the numbers.

Everything else looks mergeable. Details inline.

causalpy/reporting.py
drbenvincent: Good catch — you're right, this was a real behavior change hiding under a refactor label, thanks for running the numbers. Addressed in 6e70c31b:

1. **PR description reworded**: the checklist now explicitly carves out RD OLS `effect_summary()` intervals/p-values as a deliberate bugfix (same `(n, n)` residual-broadcast bug that #1026 fixed for DiD OLS), with the ~9x MSE / ~3x SE inflation documented in the Changes section. The repo has no CHANGELOG file — release notes are assembled from PR descriptions — so the callout lives there.
2. **Regression pin added**: `test_effect_summary_ols_rd_residuals_are_per_observation` asserts the per-observation residual shape and pins the reported SE against an independent statsmodels fit (up to the documented RSS/n vs RSS/(n-p) denominator convention, same as the DiD pin), plus a guard asserting the SE stays below 2x the unbiased value so the broadcast variant can't come back silently.

Agreed that the 1224 green tests proving nothing here was exactly the lesson — the pin mirrors the #1026 DiD test so both OLS scalar paths are now locked.

causalpy/reporting.py
drbenvincent: Fair point that the invariant was load-bearing. Rather than leave it implicit, 6e70c31b switches the collapse from `pred.isel(chain=0, draw=0)` to `pred.mean(dim=["chain", "draw"])` — identical for the singleton containers this path is defined for, but well-defined (posterior mean, not an arbitrary sample) if a many-draw container ever slips through, and the docstring now states that reasoning explicitly.

PR #1059, opened 2026-07-22
Canonical score contract across model backends

### LINKED ISSUE #1051: Canonical score contract: unify score() output format across backends
## Background

**Depends on #1036 being merged first.** Follow-up applying the canonical-container idea from #1035 / #1036 to the `score()` surface.

`ModelAdapter.score()` currently has no output contract — it forwards whatever the backend produces:

- PyMC backends return a `pd.Series` keyed like `unit_0_r2` / `unit_0_r2_std` (Bayesian R² with dispersion across draws).
- sklearn backends return a bare float (single-output) from `RegressorMixin.score`.

So every consumer must know which shape it got. Concretely, `SyntheticControl._get_score_title` branches on `is_bayesian` purely to unpack the two formats, and #1037 has to carry a "score labels" formatting caveat for the same reason. This is the last obviously un-contracted output of the adapter trio (fit / predict / score).

## Proposal

Define one canonical score container returned by all backends — the natural choice mirroring the prediction contract is a `pd.Series` (or small `xr.Dataset`) with per-treated-unit R² keys, where the dispersion entry (`*_r2_std`) is present only when the backend carries posterior draws:

- PyMC / pymc-forecast: unchanged, `unit_{i}_r2` + `unit_{i}_r2_std`.
- sklearn: adapter wraps the float(s) into the same keyed structure, `unit_{i}_r2` only (a point estimate has no dispersion — the same singleton-embedding logic as `predict()`).

Consumers then read one format, and the "show ± std?" decision reduces to key presence (or the draws predicate from the sibling draw-size-dispatch issue), not backend identity.

## Expected payoff

- Deletes the `is_bayesian` branch in `_get_score_title` and prevents the same branch from re-growing in other experiments' score displays.
- One score-formatting helper (shared with #1037's plot titles) instead of per-experiment unpacking.
- Multi-output sklearn scoring gets an explicit, tested home (currently `RegressorMixin.score` on 2-D `y` returns uniform-average R², silently different in kind from the per-unit PyMC keys).

## Tasks

- [ ] Specify the canonical score format and document it next to the prediction contract
- [ ] Normalize in `SklearnModelAdapter.score()`; verify `PyMCForecastAdapter` conforms
- [ ] Shared score-label formatting helper; migrate `_get_score_title` and any summary paths reading `self.score`
- [ ] Contract tests: same keys across backends, per-unit values for multi-output sklearn
- [ ] Audit stored `self.score` uses in experiments (`synthetic_control.py`, `interrupted_time_series.py`, `piecewise_its.py`, `regression_discontinuity.py`, `regression_kink.py`) for format assumptions

## Caveat

`self.score` is user-visible on fitted experiments, so this is a mild API change for sklearn users (float → keyed Series). Worth a release note; the Bayesian format is unchanged.

drbenvincent:
## Summary

- Normalize every model adapter's `score()` result to a keyed `pandas.Series`.
- Report one sklearn R² per treated unit instead of sklearn's uniform multi-output average.
- Centralize score-title formatting so consumers branch on dispersion presence rather than backend identity.

Fixes #1051

## Changes

- Define and document the canonical `unit_{i}_r2` / optional `unit_{i}_r2_std` score contract.
- Normalize sklearn scores at the adapter seam and retain the existing PyMC and pymc-forecast format.
- Migrate Synthetic Control, Interrupted Time Series, Piecewise ITS, Regression Discontinuity, and Regression Kink plot titles to the shared formatter.
- Add contract coverage for single-output, multi-output, and dispersion-aware formatting.

## Upgrade note

For sklearn-backed experiments, `experiment.score` is now a keyed `pandas.Series` rather than a float. Single-output callers should use `experiment.score["unit_0_r2"]`.

Direct adapter callers should pass `sklearn.metrics.r2_score` keyword arguments to `SklearnModelAdapter.score()` (for example `sample_weight`). These are not forwarded to the underlying estimator `score` method, and `multioutput` is fixed to `"raw_values"` so each treated unit keeps its own `unit_{i}_r2` entry.

## Test plan

- [x] `prek run --all-files`
- [x] Targeted adapter, contract, and experiment tests: 115 passed, 2 skipped
- [x] `make test-patch-cov`: 1,226 passed, 5 skipped; 100% patch coverage
- [x] `make html`: succeeded; no warnings from the changed page


drbenvincent:
## Automated triage

**Why:**
- **Public API change**: `experiment.score` for sklearn-backed experiments changes from a `float` to a keyed `pandas.Series` — existing callers accessing it as a float will break. The PR acknowledges this in an upgrade note.
- Cross-cutting change touching 6 experiment modules (Synthetic Control, ITS, Piecewise ITS, Regression Discontinuity, Regression Kink) plus `plot_utils.py` and `model_adapter.py`.
- Normalizes score formatting across backends, meaning downstream consumers that branch on backend identity need to be verified.
- All CI checks pass, 100% patch coverage, docs build clean.

**Review focus:**
1. Verify the sklearn R² per-treated-unit reporting is semantically correct (not just a shape change).
2. Check that the `unit_{i}_r2` / `unit_{i}_r2_std` naming convention is documented for downstream consumers.
3. Confirm the upgrade note in the description is sufficient for the release changelog — this is a breaking change for sklearn users.

**Confidence:** high

daimon-pymclabs:
Final review — no major concerns. Correct, well-scoped, and the natural closing move on the #1036 → #1053 arc. Good to merge.

**What I verified beyond reading.** Every adapter now honours the `unit_{i}_r2` contract: PyMC (`pymc_models.py:451`) and pymc-forecast (`pymc_forecast_models.py:395`) already emitted `unit_{i}_r2` / `unit_{i}_r2_std` per treated unit, and sklearn now joins them at the seam. So the stricter `extract_r2_score` (which dropped its old scalar and plain-`"r2"` fallbacks and hard-indexes `score[key]`) is safe: no live code path produces a bare float or a plain-`r2` container anymore. The sklearn semantic change is the real payoff and it's numerically sound: the old `RegressorMixin.score` returned one `uniform_average` float; the new path calls `r2_score(..., multioutput="raw_values")`, which is identical for single-output (the contract test pins `score["unit_0_r2"] == model.score(X, y)`) and now reports one R² per unit for multi-output instead of silently averaging. `format_r2_score` handles `round_to=None` correctly via `round_num`'s own None-default, so the removal of kink's explicit `if round_to is not None else 2` guard is a safe simplification, not a regression. The one other consumer, `maketables_adapters._safe_r2_value`, still reads the Series and means the per-unit entries, so tables are unaffected (its scalar branch is now dead but harmless).

**Three things to flag, none blocking:**

1. **Same implicit-contract shape as #1036/#1053.** `extract_r2_score` now does a bare `score[f"unit_{unit_index}_r2"]` with no guard, so a `unit_index` that outruns the scored units raises `KeyError` at plot time rather than the old `(None, None)`. It's correct today because every call site derives the index from the same `treated_units` list that produced the score (SC's `_get_score_title` especially). But it's the symmetric guard again: a one-line assert or comment pinning "one score entry per treated unit" would harden the seam for the same cheap price paid on the last two PRs.

2. **The sklearn `score` kwarg surface changed silently.** `**kwargs` used to flow to `model.score(...)`; it now flows to `r2_score(...)`. The docstring is updated to say so, and `sample_weight` survives the move, but any caller passing a score-specific kwarg would now hit `r2_score`'s signature instead. Realistically no one does, just noting the contract moved.

3. **Outside this PR, but I tripped over it while tracing consumers:** `test_integration_its_new_timeseries.py:205` asserts `"r2" in score`, and on a `pd.Series` `in` checks index labels, so against a `unit_0_r2`-keyed container that's `False`. It's untouched by #1059 and only runs with pymc-extras installed (so it's not in the green CI here), but it looks latently inconsistent with the very contract this PR canonicalizes. Worth a follow-up glance, not a blocker on this merge.

**The wins, in one place:**
- **One score type everywhere.** `experiment.score` is a keyed `pandas.Series` for every backend; consumers stop caring which backend produced it.
- **Per-unit R² for sklearn.** Multi-output experiments now surface one R² per treated unit instead of sklearn's uniform average that quietly collapsed them.
- **Titles key on data, not backend identity.** `format_r2_score` centralizes the "Bayesian $R^2$ (std = …)" vs plain "$R^2$" decision on dispersion presence, deleting the duplicated `extract → assert → branch → round` block from ITS, PITS, RD, RK, and SC. This is exactly the discriminator-unification flagged as a cheap follow-up on #1053, now done: the ITS/PITS `isinstance(score, pd.Series)` title branch is gone, replaced by the dispersion-keyed formatter.
- **Honest uncertainty.** Point estimates omit the std entry rather than inventing one, and the new `test_score_contract.py` locks single-output, multi-output, and dispersion-aware formatting.
- **−82/+189 but the net library logic shrinks;** the additions are docstrings, the contract doc, and the new test file.

The direction is right: this is the score-shaped half of the same "uncertainty rendering keyed on data, not backend" principle that #1036 and #1053 established for containers and plotting.

drbenvincent:
Thanks for the review — addressed all three observations in commit `1cca243c` (rebased onto current `main` first).

**1. Score-key seam hardening (`extract_r2_score`)**
- `extract_r2_score()` now raises a descriptive `ValueError` when `unit_{i}_r2` is missing, instead of leaking a bare `KeyError` at plot time.
- Added `test_extract_r2_score_requires_unit_key` in `test_score_contract.py`.

**2. Sklearn adapter `score()` keyword contract**
- Documented that `SklearnModelAdapter.score()` forwards kwargs to `sklearn.metrics.r2_score` (e.g. `sample_weight`), not the underlying estimator's `score` method.
- Reject caller-supplied `multioutput` so per-unit `raw_values` scoring stays fixed.
- Added tests for `multioutput` rejection and `sample_weight` forwarding.
- Extended the upgrade note and `prediction-contract.md` with the adapter kwargs contract.

**3. Stale StateSpace integration assertion**
- Updated `test_state_space_predict_and_score` to assert `unit_0_r2` and `unit_0_r2_std` instead of the obsolete plain `"r2"` key.

**Verification:** `prek run --all-files` and `make test-patch-cov` both passed locally.

PR #1063, opened 2026-07-22
Fix order-dependent interaction-term lookup in DiD effect_summary

Mari1988:
## Summary
- `_compute_statistics_did_ols` (used by `DifferenceInDifferences.effect_summary()` for OLS-fitted models) matched the interaction coefficient via a single concatenated substring: `f"{group}:{post_treatment}" in label`.
- `patsy` names the interaction column with whichever variable is written first in the formula, so a formula written as `post_treatment*group` produces a label like `post_treatment[T.True]:group` instead of `group:post_treatment[T.True]`, which the concatenated-string check silently fails to match.
- `DifferenceInDifferences.algorithm()` already fixed this exact bug for the `causal_impact` (point estimate) lookup in #994, by checking both variable names as independent substrings instead. This separate lookup in `reporting.py`, used only for the SE/CI/p-value calculation, was never updated — so today, a valid reversed-order formula produces a correct `causal_impact` but `effect_summary()` raises `ValueError: Could not find interaction term group:post_treatment in model`.
- Fixes it the same way #994 did: match `group_variable_name` and `post_treatment_variable_name` as independent substrings.

Found while reviewing #993/#994 for unrelated reasons — reading #994's fix surfaced that a sibling function had the same bug pattern it fixed, just never updated.

**Known limitation** (not fixed here, to keep this PR minimal and mirror #994's exact approach): this substring match could in principle produce a false positive if some *other* (non-interaction) coefficient's label happened to contain both variable names as substrings — e.g. a covariate literally named `post_treatment_group_flag`. `diff_in_diff.py`'s own `_is_treatment_interaction()` helper (used elsewhere) is more rigorous — it splits on `:` and requires exact factor-name equality. The two lookups have diverged in rigor; unifying them would be a reasonable, separate follow-up.

## Test plan
- [x] Added `test_effect_summary_ols_did_order_independent`, asserting `effect_summary()` produces identical mean/ci_lower/ci_upper/p_value for `group*post_treatment` and `post_treatment*group`
- [x] `pytest causalpy/tests/test_reporting.py -k did` — 12 passed
- [x] Full suite: 1234 passed, 5 skipped
- [x] `make test-patch-cov` — 100% patch coverage
- [x] Independently verified: reproduced the exact pre-fix crash by stashing the fix, confirmed post-fix output is byte-identical between formula orderings

### DIFF
--- a/causalpy/reporting.py
+++ b/causalpy/reporting.py
@@ -1538,18 +1538,28 @@ def _compute_statistics_did_ols(
     n, p = X_da.shape
     df = n - p
 
-    # Find the interaction term coefficient index
-    interaction_term = (
-        f"{result.group_variable_name}:{result.post_treatment_variable_name}"
-    )
+    # Find the interaction term coefficient index. Match both variable names
+    # as independent substrings rather than a single concatenated string
+    # ("group:post_treatment"): patsy names the interaction column with
+    # whichever variable is written first in the formula, so a formula
+    # written as "post_treatment*group" produces a label like
+    # "post_treatment[T.True]:group", which the concatenated-string check
+    # would silently fail to match (mirrors the fix already applied to the
+    # OLS causal_impact lookup in DifferenceInDifferences.algorithm()).
     coeff_idx = None
     for i, label in enumerate(result.labels):
-        if interaction_term in label:
+        if (
+            result.group_variable_name in label
+            and result.post_treatment_variable_name in label
+        ):
             coeff_idx = i
             break
 
     if coeff_idx is None:
-        raise ValueError(f"Could not find interaction term {interaction_term} in model")
+        raise ValueError(
+            f"Could not find interaction term between '{result.group_variable_name}' "
+            f"and '{result.post_treatment_variable_name}' in model"
+        )
 
     X = X_da
     try:
--- a/causalpy/tests/test_reporting.py
+++ b/causalpy/tests/test_reporting.py
@@ -394,6 +394,49 @@ def test_effect_summary_ols_did_residuals_are_per_observation(did_data):
 
 
 @pytest.mark.integration
+def test_effect_summary_ols_did_order_independent(did_data):
+    """``effect_summary()`` must not depend on which variable is written
+    first in the DiD interaction term.
+
+    ``_compute_statistics_did_ols`` looked up the interaction coefficient via
+    a single concatenated substring (``"group:post_treatment"``), which only
+    matches patsy's column naming when the formula writes the group variable
+    first. Writing the formula the other way round (``post_treatment*group``)
+    fits an identical model (``causal_impact`` was already order-independent,
+    fixed by #994 in ``DifferenceInDifferences.algorithm()``) but this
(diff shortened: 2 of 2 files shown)


drbenvincent:
## Automated triage

**Why:**
- Contained bug fix in `causalpy/reporting.py` (22 lines of production change) that makes the interaction-term lookup in `_compute_statistics_did_ols` order-independent, mirroring the same fix already applied in `DifferenceInDifferences.algorithm()` via PR #994.
- Only two files touched: `reporting.py` and `tests/test_reporting.py` (43 new lines of focused tests). No public API, dependency, workflow, or numerical-behaviour change.
- All CI checks pass (prek, test suite on 3.11/3.14, notebook tests, codecov/patch at 100%, readthedocs, pre-commit.ci).
- The author explicitly documents a known limitation (substring match could false-positive on a covariate containing both variable names) — a reasonable scope boundary.

**Review focus:**
1. Confirm the substring-match approach doesn't introduce a false positive for any realistic DiD formula with covariates.
2. Verify the test covers both orderings and that the output is byte-identical.
3. Consider whether the known limitation warrants a follow-up issue before merging.

**Confidence:** high

daimon-pymclabs (commented):
Thanks for this @Mari1988 — nicely scoped, and the write-up (including reproducing the pre-fix crash and confirming byte-identical output across formula orderings) makes it easy to trust. The bug is real: `_compute_statistics_did_ols` was the only remaining order-dependent interaction lookup, so a `post_treatment*group` formula gave a correct `causal_impact` but crashed in `effect_summary()`. The fix and the added regression test both look correct, and 100% patch coverage is appreciated.

One actionable suggestion (optional, non-blocking): rather than mirror #994's older independent-substring approach, you can reuse the more rigorous helper that already lives on this same object — `result._is_treatment_interaction(label)`. It splits on `:`, compares exact factor names (with `C(name)` handling), and closes the exact false-positive gap you flagged in the "Known limitation" section (e.g. a covariate literally named `post_treatment_group_flag`). Since `result` is the `DifferenceInDifferences` experiment, the loop becomes:

```python
coeff_idx = next(
    (i for i, label in enumerate(result.labels)
     if result._is_treatment_interaction(label)),
    None,
)
```

That would both fix this bug and unify the two lookups in one move, rather than leaving the reporting path on the weaker matcher. If you'd rather keep this PR minimal and mirror #994 exactly, that's a reasonable call too — the current fix is correct as-is. Either way, worth a maintainer eye on the SE/CI/p-value path this touches.

derwells:
Hey @Mari1988 , ported this to the PyMC v6 migration branch, which replaces `main` around September. #1153 cites you as co-author for the changes.

Tagging as `migration:needs-port` til this gets closed out

drbenvincent:
Thanks @Mari1988 — this fix is now on the v6 migration branch via #1153 (merged), which uses `_is_treatment_interaction()` as discussed and includes you as co-author. Closing this PR since the migration branch will become `main`; no need to maintain two parallel fixes. Happy to reopen if you need this on current `main` before the migration release.

Issue #1056, opened 2026-07-22
Adding Woolridge Style ETWFE models for Staggered DiD


NathanielF:
Following the discussion here: https://github.com/pymc-labs/CausalPy/pull/1033. I'm proposing to add a Woolridge Style ETWFE saturated regression model to handle Staggered DiD style cases. 

Add Wooldridge/Mundlak ETWFE estimator to StaggeredDifferenceInDifferences
## Summary
Adds an estimator="etwfe" switch to the existing staggered DiD class, implementing Wooldridge's extended two-way fixed effects estimator via the Mundlak device.

The existing BJS imputation estimator is a fit–predict–subtract pipeline: the treatment effect is a post-hoc residual, and its uncertainty is assembled by differencing posterior-predictive draws outside the model. ETWFE is the structural phrasing of the same estimand — the cohort × event-time surface τ[g,k] is a model parameter, and the aggregated ATT ⟨W, T⟩ is a pm.Deterministic inside the model, so it arrives with its own posterior. It also replaces one free dummy per unit with partially pooled intercepts plus treatment means, which is what lets it scale.

Both estimators sit behind one entry point — users reach for "staggered DiD", not for a paper name — so switching costs one keyword.

drbenvincent:
causalpy-triage-agent: Added labels: `enhancement`. This issue proposes adding new model types (Woolridge Style ETWFE) to the library.

PR #1083, opened 2026-07-27
Adding commits for ETWFE extension to staggered DiD

NathanielF:
Relates to: https://github.com/pymc-labs/CausalPy/issues/1056#issuecomment-5085143585

Add Wooldridge/Mundlak ETWFE estimator to StaggeredDifferenceInDifferences
## Summary
Adds an estimator="etwfe" switch to the existing staggered DiD class, implementing Wooldridge's extended two-way fixed effects estimator via the Mundlak device.

The existing BJS imputation estimator is a fit–predict–subtract pipeline: the treatment effect is a post-hoc residual, and its uncertainty is assembled by differencing posterior-predictive draws outside the model. ETWFE is the structural phrasing of the same estimand — the cohort × event-time surface τ[g,k] is a model parameter, and the aggregated ATT ⟨W, T⟩ is a pm.Deterministic inside the model, so it arrives with its own posterior. It also replaces one free dummy per unit with partially pooled intercepts plus treatment means, which is what lets it scale.

Both estimators sit behind one entry point — users reach for "staggered DiD", not for a paper name — so switching costs one keyword.

API
New: ETWFERegression (PyMC), att_ (in-model ATT draws / float), att_se_, tau_surface_, att_weights_, event_time_grid_, etwfe_formula_, plot_tau_surface(). att_event_time_ / att_group_time_ / data_ / hdi_prob_ keep identical schemas, so effect_summary, PreTreatmentPlaceboCheck and the maketables adapters work unchanged.

generate_staggered_did_data gains a callable treatment_effects, cohort_effect_scale, and covariates — backwards compatible, verified bit-identical output when no new argument is passed.

⚠️ Behaviour change — needs a release note
_effect_summary_staggered_did previously reported an unweighted mean over event-time rows as the headline ATT. That is not the ATT. Under staggered adoption only the earliest cohorts reach the largest event times, so late rows rest on fewer treated observations; when effects grow with event time those thin rows are also the large ones, and the summary is biased upward.

On noise-free data where the truth is exactly recoverable:

estimate	error
unweighted mean (old)	3.4200	0.2178
n_obs-weighted (new)	3.2022	0.0000
Now weighted by n_obs on both estimator paths. This changes a user-visible number, downward in the typical case. The per-event-time ATTs were never wrong — only the aggregation.

Validation
Correctness is proved algebraically rather than by MCMC, because mock_pymc_sample makes posterior recovery impossible on the default test path:

Tier 1 — on noise-free data the saturated OLS fit recovers the generator's τ[g,k] surface and the treated-cell ATT to 1e-14 or better, including with covariates, with leads, and on an unbalanced panel with a non-contiguous index. Any error in ev_idx, cohort_idx, effect_indicator, top-binning or W shows up here.
Tier 2 — noisy recovery, plus the assertion that motivates the estimator: naive single-δ TWFE misses by more than ETWFE does.
Tier 3/4 — real MCMC (not mocked): true ATT inside the 94% HDI, r̂ = 1.0023; Bayesian and OLS agree to 0.007 against a 0.25 threshold. Combined runtime 23s.
Notebook headline (σ=0.3, cohort-heterogeneous growing effects, true ATT 3.2022): naive TWFE 2.8219 (off by 4.1 SE), ETWFE 3.1876.

953 passed, 5 skipped · 37 doctests · ruff clean · interrogate 87.4% · notebook executes under the papermill runner.

Known limitations
Covariates enter additively. Wooldridge's centred-covariate × (g,k) interactions are not implemented; documented in the class Notes and notebook.
conditioning="mundlak" raises for sklearn models — D̄_unit is exactly collinear with free unit dummies, so pinv would silently drop it and return the dummy fit.
Under Mundlak, g_t is identified only by its prior (D̄_time lies in the span of the time effects) and must not be interpreted. The ATT is unaffected. Documented in the class Notes and flagged in the notebook.


daimon-pymclabs (commented):
## Review

I reviewed this across `staggered_did.py`, `pymc_models.py`, `reporting.py`, the data generator, tests, and the adapters. No MAJOR correctness bugs. Summary below, with a couple of things worth addressing before merge (none blocking).

### The ATT aggregation change is correct

The imputation ATT correction lives in `reporting.py` (`_treated_weighted_mean`, reporting.py:350-388). `n_obs` per event-time cell is `int(e_mask.sum())`, the count of treated observations at that event time. Because each per-event-time `att` is itself the mean of `tau` over the observations in that cell, an `n_obs`-weighted mean over event times collapses exactly to the simple mean of `tau` over all treated observations, i.e. the Callaway-Sant'Anna / Wooldridge "average over the treated" estimand. Pre- and post-treatment rows are split and weighted separately, so placebo rows never leak into the headline. The old unweighted mean was biased upward under staggered adoption exactly as the PR description states.

The three backends are internally consistent: the ETWFE in-model `att = (tau * att_weights).sum()` uses treated-cell shares `N_gk / ΣN_gk`, and the OLS scalar ATT uses the same weights.

### What checks out

- Indexing: the non-negative-index-plus-`effect_indicator` scheme avoids `-1` aliasing onto a real lead column. Top-binning, lead-window/reference handling, and unbalanced/non-contiguous panel re-indexing are all sound.
- `_ols_error_bar_se` is a genuine fix: ETWFE's `att_std` is already `√(w'Vw)` and must not be divided by `√n_obs` the way the imputation sample-SD is.
- Backward-compat for the imputation path is explicitly guarded (`_validate_imputation_arguments` hard-errors if any ETWFE-only arg is set). The only behavior change for existing users is the summary ATT switching to the weighted value, which is the intended fix.
- The `g_t` non-identification under Mundlak is correctly isolated from the ATT (it enters `mu_lin` additively and never touches `tau` or the weights).
- `generate_staggered_did_data` is bit-identical when no new argument is passed (covariates are drawn after the main loop, so the default RNG stream is untouched).

### Worth addressing before merge

**1. No fit-time guard on `att_weights` (robustness, not a live bug).** The in-model `att` Deterministic (pymc_models.py:2920) sums over the entire `(cohorts, ev)` surface, so its correctness rests entirely on `att_weights` being zero on every non-treated / unidentified cell. By construction they are (weights come from treated-cell counts only), so it's correct today, but given the in-model ATT is the headline feature, an assertion that `att_weights` are nonnegative, sum to 1, and are zero off the identified support would be cheap insurance against a future edit silently leaking prior-only `tau` into the ATT.

**2. The balanced validation panels don't actually discriminate weighted vs unweighted.** On the Tier-1/Tier-2 panels every treated `(g,k)` cell has equal size, so `test_tier1_att_equals_treated_cell_mean_tau` and the noisy-recovery tests would pass even with the wrong (unweighted) aggregation, despite docstrings implying they exercise the weighting. The real discrimination comes from `test_tier1_exact_on_unbalanced_panel`, `test_tier1_att_weights_are_treated_cell_shares`, and `test_staggered_did.py` (`test_staggered_did_effect_summary_reports_treated_weighted_att`, which asserts weighted ≈ true within 1e-8 while unweighted is off by >0.1). The claim holds across the suite, but making one balanced Tier-1 panel use unequal cohort sizes would make the headline test self-contained.

### Minor

- Priors scale to `sd_y` rather than rescaling inputs. Reasonable response to the too-narrow-intervals criticism, but covariates enter `beta` unscaled and assume O(1) inputs; worth a note or a standardization step.
- `beta` carries a vestigial length-1 `treated_units` dim; only `beta[0]` is ever used. Consider dropping the dim or documenting the length-1 assumption.
- Tier-3/4 real-MCMC tests are `@pytest.mark.slow`, so they'd drop out of PR-time coverage if CI ever adds `-m "not slow"`. A comment pinning that expectation would help, since they're the only genuine-sampler coverage of the ATT deterministic.

Nice work overall, the backend-contract consistency across PyMC / OLS / event-time surfaces is the part that could most easily have gone wrong, and it's coherent.

drbenvincent:
## Automated triage

**Why:**
- Major new feature: ETWFE estimator for staggered DiD (~6000-line diff), adding `ETWFERegression` PyMC model, `estimator="etwfe"` switch, new public API (`tau_surface_`, `plot_tau_surface()`, etc.), new test files, and a new documentation notebook.
- Includes a deliberate behaviour change: ATT aggregation now weighted by `n_obs` instead of an unweighted mean, changing user-visible output.
- Large surface area touching `pymc_models.py`, `staggered_did.py`, `reporting.py`, `simulate_data.py`, `model_adapter.py`, `maketables_adapters.py`, and tests.
- All CI checks pass (tests, notebooks, lint), but merge is blocked (likely needs review approval).

**Review focus:**
1. Verify the ETWFE specification (Mundlak device, τ[g,k] parameterisation, top-binning logic) is correct.
2. Confirm the ATT aggregation change (`n_obs` weighting) is documented and called out in release notes.
3. Check that the sklearn guard (`conditioning="mundlak"` raising for OLS) and known limitations (additive covariates, uninterpretable g_t under Mundlak) are clearly communicated.

**Confidence:** high

juanitorduz (approved):
This looks very interesting @NathanielF ! I love the notebook! From my side, I approve (maybe @drbenvincent has still some small comments, but for me we can iterate over them and address them if they are easy :) )

NathanielF:
Thanks @juanitorduz . FWIW @drbenvincent the fact that the PR is also a replication test of the existing Staggered DiD should give some confidence in the implementation. But i guess the bigger question might be around architecture and code bloat within the Staggered DiD class? There is definitely some work i could do to pull out the supporting functions into a separate module but let me know what you think.

juanitorduz:
> Thanks @juanitorduz . FWIW @drbenvincent the fact that the PR is also a replication test of the existing Staggered DiD should give some confidence in the implementation. But i guess the bigger question might be around architecture and code bloat within the Staggered DiD class? There is definitely some work i could do to pull out the supporting functions into a separate module but let me know what you think.

IMO, we can leave this for another iteration (please create an issue). At least I like merging MVP PRs, getting feedback, and then doing a more thorough design and refactor after (but this is just me XD)

Issue #930, opened 2026-05-25
Support dataframe-agnostic inputs with Narwhals


drbenvincent:
# Support dataframe-agnostic inputs with Narwhals

## Problem

CausalPy currently hard-wires pandas expectations across the public experiment APIs and internal data flow. Experiment constructors, docstrings, type hints, validation logic, dataset loaders, reporting helpers, and tests commonly assume `pd.DataFrame` inputs and outputs. That makes CausalPy harder to use with dataframe libraries such as Polars, PyArrow-backed data, or other Narwhals-supported dataframe implementations, even when the user data is already in a tabular format that could be handled generically.

## Existing issue search

I searched existing open and closed issues/PRs for `narwhals`, `polars`, `"pandas" "DataFrame"`, and `"dataframe"`. I did not find an issue or PR that tracks dataframe-library compatibility or Narwhals adoption. The closest hits were unrelated or only loosely related dataframe mentions, such as #273 about inconsistent pre/post dataframe splitting and #524 about an exploratory `InterruptedTimeSeries` refactor.

## Motivation

[Narwhals](https://github.com/narwhals-dev/narwhals) provides a lightweight compatibility layer for dataframe libraries. Using it at CausalPy's data boundaries would let users bring common dataframe types without forcing pandas as the only accepted input representation, while still allowing CausalPy to convert to pandas/numpy internally when a dependency such as patsy, statsmodels, scikit-learn, PyMC, or plotting code requires that representation.

## Current pandas-heavy surfaces

Examples found in the codebase include `BaseExperiment.data: pd.DataFrame`, experiment constructors such as `InterruptedTimeSeries`, `DifferenceInDifferences`, `RegressionDiscontinuity`, `SyntheticControl`, `PanelRegression`, `StaggeredDifferenceInDifferences`, `InstrumentalVariable`, `InversePropensityWeighting`, `PiecewiseITS`, `PrePostNEGD`, and `SyntheticDifferenceInDifferences`, plus pandas-returning helpers in `causalpy.reporting`, `causalpy.data.datasets`, and plotting-data methods.

## Proposed direction

Add Narwhals as the dataframe compatibility layer for public data inputs, starting at experiment boundaries. The initial implementation can normalize incoming dataframe-like objects through Narwhals, then convert to the representation required by the existing formula/model pipeline. This should avoid a large rewrite of modeling internals while removing the public API assumption that inputs must already be pandas dataframes.

[...]

PR #1087, opened 2026-07-27
feat: dataframe-agnostic input contract (1/3)

anevolbap:
## What this does

First of three PRs for #930. Adds the shared input contract that lets CausalPy accept dataframes from libraries other than pandas.

`causalpy/input_data.py` holds `to_pandas`, which normalizes any eager dataframe Narwhals supports (pandas, Polars, PyArrow, and others) into a pandas `DataFrame`, plus the `DataFrameLike` alias for annotating those arguments. Pandas inputs are copied, so callers keep ownership of their frame and later index renaming does not reach back into it. Inputs from other libraries carry no index, so conversion gives them a default `RangeIndex`.

`causalpy/tests/test_input_data.py` covers the contract: pandas passthrough, copy isolation in both directions, Polars, PyArrow, datetime dtype survival, lazy frames rejected, and the error message naming the offending argument.

Nothing calls `to_pandas` yet. The call sites arrive in #1088 (nine column-based experiments) and #1089 (`time_column` for the three index-based ones).

## Dependencies

`narwhals>=1.13.1` at runtime, `polars` and `pyarrow` for tests, and the regenerated `environment.yml`.

The floor is 1.13.1 because `from_native` gained `pass_through` in that release. 1.12.1 still has the older `strict` keyword, and 1.13.0 was never published to PyPI or conda-forge. Below the floor the call raises `TypeError`, which `to_pandas` catches and reports as "not a dataframe", so the pin is load-bearing rather than cosmetic.

## Notes

Rebased onto the integration branch, and #1088 and #1089 have been rebased to match. Once #1101 lands, this branch needs another rebase to reconcile the generated `environment.yml`.


drbenvincent:
## Automated triage

**Why:**
- New public API module `causalpy/input_data.py` with `to_pandas()` and `DataFrameLike` — introduces a Narwhals dependency for dataframe-agnostic input handling.
- Updates `pyproject.toml` (new `narwhals>=1.13.1` runtime dependency) and `environment.yml` — a dependency change with downstream implications.
- Adds `causalpy/tests/test_input_data.py` (91 lines) covering pandas passthrough, copy isolation, Polars, PyArrow, lazy-frame rejection.
- All CI checks pass, but the PR notes it will need rebasing once #1101 lands (environment.yml conflict).

**Review focus:**
1. Verify the Narwhals `pass_through` API is stable at the pinned `>=1.13.1` floor and works across pandas/Polars/PyArrow.
2. Confirm lazy frames are rejected with a clear error message (not a confusing traceback from deep inside Narwhals).
3. Check that the `DataFrameLike` type alias is accurate and wont cause mypy issues once type checking is enabled on this module.

**Confidence:** high

PR #1088, opened 2026-07-27
feat: accept dataframe-like inputs in column-based experiments (2/3)

anevolbap:
Second of three stacked PRs for #930. #1087 has landed, so this now targets the integration branch directly.

Routes nine constructors through `to_pandas()`: `DifferenceInDifferences`, `StaggeredDifferenceInDifferences`, `PanelRegression`, `RegressionDiscontinuity`, `RegressionKink`, `InversePropensityWeighting`, `InstrumentalVariable` (both `data` and `instruments_data`), `PrePostNEGD`, and `PiecewiseITS`. These read their time and unit information from columns, so they need no index and convert cleanly.

`PiecewiseITS` is in this group rather than with the index-based classes. Its time axis comes from the `step()`/`ramp()` column in the formula, not from the index.

## obs_ind contract

A pandas caller keeps whatever their index holds. Polars and PyArrow have no index, so conversion assigns `0..n-1`. For these nine classes `obs_ind` is row identity rather than a semantic axis, and no existing caller can regress: a frame with no index never had labels to lose. Adding an optional id column later stays purely additive. Pinned by `test_polars_input_gets_positional_obs_ind`.

## Behavior change

`to_pandas()` always copies, so `data.index.name = "obs_ind"` now lands on CausalPy's copy instead of the caller's dataframe. That in-place rename was observable before. Covered by `test_pandas_input_index_is_not_renamed` and `test_panel_regression_pandas_input_index_is_not_renamed`, the two classes whose behavior actually changed.

## Tests

13 tests in `causalpy/tests/test_dataframe_backends.py`. Each experiment is built twice, from the pandas fixture and from the same data as Polars. Content is compared with `assert_data_matches`, and `assert_model_input_matches` compares the design matrices and labels, which are a deterministic function of data and formula. Posterior draws are not compared, since sampling is mocked suite-wide.

Verified on the current integration head: 2045 passed, 18 skipped.

Stack: #1087 contract (merged), this PR, then #1089 `time_column` for the index-based classes.

### DIFF
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -16,7 +16,6 @@
 from typing import Any, Literal
 
 import numpy as np
-import pandas as pd
 import seaborn as sns
 import xarray as xr
 from matplotlib import pyplot as plt
@@ -30,6 +29,7 @@ from causalpy.custom_exceptions import (
 )
 from causalpy.experiments.model_adapter import build_coords
 from causalpy.formula_utils import build_design_matrices, build_formula_matrices
+from causalpy.input_data import DataFrameLike, to_pandas
 from causalpy.plot_utils import (
     _PosteriorPlotStyle,
     has_posterior_draws,
@@ -63,8 +63,9 @@ class DifferenceInDifferences(BaseExperiment):
 
     Parameters
     ----------
-    data : pd.DataFrame
-        A pandas dataframe.
+    data : dataframe-like
+        Any eager dataframe Narwhals supports, such as pandas, Polars, or
+        PyArrow. Converted to pandas internally.
     formula : str
         A statistical model formula.
     time_variable_name : str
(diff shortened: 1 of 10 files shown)


drbenvincent:
## Automated triage

**Why:**
- Routes nine experiment constructors through `to_pandas()` — a cross-cutting change affecting most of CausalPys experiment classes.
- Modifies 9 experiment files, adding the conversion layer that accepts dataframe-like inputs (Polars, PyArrow) alongside pandas.
- New 359-line test suite in `test_dataframe_backends.py` with dual-pandas/Polars construction and content comparison.
- All CI checks pass.

**Review focus:**
1. Confirm the `to_pandas()` copy semantics prevent index mutation from leaking back to caller dataframes (especially the `data.index.name = "obs_ind"` path).
2. Verify that the `obs_ind` contract (positional `0..n-1` for non-pandas, preserved index for pandas) is correct for all nine experiment types.
3. Check that `PiecewiseITS` routing is correct — its time axis comes from formulas, not the index.

**Confidence:** high

cetagostini:
Superseded by #1131, which cherry-picked these commits onto the transition branch with @anevolbap authorship preserved and merged them into `pymc6_and_pymcmarketing1_migration`. Thank you @anevolbap — the dataframe-like input contract and the positional `obs_ind` decision shipped as you proposed.

PR #1089, opened 2026-07-27
feat: time_column for index-based experiments (3/3)

anevolbap:
Third of three stacked PRs for #930. Based on #1088; #1087 has landed.

`InterruptedTimeSeries`, `SyntheticControl`, and `SyntheticDifferenceInDifferences` use the pandas index as the time axis: they split pre/post on `data.index < treatment_time` and validate the index type against `treatment_time`. A dataframe from another library has no index, so a converted frame would arrive with a `RangeIndex` and `treatment_time` would silently become a positional cutoff.

This PR adds an optional `time_column` argument to those three classes and a shared `to_pandas_with_time_index()` helper. When `time_column` is given, that column becomes the index. When it is omitted, the pandas index is used exactly as before, so existing code is unaffected. A non-pandas input without `time_column` raises `DataException` rather than guessing.

## Guarding the time axis

Three ways the time axis can be wrong are refused rather than accepted:

- `time_column` passed alongside a meaningful index (named, or a `DatetimeIndex`) raises, since setting the column as the index would silently drop the existing time axis. Only an unnamed `RangeIndex` counts as absent, which is what conversion from an index-less library produces.
- A `time_column` with duplicate values raises.
- A `time_column` that is not sorted raises.

The last one is worth calling out because it was reachable before this PR and silent. `SyntheticControl` and `SyntheticDifferenceInDifferences` validate only the treatment-time type, never index uniqueness or monotonicity; `InterruptedTimeSeries` validates both. So a shuffled frame produced a `datapre` index of `[3, 6, 2, 4, 7, 1, 0, 9, 8, 5]` instead of `[0..9]`, with no warning. The split is value-based, so membership stayed correct while the time order did not, which breaks anything order-dependent downstream. `time_column` is what makes this likely rather than exotic: a dataframe library with no index also carries no row-order guarantee across joins and group-bys.

The validation lives in `to_pandas_with_time_index()`, the single place the time axis is established, so it covers all three classes. It only runs on the `time_column` path, so existing pandas callers are untouched.

## Error message wording

An earlier revision of this PR also corrected `"If data.index is not DatetimeIndex, treatment_time must be pd.Timestamp."`, which fired on the branch that runs precisely because `treatment_time` already is one. That fix landed upstream first, in `date_utils.validate_treatment_time_against_index`, which also handles the timezone case, so the source change is gone from this PR. The test survives, and now checks all three classes rather than one, since the original slip was copy-pasted into each.

## Sensitivity re-fit

`PlaceboInTime` hands its factory a slice of `experiment.data`, which is already normalized: `time_column` has been moved onto the index, so the column no longer exists. Replaying the original constructor kwargs therefore raised `DataException` on every fold, and the check swallows per-fold failures, so the user got an inconclusive result rather than an error. The argument is now dropped in that factory. The other four checks re-fit from the caller's untouched `context.data` and do still need it, which is why this is fixed there rather than at the source.

## Tests

26 tests in `causalpy/tests/test_time_column.py`, covering the helper, the pandas path, the Polars path, the refusal when `time_column` is missing, each of the three guards, the message on all three classes, and the sensitivity-refit case below.

Verified on the current integration head: 2073 passed, 18 skipped; doctest sweep 39 passed, 13 skipped.

Stack: #1087 contract (merged), #1088 column-based experiments, then this PR.

### DIFF
--- a/causalpy/checks/placebo_in_time.py
+++ b/causalpy/checks/placebo_in_time.py
@@ -453,6 +453,12 @@ class PlaceboInTime:
             """Create a fresh experiment with the given treatment time."""
             kw = dict(kwargs)
             kw["treatment_time"] = treatment_time
+            # This factory receives a slice of experiment.data, which is already
+            # normalized: time_column has been moved onto the index, so the
+            # column no longer exists and replaying the argument would fail.
+            # The other checks re-fit from the caller's original data and do
+            # need it, so this is dropped here rather than at the source.
+            kw.pop("time_column", None)
             if model_template is not None:
                 kw["model"] = self._clone_model_for_fold(
                     model_template, fold_random_seed
--- a/causalpy/experiments/interrupted_time_series.py
+++ b/causalpy/experiments/interrupted_time_series.py
@@ -31,6 +31,7 @@ from causalpy.date_utils import (
 )
 from causalpy.experiments.model_adapter import build_coords
 from causalpy.formula_utils import build_design_matrices, build_formula_matrices
+from causalpy.input_data import DataFrameLike, to_pandas_with_time_index
 from causalpy.plot_utils import (
     _PosteriorPlotStyle,
     format_r2_score,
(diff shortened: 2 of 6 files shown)


drbenvincent:
## Automated triage

**Why:**
- Public API change: adds a new `time_column` parameter to `InterruptedTimeSeries`, `SyntheticControl`, and `SyntheticDifferenceInDifferences` — affects the constructor contract of three exported classes.
- New shared helper `to_pandas_with_time_index()` in `causalpy/input_data.py` (+121 lines) with index validation (duplicates, sort order).
- 358-line new test suite in `test_time_column.py` covering guards, Polars input, and error messages.
- All CI checks pass.

**Review focus:**
1. Verify the `time_column` validation raises `DataException` for all documented refusal cases (meaningful index, duplicates, unsorted).
2. Confirm existing pandas callers are unaffected — the `time_column` path should only activate when the argument is provided.
3. Check the corrected error message in all three experiment classes matches the intended wording.

**Confidence:** high

cetagostini:
Superseded by #1131, which cherry-picked these commits onto the transition branch with @anevolbap authorship preserved and merged them into `pymc6_and_pymcmarketing1_migration`. Thank you @anevolbap — the `time_column` contract shipped exactly as you designed it, including the collision and duplicate/out-of-order rejections.

Issue #645, opened 2026-01-11
`InversePropensityWeighting.plot_ate()` fails with NaN error when propensity scores are extreme


drbenvincent:
## Description

`plot_ate()` fails with `ValueError: autodetected range of [nan, nan] is not finite` when propensity scores include extreme values (0 or 1). This occurs because the inverse propensity weighting calculations produce `inf`/`NaN` values due to division by zero, which then propagate to the histogram plotting.

## Steps to Reproduce

1. Open `docs/source/notebooks/inv_prop_pymc.ipynb`
2. Run all cells through the NHEFS data section (cell 27)
3. Run `result.plot_ate(method="robust", prop_draws=10, ate_draws=500)`

## Expected Behavior

Either:
- The plot renders with a warning about extreme propensity scores, or
- A clear error message indicating the propensity scores are too extreme for the selected weighting scheme

## Actual Behavior

```
ValueError: autodetected range of [nan, nan] is not finite
```

The error occurs in `matplotlib`'s histogram when it receives NaN values. The underlying issue is in `make_robust_adjustments()` at line 141:

```python
X["i_ps"] = np.where(t == 1, (p_of_t / X["ps"]), (1 - p_of_t) / (1 - X["ps"]))
```

When `ps = 1.0`, the term `1 / (1 - ps)` produces `inf`, which propagates through weighted outcome calculations as `NaN`.

## Environment

- CausalPy version: current main branch
- Python version: 3.14
- OS: macOS

## Root Cause Analysis

The NHEFS model shows significant convergence issues ("4000 divergences after tuning"), resulting in some posterior samples with extreme propensity scores (exactly 0 or 1). The weighting methods don't guard against this:

1. `make_robust_adjustments()` - divides by `(1 - ps)` for control weights
2. `make_raw_adjustments()` - divides by `ps` directly  
3. `make_overlap_adjustments()` - similar issues

## Proposed Solution

Add propensity score clipping/trimming before weight calculation:

```python
# Clip propensity scores to avoid division by zero
eps = 1e-6
ps_clipped = np.clip(ps, eps, 1 - eps)
```

This is a common practice in propensity score methods (see Austin & Stuart, 2015). Options:
1. **Default clipping** with a small epsilon (e.g., 1e-6) and emit a warning
2. **User-configurable** `ps_clip` parameter on the experiment class
3. **Raise informative error** when extreme propensity scores are detected

[...]

PR #1098, opened 2026-07-29
fix: render clipped IPW propensity endpoints

cetagostini:
## Summary

Fixes the `plot_ate(method="robust", prop_draws=10, ate_draws=500)` rendering reported in #645, where the propensity score panel came out empty (y axis autoscaled to `1e-17`) and the outcome/ATE panels blew up to ~`1e6`.

Three distinct defects fed that plot:

- **Posterior extraction.** `az.extract(idata)["p"]` combined every posterior variable just to slice one. Extraction is now normalized on the explicitly combined `p` posterior, accepting both the version-dependent `Dataset` and named `DataArray` return forms, and raising one clear error when `p` is missing.
- **Histogram range and bar geometry.** Bins ran `np.arange(0.025, 0.99, 0.005)` while bars were drawn at `bins[:-1] + 0.025` with a fixed `width=0.04` — eight times the bin width and offset by five bins. Observations clipped to the `[1e-6, 1 - 1e-6]` endpoints fell outside the binned range entirely, so the mass responsible for the pathology was invisible. Bins now cover the full `[0, 1]` range and each bar uses its own bin's true center and width.
- **Vertical scale.** Clipping extreme scores makes the IPW weights finite but astronomically large, so the weighted pseudo-population mass exceeds the observation counts by orders of magnitude. On a linear axis that flattens the propensity score distribution to invisibility — the plot in the issue. The panel now anchors a symmetric log scale at the peak observation count, so the counts stay in the linear region and only the inflated mass above them is compressed. When the weights are well behaved every bar falls inside the linear region and the panel renders as before.

## Tests

`causalpy/tests/test_ipw_extreme_ps.py` reproduces the NHEFS setup from the affected notebook with adversarial treated=0 / control=1 propensity draws and asserts:

- the informative extreme-score `UserWarning` is raised and no `RuntimeWarning` is;
- exact finite robust histogram weights and Y(1)/Y(0)/ATE data, mirrored bar signs, endpoint bar geometry, and nondegenerate distributions;
- the propensity counts remain a visible share of the panel while the pseudo-population mass is ~100x larger;
- well-behaved weights (overlap scheme) stay entirely within the linear region;
- both ArviZ extraction container forms, and the missing-`p` error path.

Raw and overlap coverage and estimator behavior are unchanged; source changes are limited to posterior extraction and the top panel's display range, geometry, and scale.

## Not included

`docs/source/notebooks/inverse-propensity-pymc.ipynb` still carries outputs generated by the old bar geometry (`nb_execution_mode = "off"`, so the rendered docs serve the stored figures). Refreshing them needs a `--full` notebook run, which is out of reach in the environment this branch was worked in.


drbenvincent:
## Automated triage

**Why:**
- Contained bug fix affecting only IPW posterior extraction and histogram rendering in `causalpy/experiments/inverse_propensity_weighting.py` (+29/−9 lines).
- New test coverage for extreme propensity-score edges in `test_ipw_extreme_ps.py` (238 lines).
- No public API changes; no dependency or workflow modifications.
- All CI checks pass.

**Review focus:**
1. Confirm the ArviZ extraction normalization handles both `Dataset` and `DataArray` return forms correctly.
2. Verify the histogram bin geometry for clipped endpoint mass produces the expected visual result in the pseudo-population plot.
3. Check the `RuntimeWarning` suppression is appropriate and doesnt mask real issues.

**Confidence:** high

daimon-pymclabs:
Follow-up on the review focus points, plus one behavioural gap I found that the earlier commits on this branch did not close.

**The endpoint rendering fix alone did not fix the reported plot.** Reproducing #645 (NHEFS setup, a treated unit at `p = 0` and a control at `p = 1`, `method="robust"`) with the branch at 4f7e286, the top panel came back with y-limits of roughly ±2.4e6: the clipped endpoints now had bars, but they were the *only* thing visible, and the actual propensity score distribution was flattened to nothing. That is the same failure the issue screenshot shows, just with the spikes drawn in. Clipping to `[1e-6, 1 - 1e-6]` keeps the weights finite, but `p_of_t / 1e-6` is ~2.6e5, so the pseudo-population histogram outruns the observation counts by five or six orders of magnitude and a linear axis cannot show both.

640344d anchors a symmetric log scale on the panel at the peak observation count. Counts sit in the linear region, only the inflated mass above them is compressed, and when the weights are well behaved (e.g. the overlap scheme, where weights are ≤ 1) every bar falls inside the linear region and the panel renders exactly as it does today. So this degrades gracefully rather than switching between two rendering modes. There is a test pinning both halves of that.

**This is a presentation choice and I'd like a second opinion on it.** The alternatives I considered and rejected:
- Clamp the y limits to the count scale and let the pseudo-population bars overflow off-axis — silently hides the diagnostic.
- Always use a log scale — changes every IPW figure, including the healthy ones.
- Raise the clipping `eps` from 1e-6 to something like 0.01 — this would genuinely tame the weights, but it changes the estimator, not the plot, so it is a maintainer call rather than something to slip into a rendering fix.

Happy to swap the approach if you'd prefer one of those.

**On the review focus points:**
1. Extraction normalization is covered for both container forms and for the missing-`p` path.
2. Endpoint bin geometry is asserted against the true bin edges and widths.
3. There is no `RuntimeWarning` suppression anywhere in the diff — the test asserts no `RuntimeWarning` is emitted at all, which holds because `_prepare_ps` clips before the division rather than because anything is filtered.

**Left undone:** `docs/source/notebooks/inverse-propensity-pymc.ipynb` still stores outputs produced by the old bar geometry, and `nb_execution_mode = "off"` means the rendered docs serve those stored figures. Its `plot_ate` panels should be refreshed with a `--full` runner pass so the published figures match the fixed code. I could not do it here — the environment this branch was worked in has no Jupyter kernel installed, and installing one was out of scope.

Issue #1048, opened 2026-07-22
PyMC 6 migration: final QA — integration suite, behavioral baselines vs pymc 5, release prep


cetagostini:
Part of #1038. Final gate before merging the integration branch into `main`. Depends on all other child issues.

## Goal

Prove the migrated stack is behaviorally equivalent (where intended) and release-ready, then merge `pymc6_and_pymcmarketing1_migration` → `main`.

## Tasks

- [ ] Full test suite in the migration env, including integration/slow markers and doctests (CI matrix 3.12 and 3.14).
- [ ] **Numerical baseline comparison vs pymc 5**: for a fixed seed, compare a representative set of outputs (effect summaries, HDI tables, R² scores, counterfactual predictions) between current `main` (pymc 5 env) and the integration branch. Document any drift beyond MC noise — expected sources: sampler differences, arviz interval policy (must stay 0.94 HDI), draw-wise R² (must be identical).
- [ ] **pandas 3 sweep**: pandas 3.0.x arrives transitively (copy-on-write default, string-dtype changes). The suite is green, but scan `causalpy/data/simulate_data.py` and the experiments for chained-assignment patterns and dtype assumptions.
- [ ] **Deprecation cleanup**: `LevelTrendComponent` → `LevelTrend` (pymc-extras statespace; `causalpy/pymc_models.py:2335`); note the pymc-extras `BSplineBasis.infer_shape` fgraph DeprecationWarning resolves only with pymc-extras ≥0.13 (which needs pymc ≥6.1 — blocked by pymc-marketing's `<6.1` cap; document, don't force).
- [ ] Docs note on the new runtime characteristics: numba is the default pytensor backend (hard dep, noticeable first-compile latency; native crashes replace Python tracebacks in workers), and nutpie becomes pymc's default NUTS sampler *only if installed* (400 tune-step default vs pymc's 1000) — consider recommending it in the installation docs.
- [ ] Changelog/release notes: breaking-change section (Python ≥3.12, pymc ≥6.0.1, arviz ≥1.1; `idata` objects are now `xarray.DataTree`). Decide version bump (this is a major/breaking release).
- [ ] Open the final PR: `pymc6_and_pymcmarketing1_migration` → `main`.

## Acceptance criteria

- Everything green (unit, integration, doctests, notebooks, docs build, lint) on the integration branch.
- Baseline-comparison report attached to this issue.
- Final PR to `main` opened with the breaking-changes summary.

## Workflow

[...]

PR #1102, opened 2026-07-29
feat: add pinned PyMC migration baseline harness

cetagostini:
## Summary

Adds the permanent PyMC 5 → PyMC 6 numerical baseline harness for the
numerical-baseline-comparison item of #1048: a fixed-seed capture of two
representative Bayesian experiments in each stack, and a comparator that decides
migration drift against preregistered gates.

`scripts/migration_baseline/harness.py` provides two commands:

- `capture` — runs `DifferenceInDifferences` (with `LinearRegression`) and
  `SyntheticControl` (with `WeightedSumFitter`) on fixture rows serialized inside
  the harness, under a fixed protocol (4 chains, `cores=1`, 1000 tune / 1000
  draws, seed 1048, `target_accept=0.95`, `max_treedepth=12`), and serializes the
  four output classes #1048 asks for: public effect-summary tables, explicit 0.94
  HDI bounds, draw-wise R², and `mu` counterfactual predictions.
- `compare` — takes four artifacts (two per stack), checks within-stack
  determinism and cross-stack runtime isolation, applies the mean-drift gates
  relative to MC noise, and renders an issue-attachment Markdown report.

Evidence integrity is enforced rather than assumed: pinned source revisions,
clean checkouts, executing-harness SHA-256/blob/commit binding, imported-runtime
provenance, role-bound batch and capture UUIDs, and create-only output paths.

## Migration gates

1. `abs(candidate_mean - reference_mean) <= max(4 * hypot(reference_mcse, candidate_mcse), 1e-6 + 1e-4 * abs(reference_mean))`
2. `abs(candidate_mean - reference_mean) / pooled_posterior_sd <= 0.1`
3. Effect-table schema, 0.94 HDI column labels, metric selectors, dimension order,
   shape and coordinate values must match exactly.

Mutual mean-in-94%-HDI containment is reported as a diagnostic only.

## Verification

The capture path is exercised against the installed stack on every CI run:
`test_capture_reproduces_the_fixed_manifest_on_the_installed_stack` runs both
fixed scenarios end to end at a reduced posterior size and asserts the captured
fixtures, effect-table bindings, series semantics, metric selectors and
sampling-quality evidence reproduce `_scenario_manifest` exactly. That reduced
protocol is verification only — `REGISTERED_SAMPLING` is the sole protocol
`capture` uses and the sole one `compare` validates against. Without it, a drift
in the CausalPy or ArviZ APIs the harness depends on would only surface after
hours of coordinator sampling.

`_harness_identity` — the first gate of every capture — is likewise covered
against a real (throwaway) git checkout rather than only through monkeypatched
stand-ins.

## Still required from the coordinator

Generating the attachable evidence needs two separately provisioned editable
installs (a PyMC 5 prefix at `79c0a87` and a PyMC 6 prefix at `18a524a`), which
cannot be produced from CI or from a single environment. The procedure is in
`scripts/migration_baseline/README.md`: four captures under one batch UUID, then
`compare`, then attach the JSON/Markdown evidence to #1048.

The historical schema-v1 PASS and its gist are intentionally invalidated by
schema v2 and must not be relabeled as v2 evidence.


daimon-pymclabs:
The two red checks here are both the same one-line fix, not anything to do with the harness itself.

`prek` and `pre-commit.ci` are failing on `ruff-format` only (`hook id: ruff-format`, "files were modified by this hook"), against the four files in the diff. Running `ruff format scripts/migration_baseline/harness.py causalpy/tests/test_migration_baseline_harness.py` and committing the result should clear both. `prek` reports the failure well past the useful part of a very long trace log, which is why it is easy to read as something worse than it is.

`test (3.12)` is a separate story: it was cancelled at the 6h runner limit rather than failing an assertion, while `test (3.14)` passed in normal time. Worth a re-run after the format fix to see whether it reproduces, since a 3.12-only six-hour hang in a PR that adds a pinned baseline harness would be the interesting finding, and a stuck runner would be the boring one.

PR #1105, opened 2026-07-29
Fix PlaceboInTime eligibility and reproducibility

### LINKED ISSUE #875: PlaceboInTime: guard against placebo folds with insufficient pre-period
## Problem

`PlaceboInTime` derives `intervention_length` from the original experiment's data span (`data.index.max() - treatment_time`) and shifts the treatment time backwards by multiples of that length to construct each placebo fold. When `n_folds * intervention_length` approaches the size of the pre-treatment period, the earliest folds end up with very few pre-treatment observations.

For Synthetic Control in particular, that means each early fold is fitting weights on (say) 7 donors with only 12 pre-treatment rows. The resulting per-fold posterior cumulative impact has a very large `sd`, which inflates `tau` in the hierarchical status-quo model, which in turn makes the inferred null distribution wide enough to swallow even a clear real effect — producing a spurious `NOT SUPPORTED` verdict.

## Concrete repro

The standard `sc` example dataset hits this with the default `treatment_time=70` (pre-period 70 rows, post-period 30 rows):

```python
import causalpy as cp

df = cp.load_data(\"sc\")
treatment_time = 70
seed = 42

result = cp.Pipeline(
    data=df,
    steps=[
        cp.EstimateEffect(
            method=cp.SyntheticControl,
            treatment_time=treatment_time,
            control_units=[\"a\", \"b\", \"c\", \"d\", \"e\", \"f\", \"g\"],
            treated_units=[\"actual\"],
            model=cp.pymc_models.WeightedSumFitter(
                sample_kwargs={\"target_accept\": 0.95, \"random_seed\": seed},
            ),
        ),
        cp.SensitivityAnalysis(checks=[cp.checks.PlaceboInTime(n_folds=2, random_seed=seed)]),
    ],
).run()

for cr in result.sensitivity_results:
    print(cr.text)
```

Output:

```
Placebo-in-time analysis: 2 of 2 folds completed.
Hierarchical status-quo model: mu=1.51, tau=13.18.
Actual cumulative impact: -38.00. P(actual outside null) = 0.905.
NOT SUPPORTED — actual effect is within the null distribution.
Fold 1: pseudo treatment at 12 — mean=16.71, sd=30.53   ← 12 pre-rows, 7 donors
Fold 2: pseudo treatment at 41 — mean=-1.81, sd=9.18
```

The actual effect (~−38) is clearly large; the failure is driven entirely by fold 1's extreme `sd=30.53`. Bumping `n_folds` doesn't help because `intervention_length=29` means earlier folds either land before the data start (and get skipped) or land at the same `tt=12` position.

This was hit while writing the SC sensitivity walkthrough docs (#789, PR #871) — the example currently has to ship with a documented \"this is expected to fail on this dataset\" callout because there's no way to make it pass without changing the dataset or the headline analysis.

## Possible directions

Any subset of these would help — they're not mutually exclusive:

1. **Allow `intervention_length` as an explicit `PlaceboInTime` argument** so users can choose a placebo window smaller than the post-period and squeeze in more well-supported folds.
2. **Add a configurable minimum pre-period length per fold** and skip (rather than silently fit) folds that fall below it. Today the only skip rule is `MIN_FOLD_OBSERVATIONS = 3` on total observations.
3. **Warn when the earliest planned fold has very few pre-treatment rows relative to the donor count** (e.g. `pre_rows < k * n_donors` with `k` ~ 2–5), pointing at the `experiment_factory` / `intervention_length` workarounds.
4. **Document the constraint** in the `PlaceboInTime` docstring and in `docs/source/notebooks/sensitivity_checks.md`.

## Acceptance criteria

- `PlaceboInTime` exposes a way (parameter or documented `experiment_factory` recipe) to avoid early folds with insufficient pre-period.
- The standard `sc` walkthrough in `sc_pymc.ipynb` can be updated to demonstrate a clean PASS rather than a NOT SUPPORTED verdict, while keeping the headline analysis (`treatment_time=70`, full donor pool) unchanged.
- Behaviour is documented and a clear warning fires (or fold is skipped) when the configured minimum can't be met.

## Out of scope

- Changes to the headline SC fitting algorithm.
- Changes to `intervention_length` defaults for non-SC experiments (ITS already has a longer typical pre-period and is not affected in the same way).

## Related

- Surfaced in PR #871 (docs walkthrough for #789).
- Adjacent plotting work in #874.

### LINKED ISSUE #876: PlaceboInTime: end-to-end reproducibility — seed posterior predictive and unify seed surface
## Problem

`PlaceboInTime` has three independent stochastic stages, and the seed handling for them is fragmented and incomplete. As a result, even when a user passes every seed argument the API currently exposes, the printed verdict (`P(actual outside null)`) and the rendered HTML report can drift between runs.

### Current state

Looking at `causalpy/checks/placebo_in_time.py`:

| Stage | Where in code | Current seeding |
|---|---|---|
| Per-fold experiment fits | `_get_factory` → `clone_model(kw[\"model\"])` | Inherits the headline model's `sample_kwargs[\"random_seed\"]`. Reproducible, but every fold gets the *same* seed. |
| Hierarchical status-quo `pm.sample` | line 389: `pm.sample(**self.sample_kwargs)` | Only if user passes `PlaceboInTime(sample_kwargs={\"random_seed\": ...})`. Undocumented as the recommended way to seed this stage. |
| Hierarchical `pm.sample_posterior_predictive` for `theta_new` | line 399 | **Not seeded — no parameter exposed.** This is the real gap. |
| Assurance simulation `np.random.default_rng` | line 490 | Uses the constructor's `random_seed=` argument. |

The constructor docstring says \`random_seed\` is \"RNG seed for the assurance simulation\", which is technically accurate but misleading: a user who reads it as \"the seed for this check\" will still get non-reproducible \`P(actual outside null)\` because the posterior predictive call is unseeded.

### Repro

Even with everything possible seeded today:

```python
import causalpy as cp

seed = 42
df = cp.load_data(\"sc\")

result = cp.Pipeline(
    data=df,
    steps=[
        cp.EstimateEffect(
            method=cp.SyntheticControl,
            treatment_time=70,
            control_units=[\"a\", \"b\", \"c\", \"d\", \"e\", \"f\", \"g\"],
            treated_units=[\"actual\"],
            model=cp.pymc_models.WeightedSumFitter(
                sample_kwargs={\"random_seed\": seed},
            ),
        ),
        cp.SensitivityAnalysis(
            checks=[
                cp.checks.PlaceboInTime(
                    n_folds=2,
                    sample_kwargs={\"random_seed\": seed},
                    random_seed=seed,
                ),
            ],
        ),
    ],
).run()

for cr in result.sensitivity_results:
    print(cr.text)
```

Two consecutive runs of the same code produce, for example:

- Run A: \`P(actual outside null) = 0.915\`
- Run B: \`P(actual outside null) = 0.923\`

Both runs share the same per-fold means/sds and the same hierarchical \`mu\` / \`tau\`. The drift is entirely from the unseeded \`pm.sample_posterior_predictive\` on \`theta_new\`. This was hit in PR #871 while writing the SC sensitivity walkthrough docs.

## Proposed fix

Unify the seed surface so a single argument makes the *entire* check deterministic.

1. Pass an explicit seed to `pm.sample_posterior_predictive`, e.g.:

  ```python
   pp = pm.sample_posterior_predictive(
       idata, var_names=[\"theta_new\"], random_seed=self.random_seed,
   )
   ```

2. Reuse the existing constructor `random_seed` argument as the master seed for *all* stochastic stages of the check (hierarchical `pm.sample`, `pm.sample_posterior_predictive`, assurance simulation, and ideally also fanned out across fold experiments so each fold gets a distinct but reproducible seed).
3. Update the docstring so `random_seed` is described as the master seed for the whole check, not just for the assurance simulation. Document that values inside `sample_kwargs[\"random_seed\"]` (if explicitly provided) take precedence for the hierarchical fit.
4. Optional: derive per-fold seeds deterministically (e.g. `seed + fold_idx`) so folds are seeded independently rather than identically — this matters once #875 lands and folds are more numerous.

## Acceptance criteria

- Two back-to-back runs of `PlaceboInTime` with the same `random_seed` produce *identical* `CheckResult.text`, `metadata[\"p_effect_outside_null\"]`, and `metadata[\"null_samples\"]` arrays.
- Setting `random_seed` alone (without also threading a seed through `sample_kwargs`) is sufficient for full reproducibility.
- Existing API surface remains backward compatible (the assurance simulation still respects `random_seed`).
- Docstring clearly describes the seeding contract.

## Related

- Surfaced in PR #871 (docs walkthrough for #789).
- #875 — placebo-fold-too-early limitation; reproducibility helps verify fixes there too.

cetagostini:
## Summary
- Skip placebo folds that lack one full intervention window of observed pre-treatment history, with deterministic skipped-fold metadata and an inconclusive result when none remain.
- Make random selection return exact maximum feasible partial subsets rather than failing on data limitations.
- Use `random_seed` across fold fits, hierarchical sampling, posterior predictive sampling, and assurance; preserve legacy custom priors with an explicit warning and run-local metadata.

Fixes #875
Fixes #876

## Validation
- `89 focused non-slow PlaceboInTime tests` passed.
- Changed-file `prek` passed every hook.
- `test_master_seed_makes_full_placebo_run_reproducible` passed with real PyMC sampling under `slow` and `correctness` markers.
- Narrow documentation notebook validated without execution.

## Review
Focused bug and adversarial reviews converged with no residual correctness findings.


drbenvincent:
## Automated triage

**Why:**
- Major rewrite of `causalpy/checks/placebo_in_time.py` (+378/−92 lines) — core causal check logic with bug fixes for eligibility and reproducibility.
- Adds 606 lines of new tests in `test_placebo_in_time.py`, including a slow/correctness-marked test with real PyMC sampling.
- Large notebook diff (+2806/−2806 lines) requires careful verification that the output changes match the code changes.
- CI is fully green; all checks pass.

**Review focus:**
1. Verify the placebo-folding eligibility logic correctly handles edge cases (short pre-treatment history, no folds remaining).
2. Confirm the `random_seed` propagation covers all sampling paths (fold fits, hierarchical sampling, posterior predictive).
3. Spot-check the notebook output diff to ensure the visual results are consistent with the algorithmic fixes.

**Confidence:** high

daimon-pymclabs:
## Open maintainer decision: comparison-window asymmetry

While closing the #875 gap I hit a semantics question that I do not think an implementer should settle alone, so I have implemented the conservative option and am flagging the alternative here.

**The asymmetry.** Placebo windows are half-open, `[t, t + intervention_length)`, but the actual effect is summarised over *every* post-intervention observation. With the derived default `intervention_length = data.index.max() - treatment_time` these two spans differ by exactly one observation. On the `sc` example from #875: each fold's cumulative impact sums 29 observations while the actual cumulative impact sums 30. The null is therefore built from slightly shorter windows than the quantity compared against it, which biases `P(actual outside null)` upward. The same asymmetry appears, larger, when `treatment_end_time` is set, and larger still now that `intervention_length` is user-settable — with `intervention_length=10` on `sc` the actual sums 30 observations against a null built from 10-observation windows.

**What this PR does (option A, implemented).** Leave the reported `actual cumulative impact` exactly as it is, record both spans in `metadata["comparison_window"]`, and warn when they disagree by more than the one-observation half-open artefact. No existing number moves; the footgun is visible rather than silent.

**The alternative (option B, not implemented).** Sum the actual impact over `[treatment_time, treatment_time + intervention_length)` too, making the check like-for-like by construction. This is the statistically cleaner definition and would also fix the `treatment_end_time` case, but it has two costs: every existing verdict shifts slightly (the default drops one post-period observation), and the check's reported `actual cumulative impact` would then differ from the experiment's own cumulative impact, which is likely to confuse users comparing the two. It would also make the committed outputs of `interrupted-time-series-placebo-in-time-analysis.ipynb` stale; that notebook pulls its data from Our World in Data at runtime, so I could not re-execute it here to refresh them.

A third option would be to define the comparison horizon in observation counts and slice both sides positionally, which removes the half-open artefact entirely, but that rewrites the fold geometry (currently index arithmetic, so that `pd.DateOffset` and numeric indices work uniformly) and ripples through the random-selection and overlap logic.

**Recommendation:** keep option A in this PR, and open a follow-up for option B or the positional-horizon variant, since changing the reported effect of an existing check deserves its own PR and its own notebook refresh. Happy to implement either here if you would rather it land together.

PR #1106, opened 2026-07-29
Data-scale SyntheticControl observation-noise priors

### LINKED ISSUE #887: Make `y_hat` σ prior data-scaled in `WeightedSumFitter` and `SoftmaxWeightedSumFitter`
## Summary

Make the observation-noise prior in `WeightedSumFitter` and `SoftmaxWeightedSumFitter` scale automatically with the outcome's spread, instead of using a fixed `HalfNormal(sigma=1)`. The current default works well only when the outcome is on a unit-ish scale (e.g. standardised z-scores), and silently degrades sampling stability and runtime on outcomes with larger natural scales.

## Background

Both `WeightedSumFitter` and `SoftmaxWeightedSumFitter` define the likelihood as

```python
default_priors = {
    "y_hat": Prior(
        "Normal",
        sigma=Prior("HalfNormal", sigma=1, dims=["treated_units"]),
        dims=["obs_ind", "treated_units"],
    ),
}
```

The `HalfNormal(1)` prior on σ has a median of about 0.67 and a 95th percentile of about 1.96. Whenever the outcome's natural standard deviation is meaningfully larger than 1, the posterior σ is pushed into the tail of the prior. This creates prior–data tension that has several practical costs:

- **Stiff posterior geometry.** When σ is forced small relative to what the data want, the conditional posterior over the donor weights becomes a narrow ridge, which is hard for NUTS to explore.
- **NUTS warmup instability.** During the early warmup phase, before the mass matrix has adapted, jittered initial points combined with a stiff posterior produce kinetic-energy overflows (`RuntimeWarning: overflow encountered in dot` in `pymc/step_methods/hmc/quadpotential.py`). PyMC handles them by rejecting the trajectory, but they are emitted to stderr and concern users.
- **Reduced effective sample size and slower wall-time.** In a controlled comparison on a real-world SC outcome, switching the σ prior from `HalfNormal(1)` to a scale that matches the outcome's spread roughly doubled `ess_bulk_min` and halved sampling wall-time, with no other change.
- **Misleading downstream summaries.** Power-curve detection rates can look artificially high when the σ posterior is held narrow by an incompatible prior, because credible intervals on the cumulative impact tighten for the wrong reason.

## Proposed change

Add a `priors_from_data(self, X, y)` method to both `WeightedSumFitter` and `SoftmaxWeightedSumFitter` that returns a `y_hat` prior whose σ scale is derived from the data, for example

```python
def priors_from_data(self, X, y):
    base = super().priors_from_data(X, y)  # preserves Dirichlet/softmax β prior
    y_std = float(np.asarray(y).std(ddof=0))
    sigma_scale = max(y_std, 1.0)  # avoid pathological zero-variance edge cases
    base["y_hat"] = Prior(
        "Normal",
        sigma=Prior("HalfNormal", sigma=sigma_scale, dims=["treated_units"]),
        dims=["obs_ind", "treated_units"],
    )
    return base
```

The exact functional form is open for discussion — alternatives include a fraction of `y_std`, a robust spread (MAD / IQR), or treating the prior as `HalfNormal(sigma=k * y_std)` with a small `k` (e.g. `0.5`) so the prior remains weakly informative rather than centred on the empirical noise. The principle is that the default should adapt to the outcome's scale rather than assume it.

For `SoftmaxWeightedSumFitter` the change is the same: only the likelihood σ is data-scaled; the softmax β logits prior is unaffected.

The library already follows this pattern elsewhere — `LinearRegression` documents the `priors_from_data` hook for exactly this purpose, so adopting it here is consistent.

## Backwards compatibility

- Users who explicitly pass `priors={"y_hat": ...}` are unaffected: the merge order in `PyMCModel.__init__` is `default_priors → priors_from_data → user-supplied`, so explicit overrides continue to win.
- Users relying on the default will see slightly different posteriors, generally better-conditioned. Worth a `CHANGELOG` note flagging it as a behaviour change. It is not a breaking API change.

## Acceptance criteria

- [ ] `WeightedSumFitter.priors_from_data` returns a data-scaled `y_hat` prior in addition to the existing Dirichlet `beta` prior.
- [ ] `SoftmaxWeightedSumFitter.priors_from_data` returns a data-scaled `y_hat` prior.
- [ ] Existing user override behaviour is preserved: passing `priors={"y_hat": ...}` to either class continues to take precedence.
- [ ] Unit tests cover the new behaviour: default prior scale changes with `y.std()`; explicit override still wins.
- [ ] Docstrings on both classes mention that the σ prior is data-scaled and explain the rationale.
- [ ] `CHANGELOG` entry under "Changes" notes the default-prior adjustment and links here.

## Out of scope

- Reparameterising the simplex β prior in `SoftmaxWeightedSumFitter`. (Separate concern; this issue is only about likelihood σ.)
- Changing the priors on `LinearRegression`, `InstrumentalVariableRegression`, or other PyMC models that already handle σ scaling differently.

---

Any thoughts on this @thomaspinder ? Sound like a good idea?

cetagostini:
## Summary
- Derive each stock `WeightedSumFitter` and `SoftmaxWeightedSumFitter` `y_hat` noise prior from its treated outcome's pre-treatment sample standard deviation: `Exponential(lam=2/s_i)`.
- Rebuild effective priors on every fit in strict `default -> data-derived -> user` order, preserving explicit `y_hat` overrides and preventing derived-prior reuse.
- Keep `SyntheticControl(auto_scale_sigma=False)` as a fit-local legacy `HalfNormal(1)` opt-out; reject non-finite, zero, or rate-invalid scales with treated-unit diagnostics.
- Add focused no-MCMC behavioral coverage for both fitters, direct fitting, multiple treated units, overrides, opt-out, cloning/reuse, invalid and dimension-only inputs, rate boundaries, and failure cleanup.

## Contributor attribution and replacement
This same-repository PR supersedes, but does **not** close or modify, contributor PR #976 from @szego. It preserves @szego's authored implementation commits (`76bd144`, `f9efa02`, `5e8858e`, and `8b67dea`) as rebased/cherry-picked commits with original authorship.

## Notebook decision
The three large notebook-output diffs from #976 are intentionally excluded. They were generated before the transition's PyMC 6 / ArviZ 1 notebook refresh and are not authoritative against this base. No MCMC notebooks were executed here; coordinator final QA should re-execute affected stock-fitter notebooks serially after source validation.

Closes #887.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ docs/source/_static/thumbnails/
 # Local scratch space (not tracked)
 .scratch/
 coverage.xml
+
+# PyTensor compilation cache
+.pytensor/
--- a/causalpy/experiments/synthetic_control.py
+++ b/causalpy/experiments/synthetic_control.py
@@ -303,7 +303,9 @@ class SyntheticControl(BaseExperiment):
 
     def algorithm(self) -> None:
         """Run the experiment algorithm: fit model, predict, and calculate causal impact."""
-        # Auto-scale the sigma prior if the user didn't customize it
+        # Auto-scale the sigma prior if the user didn't customize it.
+        # Triggers only for certain model backends because only these expose
+        # a y_hat prior.
         if (
             self.auto_scale_sigma
             and isinstance(self.model, (WeightedSumFitter, SoftmaxWeightedSumFitter))
--- a/causalpy/experiments/synthetic_control.py
+++ b/causalpy/experiments/synthetic_control.py
@@ -34,9 +34,10 @@ from causalpy.plot_utils import (
     plot_posterior_over_x,
 )
 from causalpy.pymc_models import (
+    _LEGACY_Y_HAT_PRIOR,
     PyMCModel,
-    SoftmaxWeightedSumFitter,
     WeightedSumFitter,
+    _uses_stock_y_hat_default,
 )
 from causalpy.reporting import EffectSummary
 from causalpy.utils import check_convex_hull_violation
@@ -65,14 +66,16 @@ class SyntheticControl(BaseExperiment):
         threshold trigger a ``UserWarning``. Defaults to ``0.0`` (warn on
         negatively correlated donors).
     auto_scale_sigma : bool, default True
-        If ``True`` (default) and the model is a ``WeightedSumFitter`` or
-        ``SoftmaxWeightedSumFitter`` whose ``y_hat`` prior has not been
-        explicitly overridden, the ``sigma ~ HalfNormal(1)`` default prior is
+        If ``True`` (default) and the model still carries the weighted-sum
+        fitters' stock ``y_hat`` prior, that ``sigma ~ HalfNormal(1)`` default is
         replaced by ``sigma ~ Exponential(2/s)``. The scale is computed per
         treated unit, with *s* the standard deviation of that unit's
         pre-treatment data, so units on different scales are each calibrated
-        separately. Set to ``False`` to use the original ``HalfNormal(1)``
-        default.
+        separately. Set to ``False`` to keep the original ``HalfNormal(1)``
+        default; the experiment then fits a copy of the model with that prior
+        pinned explicitly, leaving the instance you passed in untouched. A model
+        constructed with an explicit ``y_hat`` prior is never rescaled either
+        way.
     **kwargs
         Additional keyword arguments forwarded to :class:`BaseExperiment`.
 
@@ -134,6 +137,8 @@ class SyntheticControl(BaseExperiment):
         self.labels = control_units
         self.treated_units = treated_units
         self.auto_scale_sigma = auto_scale_sigma
+        if not auto_scale_sigma:
+            self._pin_legacy_sigma_prior()
         # Backend-identity check is justified here: constructor-time
         # capability validation (trust boundary), not statistical dispatch.
         if self._model_backend.is_ols and len(treated_units) > 1:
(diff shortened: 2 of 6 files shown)


drbenvincent:
## Automated triage

**Why:**
- Changes the default prior for `SyntheticControl` observation-noise from a fixed `HalfNormal(1)` to a data-scaled `Exponential(lam=2/s_i)` — a numerical behaviour change affecting a causal methods default priors.
- Touches `causalpy/pymc_models.py` and `causalpy/experiments/synthetic_control.py` — core model code and experiment wiring.
- Adds 444 lines of focused no-MCMC test coverage in `test_auto_scale_sigma.py`.
- CI has one cancelled test leg (Python 3.12); the rest pass.

**Review focus:**
1. Confirm the data-derived prior scales are reasonable across edge cases (small samples, zero-variance pre-treatment data).
2. Verify `auto_scale_sigma=False` correctly reverts to the legacy `HalfNormal(1)`.
3. Check that rebased commits from #976 preserve correct attribution and dont introduce merge artifacts.

**Confidence:** medium — the cancelled CI leg and the fact that MCMC notebooks were intentionally excluded are reasons to verify manually.

daimon-pymclabs:
### Open maintainer decision: the functional form of the scaling rule

Issue #887 deliberately left the functional form open ("The exact functional form is open for discussion"). This branch picked `sigma_i ~ Exponential(lam = 2 / s_i)`, i.e. a prior mean of `s_i / 2`, per treated unit. I have **not** changed that — it is a statistical-semantics call that belongs to a maintainer, so flagging it explicitly rather than quietly swapping it.

The two candidates on the table:

1. **`Exponential(2 / s_i)` (what is on the branch).** Fully scale-equivariant: rescaling the outcome rescales the posterior identically, and standardised data behaves the same as revenue data. Prior mass concentrated near zero with an exponential tail, so it is weakly informative in the right direction for a model whose noise should be small relative to the outcome's spread.
2. **`HalfNormal(max(s_i, 1))`, the sketch in #887.** Softer migration story — for outcomes already on a unit-ish scale (the standardised case), `max(s, 1)` collapses back to exactly the old `HalfNormal(1)`, so those users see no change at all. The cost is that it is not scale-equivariant: the floor at 1 is a unit-dependent constant, and it reintroduces the mismatch (in the too-wide direction) for genuinely small-scale outcomes.

My recommendation is to keep option 1. The whole complaint in #887 is that the default assumed a scale; a floor at 1 keeps assuming one. If you would rather trade equivariance for "standardised users see literally no change", say so and I will swap it — it is a small diff in `_data_scaled_y_hat_prior`.

### Changes I did make

- **Degenerate scales no longer raise.** A constant (or single-observation) treated pre-period made `priors_from_data` raise `ValueError` on data that fitted fine under `HalfNormal(1)`. That is a regression in a default code path, and #887 explicitly wanted the zero-variance case handled ("avoid pathological zero-variance edge cases"). It now falls back to `s_i = 1` for the affected unit only, with a `UserWarning` naming it. Non-finite outcomes still raise, because that is a data defect no prior can paper over.
- **The `auto_scale_sigma=False` opt-out is now carried by the model, not by a fit-local flag.** It used to clone the fitter inside `algorithm()`, set a private `_auto_scale_sigma` attribute, and delete it again in a `finally`, so the opt-out only held for that one fit — anything that refit or re-cloned `result.model` afterwards (sensitivity checks, placebo runs) silently got the data-scaled prior back. It is now applied at construction by fitting a copy whose `y_hat` prior is the legacy `HalfNormal(1)` pinned as an ordinary user prior, which the existing `defaults -> data-derived -> user` precedence already handles. This mirrors what the sklearn adapter does with `fit_intercept`.
- **Eligibility is no longer exact-type identity.** A model is auto-scaled when it still declares the stock `y_hat` default, so subclasses that customise other parts of the model are scaled too, while a subclass that declares its own `y_hat` default keeps it.
- **Docs.** #887 asks for a `CHANGELOG` entry; this repo keeps no `CHANGELOG` file, so the behaviour change and the `auto_scale_sigma` escape hatch are documented in the user-facing skill references (`custom_priors.md`, `synthetic_control.md`) and in the class docstrings, which now carry the rationale as well as the formula. **If release notes live somewhere I have not found, point me at it and I will add the entry there.**

### CI note

The `test (3.12)` leg on `e253d45` was cancelled after hitting the 6h job limit, hung inside `Run doctests`. The immediately preceding commit ran the same doctests with identical statistical behaviour in 17m58s, and `test (3.14)` passed on both, so this looks like a stuck runner rather than a slow prior. The new pushes will confirm.

PR #1112, opened 2026-07-30
feat: variable selection priors for SSTS covariates

anevolbap:
Closes #981. Part of #758 (P0). Targets `pymc6_and_pymcmarketing1_migration`.

## Summary

Adds variable selection priors for the covariates of `StateSpaceTimeSeries`, reusing `causalpy/variable_selection_priors.py` unchanged. This is the same factory the IV class uses.

## API

- `vs_prior_type`: `"spike_and_slab"`, `"horseshoe"` or `"normal"`, the same set the factory accepts. When set, the factory builds the `beta_exog` prior instead of the default.
- `vs_hyperparams`: optional dict passed to the factory.
- `get_inclusion_probabilities()` and `get_shrinkage_factors()` pass through to the factory and index the rows by regressor name, since the factory's tables are positional. This prepares the P6 plotting item of #758.

## Behavior

- `"normal"` is a plain `Normal(0, 1)` with no selection. It is much tighter than the class default `Normal(0, 50)` for `beta_exog`, and the docstring says so.
- Passing both `priors={"beta_exog": ...}` and `vs_prior_type` lets variable selection win with a warning, same as the IV class. The warning points at the calling line.
- Setting `vs_prior_type` without covariates raises at build time. The check cannot run earlier, because covariates come from the experiment formula and only reach the model when the experiment builds it.

## Design

Unlike the IV class, the two arguments are constructor kwargs rather than `build_model` arguments. `InterruptedTimeSeries` only feeds `X, y, coords` through the shared model adapter, so constructor configuration makes selection usable from the experiment with no adapter changes. It also validates the prior type before any data work starts.

## Known caveat

`beta_exog` point estimates shrink toward zero, because `P0` (the initial state covariance, unrelated to the P0 priority label of #758) lets the regression states drift away from the parameter. Pinning the exog `P0` entries fixes the statistics on paper but makes NUTS far slower (15+ minutes against about 2 in an earlier run), so it is not done here. Counterfactual forecasts use the smoothed states, which do recover the truth, and selection uses the inclusion-probability ranking, which is correct. The class docstring documents this.

## Tests

- Structure tests cover all three prior types. They assert API shape, dims and coords only, since the suite mocks `pm.sample`.
- A correctness-marked test (`make test-correctness`, real NUTS) covers the selection contract from #981: 6 candidate covariates with 2 in the data-generating process, gated on the inclusion-probability ranking and separation.
[...]

Issue #982, opened 2026-07-07
Graduate a BSTS model out of FutureWarning and make it the ITS default


daimon-pymclabs:
## Graduate a BSTS model out of `FutureWarning` and make it the ITS default

Part of the CausalImpact-parity effort (#758, P0/P1).

### Problem

CausalPy currently has two BSTS-flavoured models in `causalpy/pymc_models.py`, both carrying a `FutureWarning` that they are experimental and "not recommended for production use":

- `StateSpaceTimeSeries` — true state-space model via `pymc_extras.statespace.structural` (local level/trend + frequency seasonality). Architecturally closest to CausalImpact.
- `BayesianBasisExpansionTimeSeries` — Prophet-style trend changepoints + Fourier seasonality, via `pymc-marketing`.

Neither is exported at the top level, neither is the default for `InterruptedTimeSeries`, and there is no one-liner entry point. Today a user has to know to reach into `cp.pymc_models`, pick one of two experimental classes, write a patsy formula, and configure `sample_kwargs` by hand. CausalImpact's whole value proposition is `CausalImpact(data, pre.period, post.period)` with sensible defaults.

### Decision to make (was left open in #758 P0)

Which model graduates. Proposed: graduate **`StateSpaceTimeSeries`** as the production BSTS model, because it is the genuine state-space form and matches Brodersen et al. most closely. `BayesianBasisExpansionTimeSeries` can remain as an accessible alternative but this issue treats the state-space model as the one that becomes the default. Confirm or override before implementation.

### What needs to be done

- [ ] Confirm the graduation target (see above).
- [ ] Fix the outstanding API inconsistencies from the closed tech-debt tracker (#570): the BSTS models use `np.ndarray` signatures where every other `PyMCModel` subclass uses `xr.DataArray`, which forces `isinstance()` branching in `InterruptedTimeSeries`. Align the interface.
- [ ] Remove the `FutureWarning` once the API is stable.
- [ ] Make the graduated model the default `model` for `InterruptedTimeSeries` when none is supplied, with data-scaled default priors (depends on #935).
- [ ] Add a convenience entry point for the common case (e.g. `cp.InterruptedTimeSeries(data, treatment_time)` auto-detecting the outcome column and using remaining columns as covariates, mirroring CausalImpact's convention).
- [ ] Export the graduated model at the top level.
- [ ] Edge-case tests: short series, no seasonality, many covariates, missing seasonality component.

### Depends on

[...]

Issue #935, opened 2026-06-03
BayesianBasisExpansionTimeSeries default priors collapse to a flat-line fit on large-magnitude series


kylejcaron:
### Summary

The default priors on BayesianBasisExpansionTimeSeries (and its sub-components LinearTrend + YearlyFourier/WeeklyFourier) assume data on smaller magnitudes. When fit on time series with magnitude ~10⁴ or larger, the posterior collapses and the counterfactual is essentially a flat line at the intercept, with no trend or seasonality picked up at all.

Nothing in the docs or warnings flags this and basically causes a silent failure.

Solutions could include target scaling or scaling priors based on the pre-intervention dataset

### Reproducer

```
import numpy as np, pandas as pd, causalpy
from causalpy.pymc_models import BayesianBasisExpansionTimeSeries
from pymc_marketing.mmm import WeeklyFourier

rng = np.random.default_rng(0)
n = 800
dates = pd.date_range("2023-01-01", periods=n, freq="D")

# realistic-scale series: trend + weekly seasonality + noise
trend = 50_000 + np.arange(n) * 20
weekly = 8_000 * np.sin(2 * np.pi * np.arange(n) / 7)
y = trend + weekly + rng.normal(0, 3_000, n)
df = pd.DataFrame({"treated": y}, index=dates)

model = BayesianBasisExpansionTimeSeries(n_order=6)
its = causalpy.InterruptedTimeSeries(
  data=df,
  treatment_time=pd.Timestamp("2024-12-01"),
  formula="treated ~ 1",
  model=model,
)
its.plot()  # → near-flat line; no trend, no weekly cycle
```

drbenvincent:
Thanks for opening this, @kylejcaron, and welcome to the project. This is a useful reproducer, especially for a first contribution.

I think this is related to #436 and technically similar to #887, but not a duplicate.

`BayesianBasisExpansionTimeSeries` has a model-specific version of the scale problem: the likelihood sigma, intercept/exogenous coefficient, trend, and Fourier priors can all be on the wrong scale for large-magnitude outcomes. It also currently bypasses the base `priors_from_data()` hook, so the fix is not identical to the synthetic-control prior-scaling work in #887.

Let's keep this issue focused on `BayesianBasisExpansionTimeSeries` and link it to #436 as part of the broader target-scaling/data-aware-priors design.

[...]

Issue #758, opened 2026-02-28
Feature parity with Google's CausalImpact


drbenvincent:
## Feature parity with Google's CausalImpact

**Context:** [Google's CausalImpact](https://google.github.io/CausalImpact/CausalImpact.html) (R package) and its [accompanying paper](https://arxiv.org/pdf/1506.00356) (Brodersen et al., 2015) define the gold standard for Bayesian time-series causal inference using structural time-series (BSTS) models. CausalPy's `InterruptedTimeSeries` and `SyntheticControl` experiments cover much of the same ground, but there are actionable gaps worth closing so that CausalPy can be a credible Python-native alternative. This issue tracks those gaps and cross-references existing issues/PRs where work is already underway.

### Scope

CausalImpact focuses on a single use-case: estimating the causal impact of an intervention on a time series using BSTS with control covariates. CausalPy is broader (DiD, RD, IV, ...), so this issue is scoped specifically to time-series causal inference: ITS, Synthetic Control, and the BSTS model classes.

### Already at parity (or better)

These CausalImpact features already have CausalPy equivalents -- no work needed:

| CausalImpact feature | CausalPy equivalent |
|---|---|
| Pre/post period specification (indices or dates) | `treatment_time` accepts `int`, `float`, or `pd.Timestamp` |
| Three-period design (intervention ends, effect persists) | `treatment_end_time` parameter, `analyze_persistence()` method (closed #559) |
| 3-panel plot (observed+counterfactual, pointwise, cumulative) | `result.plot()` produces the same 3 panels with HDI bands |
| Summary table (Average, Cumulative, CI, posterior prob) | `result.effect_summary()` returns `.table` DataFrame and `.text` prose |
| Verbal/prose report | `effect_summary().text` generates prose with average, cumulative, and relative effects |
| MCMC customization (`niter`) | `sample_kwargs` dict passed to PyMC (`draws`, `tune`, `chains`, etc.) |
| Custom model support | Inherit from `PyMCModel` or pass any sklearn `RegressorMixin`. See #740, PR #741 |
| Adjustable credible interval width (`alpha`) | `effect_summary(alpha=...)` and `get_plot_data_bayesian(hdi_prob=...)` |
| Multiple control time series as covariates | Via patsy formula (`"y ~ 1 + t + x1 + x2"`) or SC `control_units` |
| Relative (%) effects | `effect_summary(relative=True)` with HDI intervals on relative effects |

[...]

PR #1113, opened 2026-07-30
feat!: graduate StateSpaceTimeSeries

### LINKED ISSUE #696: Good introduction to BSTS
Create an introduction to BSTS with a notebook and good example on when it's better.

anevolbap:
Part of #982, of #758 (P0) and of #696. Stacked on #1112, review that one first. Targets `pymc6_and_pymcmarketing1_migration`, so until #1112 merges this diff also shows its 8 commits.

- Remove the `FutureWarning` from `StateSpaceTimeSeries`.
- `BayesianBasisExpansionTimeSeries` is unchanged. It stays experimental and keeps its `FutureWarning`, per review. A test pins that.
- Edge-case guards and tests: short series, `level_order` below 1, `seasonal_length` below 2, missing values in `y` (the Kalman filter imputes them natively, which covers the P2 item of #758 for this model), integer-index error path, and a warning when out-of-sample dates do not continue the training frequency. The missing-value test is `correctness`-marked so it runs against real NUTS: asserting that a posterior is finite says nothing against the suite's mocked `pm.sample`.
- Two fixes to what the graduated model emits. It recovers the frequency of a regularly spaced training index, which rebuilding the index from raw values was dropping, so pymc-extras no longer warns about a missing frequency on every fit. And it builds the state-space model with `verbose=False`, so construction no longer prints the pymc-extras "Model Requirements" table, which tells the reader to declare priors this class declares itself.
- New notebook `interrupted-time-series-bsts.ipynb`: trend and seasonality, then covariates, then spike-and-slab with inclusion probabilities. Added to `gallery.yaml`.
- Document the smoothed-posterior R2 caveat on `score()`.
- Release note under the 1.0.0 breaking-change section covering the removed warning and the new `seasonal_length` and `level_order` `ValueError`s.

Three items from #982 are deliberately not here, since each is a separate behavior change worth its own review: making `StateSpaceTimeSeries` the default model for `InterruptedTimeSeries`, exporting it at top level, and adding a `cp.InterruptedTimeSeries(data, treatment_time)` convenience entry point. The first depends on data-scaled priors (#935). The `np.ndarray` signature item listed in #982 is already stale: both time-series models take `xr.DataArray` today, and `InterruptedTimeSeries` dispatches on `PyMCModel` against `RegressorMixin`, not on the model class.

One gap worth naming: #758 and #982 both list "no seasonality" as an edge case to test, and no such path exists. `build_model` always adds a seasonal component, and `pymc-extras` ships no identity component, so the `seasonal_length` guard now says that plainly instead of pointing at a `seasonality_component` workaround that does not produce a seasonality-free model. A real trend-only option is a separate change if you want it.

Note for reviewers: `score()` is computed on smoothed in-sample predictions, so it reads optimistic relative to other models. Documented, not changed here.

On CI: the base was changed to `pymc6_and_pymcmarketing1_migration` after the last push, and `.github/workflows/ci.yml` runs on the default `pull_request` events only, so the test, mypy and patch-coverage lanes have not run here yet. They start with the next push, which will be the rebase after #1112 merges. Until then the numbers below are local.

On the notebook: it is honest about what the demo shows rather than overselling it. With 30 post-period days and diffuse default priors, the trend-only model does not identify the effect (95% interval includes zero, posterior probability of an increase around 0.75). Adding `x1` and `x2` tightens the interval substantially and raises that probability to about 0.87, but the effect is still not credibly non-zero at the 95% level. The plot titles report a Bayesian R² of 1 on the pre-period; the notebook explains that this comes from the Kalman smoother conditioning on the observed outcome and is not a measure of fit. The `beta_exog` posterior means sit on the true 2.0 and -1.5 while the individual 94% HDIs span zero, because the state-space level and the regressors compete for the same variation. In the selection example the two real predictors take the top two inclusion places but nothing clears 0.5, so the notebook tells the reader to use the ordering and not the level. If you would rather the notebook demonstrate a clean detection, the setup needs a change (shorter horizon, `level_order=1`, or data-scaled priors) and I will make it.

Verified locally with pymc 6.2.0, arviz 1.2.0, pymc-extras 0.14.0, rebased on the current `upstream/issue-758-ssts-variable-selection` (the lazy experiment lifecycle): 2518 passed, 20 skipped, 5 deselected; the correctness lane 5 passed; `diff-cover` against that branch covers 100% of the 10 changed executable lines; `mypy` clean; `prek` clean on every commit; doctests for `pymc_models.py` 8 passed. After the rebase the notebook calls `.fit()` on its three experiments and runs end to end under the mock runner (`runner.py --notebook`, 16 cells). Its stored outputs come from the earlier full run (`runner.py --full`, no mocks) and were not regenerated. The selection model reports rhat above 1.01, which is inherent to a spike-and-slab posterior rather than a sampling setup problem, so the notebook says so and points the reader at the inclusion probabilities; the covariate model shows 2 divergences. Sampling is not bit-reproducible across runs even with `random_seed` set, so the notebook prose avoids quoting values that drift.


cetagostini:
indeed `BayesianBasisExpansionTimeSeries ` should not be deprecated.

cetagostini:
@anevolbap I don't recognize the branch, can we point to migration branch instead?

anevolbap:
Dropped the deprecation: `BayesianBasisExpansionTimeSeries` is back to its base state (experimental, same `FutureWarning`), and only `StateSpaceTimeSeries` graduates. Title and body updated.

Also rebased onto the current #1112 head, so the conflict is gone, and the notebook now calls `.fit()` for the lazy lifecycle.

cetagostini:
@anevolbap if we are ready then let me know moving this to ready for review, so I can merge.

PR #1116, opened 2026-07-31
chore: fix the diff-cover exclude pattern for test files

### LINKED ISSUE #1115: make test-patch-cov never excludes test files from patch coverage
`DIFF_COVER_EXCLUDE ?= causalpy/tests/*` in the Makefile has no effect.

diff-cover (checked on 10.3.0, `diff_reporter.py`, `_is_path_excluded`) matches an exclude pattern against the file basename first, then the absolute path. It never sees the repo-relative path. `causalpy/tests/test_utils.py` fails both: the basename is `test_utils.py`, and the absolute path is `/…/CausalPy/causalpy/tests/test_utils.py`, which a pattern starting with `causalpy/` cannot match. A leading wildcard works, since `fnmatch` lets `*` cross slashes.

Measured on a real `coverage.xml`: with the current pattern the report covers 98 diff lines across `pymc_models.py` and two test files; with a working pattern it covers 23 lines, source only.

The result is that the number `make test-patch-cov` prints is not the one the Makefile says it computes. It can move in either direction: an uncovered `pytest.skip` branch in a test file lowers it, and a large, fully executed test diff can mask poorly covered source. This is the local milestone check from `AGENTS.md`, not a CI gate, since no workflow calls the target.

Fix is one line: `DIFF_COVER_EXCLUDE ?= */$(PACKAGE_DIR)/tests/*`. Keeping the package directory matters, because a bare `*/tests/*` matches every file when the repository itself sits under a directory named `tests`.

Separate question, not part of this: `codecov.yml` ignores only `*/conftest.py`, so the Codecov patch status counts test files as well. Happy to open that separately if you want it changed.

anevolbap:
Closes #1115.

diff-cover matches an exclude pattern against the file basename and then the absolute path, never the repo-relative path (checked on 10.3.0 and 10.4.1, `diff_reporter.py`, `_is_path_excluded`). `causalpy/tests/*` matches neither, so `make test-patch-cov` has always counted test files, and the number it prints is not the one the Makefile says it computes. Test files usually pull it up, since they run end to end, so poorly covered source can pass the threshold. `CONTRIBUTING.md` and `AGENTS.md` both already describe the gate as excluding `causalpy/tests/*`, which this makes true.

`$(CURDIR)/$(PACKAGE_DIR)/tests/*` matches the absolute path exactly. A leading-wildcard form such as `*/causalpy/tests/*` also works in the normal case, but `fnmatch` lets `*` cross slashes, so it excludes the whole diff for any checkout sitting under a matching path, and diff-cover then prints "No lines with coverage information in this diff" and exits 0 under `--fail-under`. Anchoring at `CURDIR` rules that out, and `CURDIR` is the physical path even when the checkout is entered through a symlink, so it lines up with `os.path.abspath`.

The anchored form has one limit, noted in the comment: the pattern still goes through `fnmatch`, so a checkout path containing `[`, `?` or `*` disables the exclusion. That degrades to the current behavior on `main` rather than to a gate that measures nothing, which is why I kept the anchor.

Reproduction, throwaway repo with one source file and one test file changed and a matching `coverage.xml`:

```
--exclude 'causalpy/tests/*'        Total: 6 lines, Coverage: 83%  (test file counted)
--exclude "$PWD/causalpy/tests/*"   Total: 2 lines, Coverage: 50%  (source only)
```

No CI change, since no workflow calls this target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,10 @@ PACKAGE_DIR = causalpy
 
 DIFF_COVER_COMPARE_BRANCH ?= $(shell if git show-ref --verify --quiet refs/remotes/upstream/main; then printf "upstream/main"; else printf "origin/main"; fi)
 DIFF_COVER_FAIL_UNDER ?= 96
-DIFF_COVER_EXCLUDE ?= causalpy/tests/*
+# diff-cover (10.3.0, 10.4.1) matches exclude patterns against the basename
+# and then the absolute path, never the repo-relative path. The pattern goes
+# through fnmatch, so a checkout path containing [ ? or * disables it.
+DIFF_COVER_EXCLUDE ?= $(CURDIR)/$(PACKAGE_DIR)/tests/*
 
 init: ## Install the package in editable mode
 	python -m pip install -e . --no-deps


anevolbap:
Related to what this PR is really about, which is what the patch gate measures. There are three coverage thresholds in the repo and they do not agree.

- `DIFF_COVER_FAIL_UNDER ?= 96` in the `Makefile`
- `target: 95%` for the codecov patch status in `codecov.yml`
- `fail_under = 80` in `[tool.coverage.report]`

The first two are meant to be the same thing. AGENTS.md describes `make test-patch-cov` as "approximating the remote Codecov patch check", so the local gate is one point stricter than the CI one it approximates. A patch landing at 95.5% passes CI and fails the local check. Deliberate margin, or drift? If it is deliberate it probably deserves a word next to the number, since the docs say "approximating" rather than "deliberately stricter".

The third is a different animal. Project coverage is 95.98% right now, so an 80% total floor has sixteen points of slack and would only fire after a very large regression. The codecov `project` status (`target: auto`, `threshold: 2%`) is what actually catches regressions. So `fail_under = 80` is not wrong, it is just not doing anything, while being the first number a contributor sees because it lives in `pyproject.toml`.

Not entirely inert, though: on `pymc6_and_pymcmarketing1_migration` the correctness step in `ci.yml` has to pass `--cov-fail-under=0` explicitly to stop that partial run tripping the config floor. That override is not on `main`.

No change proposed, and none of this belongs in this PR. Mainly asking which number is the real contract, so the other two can follow it or go.

drbenvincent (approved):
## Review

Approving — this correctly fixes a broken local gate.

**What it does:** `DIFF_COVER_EXCLUDE` is changed from `causalpy/tests/*` to `$(CURDIR)/$(PACKAGE_DIR)/tests/*` so diff-cover’s absolute-path `fnmatch` actually excludes test files. Verified against diff-cover 10.3.0 behavior: the old pattern never matched test paths, so `make test-patch-cov` could count test diffs and let poorly covered source pass the documented 96% gate.

**Checks:** CI green; branch up to date with `main`; no conflicts.

**Non-blocking notes:**
- Codecov patch status still counts test files — aligning `codecov.yml` would be a separate follow-up if we want remote patch to match the local gate.
- The anchored `$(CURDIR)/…` form is preferable to `*/…` (avoids `fnmatch` `*` spanning path segments).

Closes #1115.

Issue #1117, opened 2026-07-31
ci: add timeout-minutes to the test job so a wedged doctest fails fast


anevolbap:
The `test` job in `ci.yml` has no `timeout-minutes`, so it inherits the 6 hour GitHub default. A doctest that wedges burns a full runner for six hours and reports `cancelled`, which reads like an infrastructure blip rather than a stuck job.

Seen twice in a row on #1087, both times in the `Run doctests` step on 3.12. The run stalls right after `inverse_propensity_weighting.py` at 41% and emits nothing for six hours, then GitHub kills it and the log ends with `Terminate orphan process: pid (2603) (pytest)`. It passed on a fresh runner afterwards with no change to the commit.

Evidence it is runner-level rather than a specific doctest being slow: `panel_regression.py`'s doctest runs in about 10 seconds locally, the same step passes on #1088 and #1089 which actually modify that file, and #1087's own 3.14 job clears it in 11 minutes.

`timeout-minutes: 30` on that job would turn six wasted hours into a fast, clearly labelled failure. The doctest step normally finishes in about 4 minutes, and the whole job in under 20, so 30 leaves plenty of headroom.

anevolbap:
Implemented on `pymc6_and_pymcmarketing1_migration` by f90f5f95 (in #1119): the `test` job now has a job-level `timeout-minutes: 30`, with a comment pointing back here. It is not on `main` yet, so it arrives there when the migration branch merges.

Close now, or keep it open until the migration lands on `main`?

anevolbap:
Closing since it's my own issue 😅

Issue #1067, opened 2026-07-24
Remove temporary IV cores=1 workaround after PyMC fork fix


cetagostini:
## Description

`InstrumentalVariableRegression` temporarily defaults missing or `None` `sample_kwargs["cores"]` to `1` to avoid a native worker crash in PyMC's default fork multiprocessing path on macOS arm64/Python 3.14. The crash reproduces for the pure-PyMC continuous IV graph through PyMC 6.2.0/PyTensor 3.2.2.

## Upstream

- PyMC: https://github.com/pymc-devs/pymc/issues/8377
- CausalPy mitigation: https://github.com/pymc-labs/CausalPy/issues/1044

## Acceptance criteria

- Confirm the upstream fix on the affected macOS arm64/Python 3.14 continuous IV reproduction with `chains=2, cores=2`.
- Raise CausalPy's supported PyMC floor so affected releases 6.0.1 through 6.2.0 are excluded.
- Remove the IV-only default `cores=1` workaround, its temporary comment, and tests specific to the workaround while preserving normal explicit sampling kwargs behavior.
- Add or retain regression coverage appropriate to the fixed upstream behavior.

drbenvincent:
causalpy-triage-agent: Added labels: . This issue involves removing a temporary workaround and cleaning up the code once the upstream issue is resolved.

drbenvincent:
causalpy-triage-agent: Added labels: `refactor`. This issue involves removing a temporary workaround and cleaning up the code once the upstream issue is resolved.

PR #1119, opened 2026-08-01
fix: doctest sampling leak — apply the mock at its root cause, guard the cost

daimon-pymclabs:
Part of #1048.

## What this fixes

The `--doctest-modules` CI leg (`.github/workflows/ci.yml`) ran **real MCMC
sampling**. Its mock never applied. The autouse fixture `mock_sample_for_doctest`
lived in `causalpy/tests/conftest.py`, and a conftest fixture only applies to
items collected **under its own directory**. The doctests are collected from
`causalpy/*.py` (outside `tests/`), so the fixture could never reach them: every
model-fitting doctest sampled for real, multi-core by default.

Verified empirically on pytest 9.1.1 with a probe doctest asserting
`pm.sample.__name__`: under the old command it read `sample` (the real sampler),
not `mock_sample`.

`--config-file=causalpy/tests/conftest.py` was **not** the cause — but it is not
a pure no-op either. A `.py` file carries no ini config, and its only effect was
to repoint pytest's rootdir at `causalpy/tests`. It is dropped.

## The change

- **The fix.** Move the mock into a pytest plugin
  (`causalpy/tests/doctest_sampling.py`) loaded explicitly with
  `-p causalpy.tests.doctest_sampling`. Plugin hooks register globally,
  independent of the collection tree, so `pm.sample` is mocked for the doctests.
- **The proof (a check that fails if a doctest reaches the real sampler).** The
  same plugin booby-traps the real MCMC entry points (`_sample_many`,
  `_mp_sample`, `_sample_population`, `_sample_external_nuts`) to raise. If the
  mock ever stops applying, the leg fails **loudly and immediately** instead of
  sampling for real. `mock_sample` routes through `sample_prior_predictive` and
  never touches those functions, so the guard never false-fires (confirmed: the
  full doctest suite passes green under it). `causalpy/tests/test_doctest_sampling.py`
  asserts both that the mock installs and that the guard actually raises, and
  pins the guarded names so an upstream rename can't silently disarm it.
- **The guard for #1117 (separate concern).** `timeout-minutes: 30` on the
  `test` job. A wedged run now fails as `failure` in minutes rather than being
  `cancelled` at GitHub's 6h ceiling. **This is a guard, not the fix** — the mock
  is what removes the wedge; the cap only bounds the worst case. Ordered after
  the fix so the timing improvement is measurable before the cap could mask it.
- Drop `--config-file` from the doctest command in CI and the `Makefile`; remove
  the now-dead `mock_sample_for_doctest` fixture.

## Evidence

**Mock now applies.** Probe doctest `print(pm.sample.__name__)` → `mock_sample`
under the new command (was `sample`).

**No real sampling.** Full doctest suite: `37 passed, 13 skipped` under the
plugin, with the real-sampler guard armed — nothing reached the guarded entry
points.

**Wall time (local, single core; CI runs multi-core so the absolute numbers
differ, the direction does not):**

| target | before (real) | after (mock) |
| --- | --- | --- |
| full doctest leg | 178 s | 44 s |
| `causalpy/pymc_models.py` | 64 s | 8 s |
| `causalpy/experiments/regression_discontinuity.py` | 7.6 s | 1.0 s |

**Not shipped in the wheel.** The plugin lives under `causalpy/tests/`, already
excluded by `[tool.setuptools.packages.find] exclude = ["causalpy.test*"]`.
`setuptools.find_packages(exclude=["causalpy.test*", "docs*"])` returns no
`causalpy.tests*` entry, so no file under `causalpy/tests/` (plugin included) is
packaged. (`build`/`wheel` are not installable in the shared fleet env, so this
is shown via the discovery API rather than a built wheel.)

## Intermittency

The 6-hour cancellations are intermittent (PRs #1087, #1110), consistent with
real sampling making the leg long enough to hit a race under py3.12's `fork`
default; 3.14 defaults to `forkserver` and is unaffected. This PR removes the
real sampling, so the race can no longer be triggered from the doctest leg; the
timeout is the backstop. No determinism is claimed for the original hang.

## Deliberately not done (left to maintainers)

- **No `"cores": 1` sprinkled across the remaining doctests.** It is redundant
  once the mock applies, and it would edit `causalpy/pymc_models.py`, which open
  PR #1110 rewrites. Left to a maintainer.
- **IV's `cores: 1` untouched** (issue #1067 is blocked upstream).
- **Placement decision posted as a PR comment** for maintainer override.

## Beehive gates

Explore → Plan → adversarial plan review (4 lenses) → implement → adversarial
implementation review (4 lenses) → fact-check → supervise. Details in-thread.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,13 @@ jobs:
               )
           PY
       - name: Run doctests
-        run: pytest --doctest-modules --ignore=causalpy/tests/ causalpy/ --config-file=causalpy/tests/conftest.py --no-cov
+        # -p causalpy.tests.doctest_sampling loads the plugin that mocks
+        # pm.sample for doctests (an autouse fixture in causalpy/tests/conftest.py
+        # cannot reach doctests collected from causalpy/*.py) and that fails the
+        # leg fast if a doctest ever reaches the real MCMC sampler. The old
+        # --config-file=causalpy/tests/conftest.py pointed pytest's rootdir at
+        # causalpy/tests (a .py file carries no ini config), so it is dropped.
+        run: pytest --doctest-modules -p causalpy.tests.doctest_sampling --ignore=causalpy/tests/ causalpy/ --no-cov
       - name: Run extra tests
         run: pytest docs/source/.codespell/test_notebook_to_markdown.py --no-cov
       - name: Run statistical correctness tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,11 @@ jobs:
   test:
     name: Test (Python ${{ matrix.python-version }}, pandas ${{ matrix.pandas-version }})
     runs-on: ubuntu-latest
+    # Guard (issue #1117): a wedged run fails as `failure` in minutes instead of
+    # being `cancelled` at GitHub's 6h ceiling. Headroom over the observed ~21 min
+    # max for this job (incl. cold micromamba-cache solves); tune if legitimately
+    # slower runs appear.
+    timeout-minutes: 30
     permissions:
       contents: read
     defaults:
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ check-architecture: ## Verify ARCHITECTURE.md experiment inventory matches code
 	python scripts/check_architecture_inventory.py --check
 
 doctest: ## Run doctests for the causalpy module
-	python -m pytest --doctest-modules --ignore=causalpy/tests/ causalpy/ --config-file=causalpy/tests/conftest.py
+	python -m pytest --doctest-modules -p causalpy.tests.doctest_sampling --ignore=causalpy/tests/ causalpy/
 
 test: ## Run default tests with pytest
 	python -m pytest
--- a/causalpy/tests/conftest.py
+++ b/causalpy/tests/conftest.py
@@ -42,11 +42,10 @@ from causalpy.data.simulate_data import (
 
 # Try to use PyMC's testing helpers if available; otherwise, fall back to no-op fixtures
 try:  # pragma: no cover - conditional import for compatibility across PyMC versions
-    from pymc.testing import mock_sample, mock_sample_setup_and_teardown  # type: ignore
+    from pymc.testing import mock_sample_setup_and_teardown  # type: ignore
 
     _HAVE_PYMC_TESTING = True
 except Exception:  # pragma: no cover
-    mock_sample = None  # type: ignore
     mock_sample_setup_and_teardown = None  # type: ignore
     _HAVE_PYMC_TESTING = False
 
@@ -66,18 +65,6 @@ else:
     def mock_pymc_sample():  # pragma: no cover - compatibility no-op
         # No-op fixture to satisfy tests when PyMC testing helpers are unavailable
         yield
-
-
-@pytest.fixture(autouse=True)
-def mock_sample_for_doctest(request):
-    if not request.config.getoption("--doctest-modules", default=False):
-        return
-
-    if not _HAVE_PYMC_TESTING or mock_sample is None:
-        return
-    import pymc as pm
-
-    pm.sample = mock_sample
 
 
 @pytest.fixture(scope="session")
--- a/causalpy/tests/doctest_sampling.py
+++ b/causalpy/tests/doctest_sampling.py
@@ -29,7 +29,7 @@ the doctests.
 Why here and not a package-root ``causalpy/conftest.py``
 --------------------------------------------------------
 This module lives under ``causalpy/tests/``, which is excluded from the built
-wheel by ``[tool.setuptools.packages.find] exclude = ["causalpy.test*"]``. No
+wheel by ``[tool.setuptools.packages.find] exclude = ["causalpy.test*", "docs*"]``. No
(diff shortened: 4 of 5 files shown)


daimon-pymclabs:
## Placement decision (maintainer-facing) — please sanity-check or overrule

The obvious fix puts test-only sampling logic somewhere. Three options were
evaluated:

**(a) `causalpy/conftest.py` (package-root conftest) + exclude from the wheel.**
This is the canonical pytest idiom and would work (a package-root conftest is in
scope for `causalpy/*.py` doctests). Rejected because **excluding a single
`conftest.py` module from a setuptools wheel is not cleanly expressible**:
`[tool.setuptools.packages.find] exclude` operates on *packages*, not modules,
and `MANIFEST.in` only shapes the sdist, not the wheel. So we could not prove the
shim stays out of the built wheel without contortions. (It also failed under the
*old* command anyway: `--config-file` set rootdir→`causalpy/tests`, whose
`confcutdir` excludes a `causalpy/conftest.py` above it.)

**(b) a pytest plugin under `causalpy/tests/`, loaded with `-p` — chosen.**
`causalpy/tests/` is *already* excluded from the wheel
(`exclude = ["causalpy.test*"]`), so no new packaging surface and the exclusion
is trivially provable via `find_packages`. A `-p` plugin's hooks are global
(independent of the collection tree), which is exactly the property the broken
conftest fixture lacked. The `-p` flag also makes the intent visible right in the
CI command. It naturally hosts the "real sampler must not run" guard.

**(c) invoke doctests with a rootdir where `causalpy/tests/conftest.py` is in
scope.** Rejected as non-viable: an autouse fixture's scope is *directory
containment*, not rootdir. `causalpy/tests/conftest.py` can never apply to items
in `causalpy/*.py` regardless of rootdir, because those items are not under
`tests/`.

**Recommendation:** keep (b). If you'd rather have the package-root conftest (a)
for familiarity and are content to ship a tiny no-op-guarded `conftest.py` inside
the wheel (or to add a `py-modules`/build hack to strip it), say so and I'll
switch. Happy to defer.

daimon-pymclabs:
## Follow-up (not fixed here): `mock_pymc_sample` is session-scoped and leaks the mock process-wide

While root-causing the CI failure on this PR I found a pre-existing bug that is worth its own issue. **I have deliberately not changed it in this PR** — the blast radius is large and the call is the maintainers'.

### The leak

`causalpy/tests/conftest.py`:

```python
mock_pymc_sample = pytest.fixture(mock_sample_setup_and_teardown, scope="session")
```

`mock_sample_setup_and_teardown` swaps the **process-global** `pm.sample` → `pymc.testing.mock_sample`, `pm.Flat` → `pm.Normal`, `pm.HalfFlat` → `pm.HalfNormal`, and restores them at teardown. Because the fixture is **session**-scoped, teardown does not run until the whole session ends. So the first test that requests it mocks sampling for **every subsequent test in the run**, including the ~all that never asked for it. PyMC's own documented usage is `scope="function"`.

The alphabetically-first requester is `causalpy/tests/test_auto_scale_sigma.py`, so in practice almost the entire suite runs under the mock.

### This is already known, and already worked around

`causalpy/tests/test_pymc_forecast_adapter.py:41-57` exists solely to undo it, and its docstring names the leak outright:

> Undo the session-scoped `mock_pymc_sample` patch for this module. The conftest registers PyMC's `mock_sample` (prior sampling instead of MCMC) session-wide, so once any earlier test module requests it, `pm.sample` stays mocked for the rest of the run. The round-trip tests here assert effect recovery against ground truth, which is meaningless under prior sampling — restore the real sampler for this module only.

A counter-hack that has to reach in and restore the real sampler is a good sign the scope is wrong.

### It is what broke this PR

The new `test_hooks_install_and_restore_inside_the_doctest_leg` asserted `pm.sample.__name__ != "mock_sample"` after `pytest_unconfigure`. In a full-suite run `pm.sample` was *already* `mock_sample` before the test began, so the plugin correctly captured the mock as its "original" and correctly restored the mock — and the assertion failed anyway. Green when the file ran alone, red on all three CI legs. The plugin itself was never at fault.

Fixed here by asserting the round trip the plugin actually promises (restore returns the exact objects install was given), against unique sentinels, so the test no longer depends on what ran earlier.

### Why the scope is not flipped in this PR

Measured, not guessed. Changing to `scope="function"` today:

- **Hard-errors 29 tests** via two `ScopeMismatch`es — `test_ipw_get_ate.py:30` and `test_ipw_extreme_ps.py:42` — where higher-scoped fixtures depend on the mock.
- **Silently un-mocks 21 tests in `test_hdi_prob_wiring.py`** that *explicitly requested* the mock: their higher-scoped model-fitting fixtures instantiate before the function-scoped mock can apply. These would still pass while no longer testing under the mock, which is the worst outcome.
- **Converts ~77 currently-mocked tests to real MCMC**, 64 of them unmarked, most at `chains=2, cores=2`.

So the honest fix is not a one-word scope change: it needs the dependent fixtures restructured and the genuinely-needs-real-sampling tests marked, and it will move CI runtime. That is a separate PR.

### Recommendation

Open a follow-up issue to move `mock_pymc_sample` to function scope, restructure the two `ScopeMismatch` fixture chains, fix the `test_hdi_prob_wiring.py` ordering hazard, and drop the `test_pymc_forecast_adapter.py` counter-hack once the leak is gone. Flagging rather than opening it myself.

PR #1121, opened 2026-08-01
fix: PlaceboInTime single-fold tau degeneracy (spurious SUPPORTED)

daimon-pymclabs:
Part of #1048.

## Problem

`causalpy/checks/placebo_in_time.py` accepts `n_folds >= 1`. The hierarchical
status-quo (null) model derives its scale from the between-fold spread of the
placebo folds:

```python
prior_mu_scale = np.nanstd(fold_means)
if prior_mu_scale <= 0.0:
    prior_mu_scale = 1.0
```

When that spread is **unidentified**, `prior_mu_scale` falls back to a bare
`1.0`, stripping all data scaling from both the `mu_status_quo` prior
(`sigma = 5·scale·prior_mu_scale`) and the `tau_status_quo` prior
(`sigma = 2·scale·prior_mu_scale`). On a large-scale series the null `theta_new`
then collapses to a prior-driven O(1) width, so almost any actual effect reads
as "outside the null" and the verdict flips to a spurious **SUPPORTED**
(`p_effect_outside_null = 1.0`) — driven by prior width, not evidence.

The spread is unidentified in two ways:

1. **A single usable fold** — `np.nanstd([x]) == 0`. Reached either via
   `n_folds=1` directly, or via `n_folds>1` reduced to one usable fold by fold
   skips (insufficient pre-period, too few observations, fit failure).
2. **≥2 completed folds whose cumulative impacts coincide** (e.g. an
   almost-constant series) — `np.nanstd([c, c, …]) == 0`.

This is a **false-negative in a diagnostic**: a user runs a placebo check
specifically to catch a bad design, and a spuriously SUPPORTED verdict tells
them the design is sound when it is not.

## Fix

Both degenerate cases now abstain with an explicit **INCONCLUSIVE**
(`passed=None`) — mirroring the sibling `PlaceboInSpace`, which returns
`passed=None` when it lacks enough control units to characterise its null.
(Semantics choice discussed in the maintainer-decision comment below.)

1. **`run()` abstains (INCONCLUSIVE) when fewer than `MIN_USABLE_FOLDS = 2`
   folds complete**, instead of building the degenerate single-fold null.
   Covers both routes to a single usable fold with one code path.
2. **`_build_status_quo_model()` raises a private
   `_NullModelUnidentifiedError` (a `ValueError` subclass) when the between-fold
   spread is non-positive or non-finite**, removing the scale-stripping `1.0`
   fallback. This closes the residual ≥2-coincident-fold case (and all-NaN fold
   means) at the root. `run()` catches it and returns the same unified
   INCONCLUSIVE `CheckResult`, so all three degenerate cases (0 folds, 1 fold,
   ≥2 coincident folds) share one `passed=None` contract and metadata shape.

A degenerate configuration can therefore no longer surface as SUPPORTED.

## Reproduction / tests

- `test_single_fold_directly_does_not_report_supported_on_large_scale`
  (`n_folds=1`) — fails on base (`passed=True`, `p=1.0`), passes after.
- `test_skips_down_to_single_fold_does_not_report_supported_on_large_scale`
  (`n_folds=2` reduced to one usable fold by an insufficient-pre-period skip) —
  fails on base, passes after.
- `test_build_status_quo_model_raises_when_between_fold_spread_unidentified`
  and `test_identical_folds_are_inconclusive_end_to_end` — cover the residual
  ≥2-coincident-fold case.
- `test_two_folds_with_distinct_means_still_produce_a_verdict` — positive path:
  a healthy multi-fold run still builds the null and returns a boolean verdict,
  so the abstention guards cannot silently over-fire.

The status-quo MCMC is exercised by the real code path on base (not
monkeypatched); after the fix the abstention paths return before sampling, so
the single-fold tests are deterministic.

## Known residual

The identifiability guard is `np.nanstd(fold_means) > 0` (and finite), which
closes the *exact* degeneracy. A *tiny-but-positive* between-fold spread on a
large-scale series is not closed and needs a relative criterion or an
option-(b) tau prior — a modeling choice flagged for maintainers in the review
comment, not guessed here.

## User-visible semantics change

A single-usable-fold (or ≥2-coincident-fold) placebo check that previously
returned a (spurious) SUPPORTED/NOT-SUPPORTED verdict now returns INCONCLUSIVE.
Flagged here for the release-notes unit.

## Deliberately out of scope

- The #1106 sigma findings (near-zero variance, `_clone()` bypass) in
  `pymc_models.py` / `synthetic_control.py` — owned by a sibling unit. This PR
  keeps its diff to the placebo path.
- Release notes — owned by another unit (see the semantics note above).
- A fold that *completes* but yields a partly-NaN posterior (mixed
  finite/NaN `fold_means`, e.g. `[c, NaN, d]`) is a separate fold-validity
  concern. The new guard rejects the all-NaN case; the mixed case is noted for
  a follow-up.

### DIFF
--- a/causalpy/checks/placebo_in_time.py
+++ b/causalpy/checks/placebo_in_time.py
@@ -81,6 +81,16 @@ _DEFAULT_SAMPLE_KWARGS: dict[str, Any] = {
 }
 
 
+class _NullModelUnidentifiedError(ValueError):
+    """Raised when the hierarchical null's between-fold spread is unidentified.
+
+    A subclass of :class:`ValueError` so existing ``except ValueError`` callers
+    keep working.  :meth:`PlaceboInTime.run` catches it and abstains
+    (INCONCLUSIVE) rather than building a scale-free null that could report a
+    spurious verdict.
+    """
+
+
 def _is_non_positive_length(length: Any) -> bool:
     """Return whether a window length is orderable against zero and not positive.
 
@@ -814,11 +824,11 @@ class PlaceboInTime:
 
         Raises
         ------
-        ValueError
+        _NullModelUnidentifiedError
             If the between-fold spread is unidentified, i.e.
             ``np.nanstd(fold_means)`` is not positive and finite. Building the
             null in that case would collapse it to a prior-driven width and
-            could report a spurious verdict.
+            could report a spurious verdict. A subclass of ``ValueError``.
         """
         n_folds = len(fold_means)
         fold_sds = np.where(fold_sds < 1e-6, 1e-6, fold_sds)
@@ -835,7 +845,7 @@ class PlaceboInTime:
         # complete; this guards the residual case of >= 2 folds whose
         # cumulative impacts coincide (e.g. an almost-constant series).
         if not np.isfinite(prior_mu_scale) or prior_mu_scale <= 0.0:
-            raise ValueError(
+            raise _NullModelUnidentifiedError(
                 "Cannot identify the hierarchical status-quo null: the "
                 f"{n_folds} completed placebo fold(s) have no between-fold "
                 "spread in their cumulative impacts (np.nanstd(fold_means) is "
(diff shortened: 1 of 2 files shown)


daimon-pymclabs:
## Maintainer decision: what a single-usable-fold placebo check should return

This is a **published-semantics** choice — it changes the verdict a user sees —
so I'm surfacing it rather than picking silently. Both options satisfy the hard
requirement that a degenerate configuration can **never** surface as SUPPORTED.

**Option (a) — abstain (INCONCLUSIVE) when fewer than 2 usable folds.**
When only one fold completes, `run()` returns `passed=None` /
`INCONCLUSIVE` and does not build the null model.
- *Consequence:* the user is told plainly that the placebo check could not be
  characterised (the between-fold null spread is unidentified from one fold).
  No verdict — neither SUPPORTED nor NOT SUPPORTED — is issued.
- *Pros:* honest about what one fold can support; single uniform code path for
  both routes to a single fold (`n_folds=1`, and `n_folds>1` reduced by skips);
  mirrors the existing 0-fold INCONCLUSIVE branch; `fold_results` metadata is
  still returned for inspection.
- *Cons:* users who intentionally run a single fold now get no verdict.

**Option (b) — keep fitting, derive the `tau` prior scale from the fold's own
posterior SD** (the fold's `fold_sd`) instead of the `prior_mu_scale = 1.0`
fallback.
- *Consequence:* a verdict (SUPPORTED / NOT SUPPORTED) is still produced from a
  single fold, but with a data-scaled — rather than prior-driven — null width.
- *Pros:* preserves a verdict for single-fold configurations.
- *Cons:* the within-fold posterior SD is **not** the between-fold spread it is
  standing in for; it systematically understates the true null variability, so
  the null can remain over-confident and a spurious SUPPORTED is still
  reachable in a diagnostic whose entire purpose is to avoid exactly that. It
  also introduces a new, unvalidated heuristic.

**Recommendation: (a).** For a diagnostic, abstaining on genuinely
insufficient data is safer than issuing a verdict from a fabricated null. A
single fold carries no information about between-fold variability; (b) invents
some. (a) is implemented here.

**Reversibility.** The change is one guard plus a `MIN_USABLE_FOLDS = 2`
constant in `causalpy/checks/placebo_in_time.py`. Switching to (b), or lowering
the threshold, is a localized change to that constant and the `run()` branch —
no data structures or public signatures change.

Happy to switch to (b) (or a hybrid: abstain by default, opt in to single-fold
fitting) if maintainers prefer to keep a verdict for single-fold runs.

daimon-pymclabs:
## Adversarial review outcome (beehive loop)

Two rounds of parallel four-lens review (bugs / edge cases / architecture / test quality) converged. Resolved:

- **Root cause, not just the symptom.** Round 1 established the count guard (`MIN_USABLE_FOLDS = 2`) closes the single-fold case but the shared root cause is `prior_mu_scale = np.nanstd(fold_means) <= 0.0 → 1.0`, which also collapses when ≥2 completed folds have coincident cumulative impacts (e.g. a near-constant series). Now hardened in `_build_status_quo_model`.
- **Consistent contract.** Round 2 flagged that raising for the coincident-fold case (while abstaining for the single-fold case) was inconsistent — and that the sibling `PlaceboInSpace` returns `passed=None` when it lacks enough control units. `run()` now catches the internal invariant error and returns a unified INCONCLUSIVE `CheckResult` for all three degenerate cases (0 folds, 1 fold, ≥2 coincident folds).
- **Guards don't over-fire.** Added a positive-path test: a healthy two-fold large-scale run with distinct means still builds the null and returns a boolean verdict.

## Known residual (maintainer input welcome)

The identifiability guard is `np.nanstd(fold_means) > 0` (and finite). This closes the *exact* degeneracy (single fold, or truly coincident folds). It does **not** close a *tiny-but-positive* between-fold spread — e.g. fold means `5000.0000001` and `5000.0000002` on a large-scale series — where the spread is ~1e-8 and the null is still effectively prior-driven, so a spurious SUPPORTED remains reachable in principle. Fully closing that neighborhood needs a *relative* criterion (spread vs. the data / within-fold scale) or the option-(b) tau-prior discussed above, both of which are modeling choices with their own thresholds. I've deliberately not invented a threshold here; flagging it for a maintainer call. In practice this requires an already near-degenerate (almost-constant) large-scale series.

## Out of scope (noted, not fixed)

A fold that *completes* but yields a partly-NaN posterior (`fold_means = [c, NaN, d]`) is a fold-validity issue (a bad fit that should be skipped at fit time), distinct from the tau-scale degeneracy here. The new guard rejects the all-NaN case; the mixed case is left for a follow-up.

PR #1107, opened 2026-07-29
refactor: make public API signatures explicit

cetagostini:
Closes #896

## Summary
- removes undocumented public `*args` / `**kwargs` contracts across the documented/exported CausalPy surface, with explicit concrete experiment, model, dispatcher, and effect-summary signatures;
- retains exactly two narrowly documented forwarders: integrator-provided `EstimateEffect` constructors and `plot_correlations` → `seaborn.heatmap`;
- adds the AST-backed whole-public-surface inventory and expands the former plot-only structural tests into `test_public_signatures.py`;
- removes dead generic BaseExperiment `fit` / `get_plot_data` hooks, makes supported plot-data controls keyword-only, and updates all affected callers and shipped experiment references;
- broadens production numpydoc enforcement and records the signature convention in AGENTS/ARCHITECTURE.

## Survey
The refreshed issue inventory is posted at https://github.com/pymc-labs/CausalPy/issues/896#issuecomment-5116829969: 207 public callable declarations surveyed; only the two reviewed forwarders remain variadic.

## Validation
Per coordinator instruction, no project validation was intentionally run in this work unit. The coordinator should run:

```bash
$CONDA_EXE run -n CausalPy python scripts/audit_public_signatures.py
$CONDA_EXE run -n CausalPy pytest causalpy/tests/test_public_signatures.py causalpy/tests/test_pipeline.py causalpy/tests/test_instrumental_variable.py
$CONDA_EXE run -n CausalPy pytest causalpy/tests/test_integration_skl_examples.py causalpy/tests/test_integration_pymc_examples.py
$CONDA_EXE run -n CausalPy prek run --all-files
```

## Follow-up sync check
Before merge, rebase/sync with the transition base and re-run the constructor-reference comparison if #1106 / #887 has merged, so any newly introduced `SyntheticControl` constructor parameter is reflected in the shipped reference signature.

PR #1101, opened 2026-07-29
Add pandas 2.3 compatibility coverage

cetagostini:
## Summary
- Rebased onto transition `94937675`, inheriting the docs pin required by notebook/docs CI.
- Declare and exercise the supported pandas range as `pandas>=2.3,<4`, with dedicated pandas 2.3 and pandas 3 CI matrix entries.
- Normalize pandas `StringDtype` and Arrow-string inputs on copied frames at both fitted and prediction Patsy boundaries (upstream Patsy #206), preserving explicit object dtype in pandas 3 while leaving categorical, numeric, datetime, and caller-owned data unchanged.
- Preserve exact pandas timezone-aware `obs_ind` coordinates through sklearn prediction containers, then audit experiment mutation, nullable dtype, datetime, and categorical-grouping paths.

## Exact files changed
- `pyproject.toml`
- `environment.yml` (generated by coordinator `prek`; not hand-edited)
- `.github/workflows/ci.yml`
- `causalpy/formula_utils.py`
- `causalpy/experiments/diff_in_diff.py`
- `causalpy/experiments/interrupted_time_series.py`
- `causalpy/experiments/model_adapter.py`
- `causalpy/experiments/panel_regression.py`
- `causalpy/experiments/prepostnegd.py`
- `causalpy/experiments/regression_discontinuity.py`
- `causalpy/experiments/regression_kink.py`
- `causalpy/experiments/staggered_did.py`
- `causalpy/experiments/synthetic_control.py`
- `causalpy/experiments/synthetic_difference_in_differences.py`
- `causalpy/tests/test_input_validation.py` (coordinator formatting)
- `causalpy/tests/test_pandas_compat.py`
- `causalpy/tests/test_synthetic_data.py` (coordinator formatting)

## Contracts covered
- String extension inputs, including missing values, build correct Patsy fit matrices without mutating callers; the same normalization applies to out-of-sample prediction matrices.
- Categorical and nullable numeric extension columns are not coerced by the string workaround.
- Sklearn predictions retain the exact source `obs_ind` index, including timezone-aware `DatetimeIndex` metadata, so the ITS impact-alignment invariants hold for both pre- and post-treatment data.
- Experiment-owned copies protect caller frames while nullable treatments, categorical `groupby(..., observed=True)`, float event times, and timezone-aware/`NaT` datetime boundaries retain their intended behavior.

## Validation status
- Dedicated coordinator verification passed all 30 focused pandas and model-adapter tests.
[...]

PR #1097, opened 2026-07-29
Replace deprecated LevelTrend component

cetagostini:
## Summary
- Replace the deprecated `pymc_extras.statespace.structural.LevelTrendComponent` default with the PyMC-6-compatible `LevelTrend` API while preserving lazy import and per-model caching.
- Stop passing the deprecated `mode=` keyword to `PyMCStateSpace.build_statespace_graph`; the compile mode now goes to `Component.build(mode=...)`, which is what pymc-extras asks callers to do and which also carries the mode into conditional sampling and forecasting.
- Add a focused non-sampling dispatch test that proves the default calls `LevelTrend(order=level_order)` exactly once and returns the cached component on repeat access.
- Document (in `pyproject.toml`, next to the `pymc>=6.0.1,<7` pin) why the `pymc-extras` `BSplineBasis.infer_shape` `DeprecationWarning` is deliberately left visible rather than filtered.
- Preserve the existing `InstrumentalVariableRegression` `cores=1` mitigation and its #1067 upstream-blocked removal condition unchanged.

## Repaired state-space integration tests
Three `StateSpaceTimeSeries` integration tests in `causalpy/tests/test_integration_its_new_timeseries.py` guarded themselves with `import pymc_extras.statespace.structural`. In pymc-extras 0.14 that module moved to `pymc_extras.statespace.models.structural` and is only re-exported as an attribute, so the guard raised `ModuleNotFoundError` and the tests **silently skipped** — including the coverage for the very default this PR changes.

Fixing the guard surfaced three genuine migration breakages that had been hidden behind the skip:

- pandas 3 rejects `DatetimeIndex.dayofyear[:, None]`-style expansion (`ValueError: Multi-dimensional indexing ... is no longer supported`).
- pytensor 3 removed the predefined `"PyMC"` compile mode (`ValueError: No predefined mode exist for string: PyMC`).
- `test_state_space_error_conditions` still asserted the pre-xarray contract (bare ndarray `y`, `"coords must be provided"` messages). It is rewritten against the current `xr.DataArray` contract: missing `y`, missing/multiple `treated_units`, non-datetime `obs_ind`, an invalid `datetime_index` fallback, and the three out-of-sample `predict` guards.

## Validation
`cpytest causalpy/tests/test_pymc_models_compat.py` and
`cpytest causalpy/tests/test_integration_its_new_timeseries.py -m "" -W error::DeprecationWarning`
[...]

PR #1123, opened 2026-08-01
migration: re-pin the baseline harness candidate to the integration-branch head

daimon-pymclabs:
Re-pins the migration baseline harness candidate revision to the current integration-branch head. Refs #1048.

## The problem

`scripts/migration_baseline/harness.py` hard-pins the two revisions it will accept:

- PyMC 5 reference `79c0a87072fd4653bfaed1eb085f965594c7f03a`
- PyMC 6 candidate `18a524a1a8512aaa21c46e0ccddbc54501c9eb1a` (merge of #1091, 2026-07-28)

The `pymc6_and_pymcmarketing1_migration` head is now `7b3e257b4b006800f445bec6303a399ef7ec2ffc` (merge of #1120, 2026-08-01), which is **121 commits** past that candidate, touching **60 files under `causalpy/`** (`git diff --stat 18a524a1..7b3e257b -- causalpy/`: 6,720 insertions, 1,222 deletions). Merges in that range include #1096, #1099, #1100, #1101, #1102, #1103, #1097, #1098, #1105, #1106, #1107, #1108, #1087, #850, #1110 and #1120.

So the harness would only have accepted a checkout that predates almost all of the merged migration work. Any baseline captured today would have been evidence about the wrong tree — and, because the harness reports success on the revision it pins, that would not have been visibly wrong.

## The change

- `PYMC6_COMMIT` → `7b3e257b4b006800f445bec6303a399ef7ec2ffc`. `PYMC5_COMMIT` is unchanged.
- The two documents that quote the pin are moved in step: `README.md` (intro line and the `PYMC6_ROOT` worktree provisioning step) and `REPORT_TEMPLATE.md` (candidate checkout line).
- Five new tests, none of which sample:
  - `test_pinned_revisions_are_documented_consistently` (parametrized over the four documented quote sites) checks each site against `STACK_COMMITS`. The coordinator provisions its worktrees from the README and transcribes the attribution statement from the template, so a stale SHA in either document sends a real capture at the wrong tree; that failure mode is now caught in CI instead of after hours of sampling. The anchors tolerate reflowed whitespace so rewrapping a paragraph cannot masquerade as pin drift.
  - `test_pinned_revisions_are_distinct_full_shas` — the two pins are distinct, full, lowercase SHA-1s.
  - `test_capture_rejects_an_unpinned_revision_before_import` — a *clean* checkout at a superseded revision is refused before CausalPy is imported. The pre-existing dirty-checkout test returns the correct commit and never reached that branch, so this guarantee was documented but untested.
  - `test_comparator_rejects_an_artifact_from_a_superseded_pin` (parametrized over `expected_commit` / `actual_commit`) — a stale artifact cannot enter a new batch.
  - `test_superseded_pin_is_recorded_as_history_only` — the re-pinning section cannot name the current pin as its own predecessor.
- `README.md` gains a "Re-pinning the candidate revision" section and a "Host requirements" section (below).

## Why the pin is no weaker than before

Nothing in the enforcement path changed — only the pinned value.

- `capture` still resolves `git rev-parse HEAD` in `--repo-root` and raises unless it equals `STACK_COMMITS[stack]` (`harness.py`, `_capture_artifact`), still requires an empty `git status --porcelain` in the sampled checkout, and still binds the executing harness blob via `_harness_identity`.
- `_validate_artifact` still rejects any artifact whose `provenance.expected_commit` or `provenance.actual_commit` differs from the currently pinned revision, and `compare` still requires all four artifacts to bind the executing comparator's SHA-256, blob SHA-256 and commit.
- Consequence: artifacts captured at the superseded pin `18a524a1` are now *rejected* by `compare`. A stale capture cannot be silently mixed into a new batch. The re-pin is strictly narrowing, not widening — the harness accepts exactly one candidate revision before and after.
- No schema bump is needed or wanted. `ARTIFACT_SCHEMA_VERSION` / `COMPARISON_SCHEMA_VERSION` / `SCENARIO_VERSION` describe artifact structure and the fixed fixtures, none of which changed. Distinguishability of a future batch from the historical v1 evidence is already guaranteed by schema v2 rejecting v1 artifacts outright, and stale v2 artifacts are now excluded by the commit check above.
- The re-pin is only possible as a reviewed source change to a tracked constant plus two documents held in agreement by a test. It cannot be done by a flag, an environment variable or a CLI argument.

## The v1 evidence is untouched

No file under the historical evidence path is modified: the #1048 evidence comment and gist revision `ac7db2676caf0eec0ae2da46ac52e48b3b00f86a` are not referenced by this change except in the unchanged README paragraph that declares them immutable. A future run produces a NEW v2 batch alongside them.

## Compute specification for the real capture run

**The capture was NOT executed in this PR, by design.** The protocol cannot run in a CI container or agent sandbox; it needs a dedicated host. Schedule it with the following:

| Resource | Requirement |
| --- | --- |
| Environment manager | `mamba`, `micromamba` or `conda` on `PATH` |
| Environments | 2 separate prefixes: PyMC 5 / PyTensor 2 / ArviZ 0, and PyMC 6 / PyTensor 3 / ArviZ 1 |
| Checkouts | 2 clean detached worktrees: `79c0a870…` and `7b3e257b…`, each editable-installed into its own prefix |
| RAM | ≥ 8 GB available to the run |
| CPU | ≥ 4 cores |
| Disk | ~5–8 GB for the two prefixes, two worktrees and PyTensor compile caches |
| Wall time | Hours. Schedule as one long uninterrupted job |

Detail:

- **Two prefixes are mandatory.** The stacks need incompatible PyMC/PyTensor/ArviZ trees; `pip` into one environment cannot satisfy this. The harness verifies imported major versions per stack, requires the imported third-party dependency paths to sit below the active `sys.prefix` (CausalPy itself is exempt, since an editable install lives outside the prefix), and requires the active CausalPy distribution to be an editable install whose `direct_url.json` target is exactly `--repo-root`.
- **The two prefixes must otherwise match**: same platform, machine, Python version/implementation and NumPy/pandas/xarray versions, or the run is rejected as a host change rather than migration evidence.
- **CPU:** `cores=1` is a registered protocol constant, so the four chains of each model are sampled *serially*; extra cores do not shorten a capture. They are needed for the C/numba compilation steps and headroom. Do not run the four captures concurrently — the two stacks must not contend for memory.
- **Work performed:** 4 captures (`reference_first`, `reference_second`, `candidate_first`, `candidate_second`), each sampling 2 models (DiD/`LinearRegression`, SyntheticControl/`WeightedSumFitter`) at 4 chains × (1,000 tune + 1,000 draws), `target_accept=0.95`, `max_treedepth=12`, master seed 1048 — 32 serial chain runs total, plus a cold PyTensor compile in each prefix.
- **Gates:** rank-Rhat ≤ 1.01, bulk ESS ≥ 400, tail ESS ≥ 400 with the explicit `(0.05, 0.95)` probability pair, zero divergences, no tree-depth saturation, all values finite. Failing any of these invalidates the capture rather than being reported as migration drift.
- **Isolation:** clean host, no other memory-hungry work, `PYTENSOR_FLAGS=compiledir=…` pointed at a per-prefix directory so the stacks never share a compile cache.
- **Outputs:** a newly created evidence directory outside every checkout, one shared batch UUID, four artifacts plus the comparison JSON and Markdown report, attached to #1048 following `REPORT_TEMPLATE.md`.

The exact command sequence is in `scripts/migration_baseline/README.md`.

**Note for whoever schedules it:** if further behavioral change merges into `pymc6_and_pymcmarketing1_migration` before the run, `PYMC6_COMMIT` must be re-pinned again first. Merging *this* PR does not require another re-pin: it changes only the harness, its documentation and its tests, none of which can change a sampled posterior — the harness is executed from its own checkout, whose `HEAD` intentionally differs from the sampled candidate.

## Testing

`causalpy/tests/test_migration_baseline_harness.py` — 37 passed, 1 deselected (`test_capture_reproduces_the_fixed_manifest_on_the_installed_stack`, marked `slow`/`integration`; it runs a reduced 2×500 protocol and is covered by the full CI job, not by the targeted local run). No capture was executed.

### DIFF
--- a/causalpy/tests/test_migration_baseline_harness.py
+++ b/causalpy/tests/test_migration_baseline_harness.py
@@ -483,17 +483,15 @@ def test_capture_rejects_an_unpinned_revision_before_import(
     superseded_commit = "1" * 40
     assert superseded_commit not in harness.STACK_COMMITS.values()
 
-    def fake_run_git(_repo_root: Path, *arguments: str) -> str:
-        if arguments == ("rev-parse", "HEAD"):
-            return superseded_commit
-        assert arguments == ("status", "--porcelain")
-        return ""
+    git_queries: list[tuple[str, ...]] = []
+    imported_roots: list[Path] = []
 
-    def unexpected_import(_repo_root: Path) -> None:
-        pytest.fail("an unpinned revision must be rejected before importing CausalPy")
+    def fake_run_git(_repo_root: Path, *arguments: str) -> str:
+        git_queries.append(arguments)
+        return superseded_commit
 
     monkeypatch.setattr(harness, "_run_git", fake_run_git)
-    monkeypatch.setattr(harness, "_import_capture_dependencies", unexpected_import)
+    monkeypatch.setattr(harness, "_import_capture_dependencies", imported_roots.append)
 
     with pytest.raises(harness.HarnessError, match="capture requires"):
         harness._capture_artifact(
@@ -502,6 +500,9 @@ def test_capture_rejects_an_unpinned_revision_before_import(
             batch_id="00000000-0000-4000-8000-000000000000",
             capture_role="candidate_first",
         )
+
+    assert git_queries == [("rev-parse", "HEAD")]
+    assert imported_roots == []
 
 
 def test_did_capture_fixture_passes_constructor_validation_without_mcmc() -> None:
--- a/causalpy/tests/test_migration_baseline_harness.py
+++ b/causalpy/tests/test_migration_baseline_harness.py
@@ -1218,13 +1218,13 @@ DOCUMENTED_PIN_QUOTES = (
         ("pymc6",),
     ),
     (
-        "REPORT_TEMPLATE.md",
+        "REPORT_TEMPLATE.md reference checkout",
         REPORT_TEMPLATE_PATH,
         r"- Reference checkout: `([0-9a-f]{40})`",
         ("pymc5",),
     ),
     (
-        "REPORT_TEMPLATE.md",
+        "REPORT_TEMPLATE.md candidate checkout",
         REPORT_TEMPLATE_PATH,
         r"- Candidate checkout: `([0-9a-f]{40})`",
         ("pymc6",),
--- a/causalpy/tests/test_migration_baseline_harness.py
+++ b/causalpy/tests/test_migration_baseline_harness.py
@@ -470,6 +470,40 @@ def test_capture_rejects_dirty_pinned_checkout_before_import(
         )
 
 
+def test_capture_rejects_an_unpinned_revision_before_import(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path
+) -> None:
+    """A clean checkout at the wrong revision is not migration evidence.
+
+    This is what makes re-pinning ``PYMC6_COMMIT`` safe: once the pin moves,
+    the superseded worktree can no longer be captured at all, so a stale tree
+    cannot be sampled and labelled as the migration candidate.
+    """
+    harness = _load_harness_module()
+    superseded_commit = "1" * 40
+    assert superseded_commit not in harness.STACK_COMMITS.values()
+
+    def fake_run_git(_repo_root: Path, *arguments: str) -> str:
+        if arguments == ("rev-parse", "HEAD"):
+            return superseded_commit
+        assert arguments == ("status", "--porcelain")
+        return ""
+
+    def unexpected_import(_repo_root: Path) -> None:
+        pytest.fail("an unpinned revision must be rejected before importing CausalPy")
+
+    monkeypatch.setattr(harness, "_run_git", fake_run_git)
+    monkeypatch.setattr(harness, "_import_capture_dependencies", unexpected_import)
+
+    with pytest.raises(harness.HarnessError, match="capture requires"):
+        harness._capture_artifact(
+            "pymc6",
+            tmp_path,
+            batch_id="00000000-0000-4000-8000-000000000000",
+            capture_role="candidate_first",
+        )
+
+
 def test_did_capture_fixture_passes_constructor_validation_without_mcmc() -> None:
     """The pinned DiD fixture reaches real constructor validation with OLS only."""
     harness = _load_harness_module()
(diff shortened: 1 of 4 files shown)


daimon-pymclabs:
Scope decision for a maintainer: candidate pinned to the integration head (`7b3e257b`) vs the last migration PR (`436750d3` / #1108)

This PR re-pins `PYMC6_COMMIT` to `7b3e257b` — the current head of `pymc6_and_pymcmarketing1_migration` — per the requirement that the harness accept "the final integration-branch head." One scoping judgment is properly yours, because the two reasonable pins produce different numerical baselines.

**The two candidates**
- **`7b3e257b`** (this PR): integration-branch head (merge of #1120). Includes every PR merged onto the branch.
- **`436750d3`** (#1108): the last of the eight migration PRs tracked in #1038 checklist (#1097 #1098 #1101 #1102 #1105 #1106 #1107 #1108). It is a 20-commit ancestor of the head.

**What sits between them** — four PRs on the integration branch that are *not* in #1038 migration checklist: #1087 (dataframe input contract), #850 (dynamic version), #1110 (issue-758 SSTS refactor), #1120 (sigma auto-scale fix). Three of them touch code the harness actually samples (`causalpy/pymc_models.py`, `experiments/synthetic_control.py`, `input_data.py`).

**Consequence**
- Pin **head** → the baseline compares the pre-migration reference against *the exact tree proposed for `main`*. The measured delta reflects the migration **plus** those four merges. Right question if you want "does the tree we are about to ship match PyMC 5?"
- Pin **#1108** → the baseline isolates *only* the eight-PR migration delta, matching the harness "reject later features as separate changes / not migration drift" language most literally. It excludes later model-code changes that could otherwise be attributed to migration drift.

**Recommendation:** keep **head (`7b3e257b`)**. The #1048 deliverable is a baseline for the tree that ships to `main`, and the task specifies the integration-branch head. If you prefer the strict migration-only boundary instead, say so and I will move the pin to `436750d3` and align the attribution wording in `REPORT_TEMPLATE.md`. Either way the enforcement mechanism is unchanged; only the pinned value differs.

**Note:** whichever pin you choose, if further behavioral change merges into the branch before the capture is run on real hardware, `PYMC6_COMMIT` must be re-pinned again first — the harness rejects any other checkout by design.

PR #1124, opened 2026-08-02
chore: remove dead config from pyproject.toml

anevolbap:

Four entries in `pyproject.toml` that nothing uses: two dead docs dependencies, a pytest marker registered twice, and five mypy tables written in `mypy.ini` syntax that `pyproject.toml` never reads. 18 deleted lines in one file, so I kept them together rather than splitting.

Rationale for each removal is inline on the diff.

### Checked

`prek run --all-files` is green. `environment.yml` is unchanged, as expected, since it is generated from the `dev` and `test` extras. Read the Docs built this PR and passed, and I also built the docs locally as an A/B: a clean 3.12 environment installed from this branch, then the same environment with `pathlib` and `sphinx-rtd-theme` reinstalled. Both exit 0 with 103 warnings and identical warning sets.

One unrelated thing that surfaced while checking that: the skip filter in `.readthedocs.yaml` diffs against `origin/main`, not against the PR base. A pyproject-only PR based on `main` would exit 183 and skip the preview build. This one builds because its base, the integration branch, is already 109 files ahead of `main` under `causalpy/`. Harmless here, but it means whether a PR gets a docs preview depends on its base branch rather than on what the PR changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,14 +101,12 @@ docs = [
     "pymc-bart",
     "linkify-it-py",
     "myst-nb<=1.0.0",
-    "pathlib",
     "pylint",
     "sphinx",
     "sphinx-autodoc-typehints",
     "sphinx_autodoc_defaultargs",
     "labs-sphinx-theme",
     "sphinx-copybutton",
-    "sphinx-rtd-theme",
     "sphinx-sitemap",
     "statsmodels",
     "sphinxcontrib-bibtex>=2.7.0",
@@ -164,7 +162,6 @@ addopts = [
 ]
 testpaths = "causalpy/tests"
 markers = [
-    "correctness: mark as a statistical correctness test.",
     "integration: mark as an integration test.",
     "slow: mark test as slow.",
     "correctness: mark statistical correctness and reproducibility tests.",
@@ -232,21 +229,6 @@ exclude_lines = [
 [tool.mypy]
 files = "causalpy/*.py"
 exclude = "build|dist|docs|notebooks|tests|setup.py"
-
-[tool.mypy-matplotlib]
-ignore_missing_imports = true
-
-[tool.mypy-pymc]
-ignore_missing_imports = true
-
-[tool.mypy-seaborn]
-ignore_missing_imports = true
-
-[tool.mypy-sklearn]
-ignore_missing_imports = true
-
-[tool.mypy-scipy]
-ignore_missing_imports = true
 
 [tool.marimo.runtime]
 watcher_on_save = "autorun"


pyproject.toml
```diff
@@ -101,14 +101,12 @@ docs = [
     "pymc-bart",
     "linkify-it-py",
     "myst-nb<=1.0.0",
-    "pathlib",
```
anevolbap, line 104: This is the standalone PyPI backport, not the stdlib module. It resolves to 1.0.1, whose classifiers stop at Python 3.4 and include 2.6 and 2.7, while the project floor is 3.12.

Nothing else in the environment requires it, so it installs on every docs build and every Read the Docs run and is never imported.

It came in with f385bc58, the bulk move to `pyproject.toml`, rather than as a deliberate choice.

pyproject.toml
```diff
@@ -101,14 +101,12 @@ docs = [
...
     "pylint",
     "sphinx",
     "sphinx-autodoc-typehints",
     "sphinx_autodoc_defaultargs",
     "labs-sphinx-theme",
     "sphinx-copybutton",
-    "sphinx-rtd-theme",
```
anevolbap, line 111: Unused. `docs/source/conf.py` sets `html_theme = "labs_sphinx_theme"`, and a repo-wide grep finds no other reference to the rtd theme, in `extensions`, in templates, or anywhere else.

Several installed packages do name it (`sphinx-notfound-page`, `graphviz`, `ipython`, `maketables`, `sphinx_design`), but each declares it under its own doc-building extra, none of which we request, so nothing installs it transitively.

History, as far as I could check it: the dependency arrived with f385bc58 alongside `pathlib`. The earliest `docs/source/conf.py` in the log, 71fca010 from May 2023, does set `html_theme = "sphinx_rtd_theme"`, so it was genuinely in use in that era. `conf.py` later moved to `pydata_sphinx_theme` and then to `labs_sphinx_theme` in July 2024, and the dependency was never dropped.

pyproject.toml
```diff
@@ -164,7 +162,6 @@ addopts = [
 ]
 testpaths = "causalpy/tests"
 markers = [
-    "correctness: mark as a statistical correctness test.",
```
anevolbap, line 167: `correctness` was registered twice with two different descriptions. aac2d704 added the wording that stays below, and 8b0fe535 added this one about an hour later, presumably without noticing.

Kept the earlier wording, so `pytest --markers` now lists it once.

pyproject.toml
```diff
@@ -233,21 +230,6 @@ exclude_lines = [
 files = "causalpy/*.py"
 exclude = "build|dist|docs|notebooks|tests|setup.py"
 
-[tool.mypy-matplotlib]
```
anevolbap, line 236: These five tables use `mypy.ini` syntax. In `pyproject.toml`, per-module settings have to be `[[tool.mypy.overrides]]` with a `module` key, so `tool.mypy` parses to `dict_keys(['files', 'exclude'])` and mypy has never read any of them.

Checked rather than assumed: mypy output is identical before and after this deletion, `Found 123 errors in 40 files (checked 18 source files)` both ways (mypy 1.20.2 from the dev env, measured on this branch; the same run on `main` reports 131 in 38 files across 16, since the two branches differ).

Deleted rather than converted, since switching them on would change what mypy accepts. That error count is worth a separate conversation, and it now has one: #1127, covering whether mypy is meant to be a gate here at all, given that nothing in CI runs it bare and `files = "causalpy/*.py"` does not recurse into `experiments/`, `checks/` or `steps/`.

anevolbap (commented):
Inline notes explaining each removal, so the body stays short.

PR #1125, opened 2026-08-02
fix: repair and harden the UML diagram workflow

anevolbap:

Two problems in `.github/workflows/uml.yml`, so one PR.

**Stale environment.** The job pins Python 3.10, the package needs 3.11 or newer, and it installs `causalpy[docs]` from PyPI rather than the checkout. Pip backtracks to 0.5.0, from 2025-07-24, five releases back. The diagrams stayed correct anyway, so the weekly job has been green on a year-old environment.

**Token exposure.** The job runs two pip installs while holding `contents: write` with `persist-credentials: true`, so a compromised transitive dependency would execute alongside a repo-write token in `.git/config`.

Rationale for each change is inline on the diff. `prek run --all-files` is green.

### Before merge

Workflows only really test by running, and the `reviewers` input is the one part with no precedent in this repo. Could you dispatch it from this branch? A successful run will open or update a real "Update UML Diagrams" PR, so it is not side-effect free.

Targeting `main` since scheduled workflows run from the default branch. Happy to retarget behind #1048.

### DIFF
--- a/.github/workflows/uml.yml
+++ b/.github/workflows/uml.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     permissions:
-      contents: write # to push branches
+      contents: write # to push the bot branch
       pull-requests: write # to create PRs
     steps:
 
@@ -18,21 +18,20 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: main
-          persist-credentials: true # needed for git push
+          # The push-capable token is handed to create-pull-request explicitly
+          # instead of being left in .git/config while pip installs run.
+          persist-credentials: false
 
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.10"
+          python-version: "3.12"
 
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install 'causalpy[docs]'
+          pip install -e '.[docs]'
           sudo apt-get update && sudo apt-get install -y graphviz
-
-      - name: Install pylint explicitly
-        run: python -m pip install pylint
 
       - name: Verify pylint and pyreverse
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/uml.yml
```diff
@@ -10,78 +10,56 @@ jobs:
...
     steps:
 
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           ref: main
-          persist-credentials: true # needed for git push
+          # The push-capable token is handed to create-pull-request explicitly
+          # instead of being left in .git/config while pip installs run.
+          persist-credentials: false
```
anevolbap, line 23: The token now reaches only the `create-pull-request` step instead of sitting in `.git/config` for the whole job, including the pip installs.

To be clear about the limit: the job still needs `contents: write` for the action to push its branch. This narrows how long the credential is reachable, it does not remove the permission.

.github/workflows/uml.yml
```diff
@@ -10,78 +10,56 @@ jobs:
...
 
       - name: Set up Python
         uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
-          python-version: "3.10"
+          python-version: "3.12"
```
anevolbap, line 28: Was 3.10, which contradicts `requires-python = ">=3.11"`.

Picked 3.12 rather than the 3.11 floor or the 3.14 the matrix tests, because it matches `ci.yml`'s `prek` and `check-environment-yml` jobs and the `.readthedocs.yaml` docs build, which is the closest analogue to this job. Happy to move it if you would rather track the floor.

.github/workflows/uml.yml
```diff
@@ -10,78 +10,56 @@ jobs:
...
 
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install 'causalpy[docs]'
+          pip install -e '.[docs]'
```
anevolbap, line 33: Was `pip install 'causalpy[docs]'`, which fetches from PyPI. On 3.10 pip cannot resolve any current release, so it backtracks:

```
$ uv pip install --python 3.10 --dry-run causalpy
 + causalpy==0.5.0
```

The diagrams were still right because pyreverse parses the checked-out source, not the installed distribution. The weekly bot PRs (#1084, #1030, #1011) keep changing the PNGs, which a frozen 0.5.0 wheel could not do. So the job was green and correct while running on 0.5.0's docs extra rather than this repo's.

Installing the checkout editable also makes that resolution deliberate. Today it only works because the working directory precedes `site-packages` on `sys.path`.

The separate `pip install pylint` step below is dropped: the `docs` extra already provides it.

.github/workflows/uml.yml
```diff
@@ -10,78 +10,56 @@ jobs:
...
-
-          # Delete the branch locally and remotely if it exists
-          git branch -D $BRANCH_NAME 2>/dev/null || true
-          git push origin --delete $BRANCH_NAME 2>/dev/null || true
-
-          # Create and push the new branch
-          git checkout -b $BRANCH_NAME
-          git commit -m "Update UML Diagrams"
-          git push -u origin $BRANCH_NAME
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
```
anevolbap, line 47: Replaces the manual `git push` plus `gh pr create` block. Same action and same pin `contributors.yml` already uses, and it has produced merged PRs from that config (#1010, #965), so the pattern works in this repo.

Two behaviour changes worth knowing. The action skips when nothing changed, so the old detect-changes step and the git identity config are gone. And it updates an open PR in place, where the old block deleted and recreated the branch on every run.

.github/workflows/uml.yml
```diff
@@ -10,78 +10,56 @@ jobs:
...
-        env:
-          GH_TOKEN: ${{ github.token }}
+            Created automatically by the [UML workflow](https://github.com/pymc-labs/CausalPy/blob/main/.github/workflows/uml.yml)
+            from the checked-out source via `make uml`. See the
+            [logs](https://github.com/pymc-labs/CausalPy/actions/workflows/uml.yml)
+            for more details.
+          branch: update-uml-diagrams
+          delete-branch: true
+          add-paths: |
+            docs/source/_static/classes.png
```
anevolbap, line 62: Was `git add docs/source/_static/*.png`, which staged every PNG in the directory. `pyreverse -o png causalpy --output-directory docs/source/_static --ignore tests` exits 0 and writes exactly these two files, so naming them is complete and no longer sweeps up unrelated images.

anevolbap (commented):
Inline notes explaining each change, so the body stays short.

drbenvincent (approved):
## Review

Approving — well-motivated repair of a workflow that was green on a stale environment.

**What it fixes (verified in production run `30816099227`):**
- Python 3.10 + PyPI `causalpy[docs]` resolved to `causalpy==0.5.0` while `requires-python >= 3.11` on current `main`. Diagrams stayed correct only because `pyreverse` parses the checkout, not the installed wheel.
- Token exposure: `persist-credentials: false` during pip installs; push token passed only to `create-pull-request` — same pin/pattern as `contributors.yml`.

**What it does:** Python 3.12, `pip install -e '.[docs]'`, `peter-evans/create-pull-request@v8.1.1` with scoped `add-paths`, `delete-branch: true`, and preserved `reviewers: drbenvincent`.

**Checks:** PR CI green; branch up to date with `main`. Note: `uml.yml` is not exercised by PR CI.

**Before / after merge:** Please run a `workflow_dispatch` from this branch (or accept the next Monday cron as first live test). Side effects are bounded — silent success if diagrams unchanged, or an opened/updated `update-uml-diagrams` PR. Stale remote branch `update-uml-diagrams` exists; `delete-branch: true` should handle it.

**Non-blocking notes:**
- Full `pip install -e '.[docs]'` is heavier than `make uml` strictly needs — acceptable for correctness; revisit if weekly runtime becomes an issue.
- Optional: set explicit `author`/`committer` on `create-pull-request` so manual dispatch doesn’t attribute commits to the triggering user.
- PR body mentions retargeting behind #1048 — that PR doesn’t exist on this repo (likely typo).

PR #1135, opened 2026-08-03
Migrate RegressionDiscontinuity base rendering to Plotnine

### LINKED ISSUE #988: Migrate plotting from imperative matplotlib/xarray to tidydraws + plotnine
## Goal

Replace the current imperative matplotlib/xarray plotting stack with a declarative [tidydraws](https://drbenvincent.github.io/tidydraws/) + [plotnine](https://plotnine.org/) stack. The primary wins are: eliminating repeated xarray dimension guards (`isel(treated_units=0)`, `az.extract`, `.stack(sample=("chain","draw"))`), collapsing the structurally duplicated 3-panel ITS layout into a shared function, and making plots composable and themeable.

Estimated line reduction: ~75% per `_bayesian_plot` method (~248 lines → ~60 for ITS).

This is a single PR with commits at each logical milestone so the diff tells a clear story.

---

## Dependencies to add

In `pyproject.toml`, add to the main dependencies:

```toml
"tidydraws==0.4.0",
"plotnine>=0.15.0",
```

Both are on PyPI. Do not add lets-plot — plotnine only for now.

---

## Return type decision

Where a `_bayesian_plot` / `_ols_plot` method is *fully* representable in plotnine, return the `plotnine.ggplot` object instead of `tuple[plt.Figure, plt.Axes]`. Callers who need the matplotlib figure can call `p.draw()`.

**Not every method has to end up returning `ggplot`, and that's fine.** Some plots have one or two components with no plotnine equivalent (a continuous-x violin, an arrow annotation anchored to fixed data coordinates rather than an aesthetic mapping). For those, build the components that *do* map cleanly (points, ribbons, lines, vlines) as a `ggplot`, `.draw()` it once to get a matplotlib figure, and reuse the existing matplotlib snippet for the remaining component(s) as a small overlay on the drawn axes — then return `(fig, ax)`. The win is measured by how much matplotlib/seaborn/xarray boilerplate got replaced by plotnine geoms + tidydraws for the components that do migrate, **not** by whether the method's return type became `ggplot`. See "Component-wise migration" under Guardrails below.

The public `plot()` method signature stays the same (keyword arguments unchanged); only the return type changes where applicable. Update docstrings to document the actual return type per case.

---

## What to do with `plot_utils.py`

`plot_posterior_over_x` and its private helpers (`_plot_ribbon`, `_plot_spaghetti`, `_plot_histogram`, `get_hdi_to_df`) become redundant as each experiment is migrated. Delete functions as they become unused rather than leaving dead code. Once all experiments are migrated, delete `plot_utils.py` entirely and remove its imports.

The `_render_plot` template method in `base.py` (which wraps drawing in `arviz-darkgrid` style and dispatches to `_bayesian_plot`) also simplifies significantly — theming moves into `plotnine.theme()` on the returned `ggplot` object. Simplify or remove `_render_plot` as appropriate.

---

## Core pattern (applies to every experiment)

tidydraws takes ArviZ `InferenceData` and returns Polars DataFrames:

```python
import tidydraws as td
from plotnine import *

# Extract posterior predictive draws as a tidy frame
pred_df = td.prediction_draws(idata, newdata=df, var_name="mu",
                               idata_group="posterior_predictive")

# Compute point + interval summary
summary = td.point_interval(pred_df, "mu", group_by="time", probs=(ci_prob,))

# Declare the plot
p = (ggplot(summary.to_pandas(), aes("time", "mu"))
     + geom_ribbon(aes(ymin="mu_lower", ymax="mu_upper"), alpha=0.3)
     + geom_line()
     + theme_bw())
```

The `isel(treated_units=0)` dimension guards disappear — tidydraws flattens chain/draw and preserves named xarray coords as DataFrame columns automatically.

For the matplotlib escape hatch (adding `axvline`, annotations on top of a plotnine plot): call `fig = p.draw()` to get the matplotlib `Figure`, then use `fig.axes` to add overlays. Use `ax.get_subplotspec() is not None` to filter data axes from plotnine's internal strip-label axes in faceted plots.

---

## Guardrails (added after Commit 4 retro — see PR #990 discussion)

**0. Assess before coding.** For every experiment not yet migrated, before writing any plotnine code: read the current `_bayesian_plot` (and `_ols_plot` if relevant) top to bottom and list every matplotlib/arviz/seaborn construct it uses (scatter, ribbon/`fill_between`, line, `axvline`, errorbar, `az.plot_forest`, `az.plot_posterior`, violin, mosaic/`subplot_mosaic`, per-column-normalised heatmap, fixed-position `annotate`). Classify each construct as "has a native plotnine geom" (ribbon→`geom_ribbon`, vline→`geom_vline`, errorbar→`geom_pointrange`/`geom_errorbar`, forest→`geom_pointrange`+`coord_flip`, scatter→`geom_point`) or "no plotnine equivalent" (continuous-x violin, mosaic layout, per-column heatmap, an annotation that needs a real `Axes` to draw an arrow/text at a computed pixel-ish offset). Report this list before writing the implementation.

**1. Component-wise migration is the default expectation, not an exception.** Migrate whichever constructs map cleanly to plotnine geoms, even if one or two components in the same plot don't (see the DiD guidance below for a worked example). It is completely normal and desired for a method to end up building most of its output as a `ggplot`, calling `.draw()` once, and then reusing the *existing, unmodified* matplotlib snippet to add the one or two components with no plotnine equivalent, returning `(fig, ax)`. This is not a failure to migrate — it's the correct amount of migration. Do not throw away or rewrite the matplotlib snippet for a component that has no plotnine equivalent; reuse it as-is.

**2. Minimize escape-hatch surface; never duplicate a full second implementation.** The mistake in the first Commit 4 attempt wasn't building a `.draw()`-then-overlay hybrid — it was rebuilding *almost the entire plot* in plotnine and *also* keeping a full byte-for-byte copy of the original method as a fallback for other `kind=` values, so the file only grew. When you build the plotnine base, build it once and share it across every case that can use it (e.g. control/treatment ribbons are identical whether the counterfactual needs a violin or not); only the genuinely-incompatible component should differ between cases. If you find yourself maintaining two largely-parallel full implementations, stop — that's the signal to consolidate into one shared plotnine base with a small overlay branch, not two methods.

**3. Line-count check before declaring a commit done.** After implementing, run `git diff --stat` on the touched experiment file. Insertions should mostly come from genuinely new plotnine/tidydraws code for components that are migrating; deletions should mostly come from removed matplotlib/seaborn/`plot_posterior_over_x` calls for those same components. If insertions run far ahead of deletions with no corresponding component actually removed, stop and report the diff stat plus a one-paragraph explanation to the human before proceeding — don't let it surface only after tests/images are ready.

**4. Blast-radius rule for shared infrastructure.** A migration commit should touch only the one experiment file being migrated (plus its own dedicated tests). If implementing a commit seems to require editing `base.py`, `plot_utils.py`, or a test file that also covers *other*, unrelated experiments (e.g. `test_plot_show_parameter.py`, which covers ITS too), stop and get explicit approval before making that wider change, and propose it as a small, separately-reviewable piece — don't bundle a shared-infrastructure fix into the same commit as the experiment migration.

**5. Propose-before-code for non-trivial escape hatches.** If an escape hatch is more than a couple of `axvline`/`annotate`/`violinplot` calls bolted onto an otherwise-pure `ggplot`, describe the intended approach and expected line-count impact to the human *before* writing the implementation, not after.

### Worked example: `DifferenceInDifferences` (component-wise, not "kept on matplotlib")

DiD's plot has four components: (a) scatter of raw data, (b) control-group ribbon+line, (c) treatment-group ribbon+line, (d) counterfactual (a violin when there's a single post-treatment period, otherwise a normal ribbon+line), (e) a causal-impact arrow+text annotation. (a)–(c) and the multi-period case of (d) are plain `geom_point`/`geom_ribbon`/`geom_line` — clean wins, no reason to keep them on matplotlib/seaborn. Only the single-period violin has no plotnine geom. The recommended shape: build (a)–(c) as a `ggplot` (shared across both counterfactual cases), `.draw()` it, then reuse the *existing* `_plot_causal_impact_arrow` matplotlib helper unchanged for (e), and reuse the *existing* violin-plotting snippet unchanged for (d) when there's a single post-treatment period (add the counterfactual as a plotnine ribbon+line onto the same base before `.draw()` when there are multiple periods). Return `(fig, ax)`. This still deletes the `sns.scatterplot` and both `plot_posterior_over_x` calls for control/treatment, which is a real, if modest, win — without inventing a parallel plotnine-only code path for the rare multi-period case.

---

## Per-experiment status and guidance

Commits are no longer pre-assigned to a fixed order or bundled ("do these four together"). Each experiment gets its own commit, sized and scoped only after the Guardrail #0 assessment has actually been done for *that* experiment — don't assume a bundle of experiments shares a clean helper until each has been individually assessed. Work top to bottom; re-assess status as you go and update this table.

| Experiment | Status | Guidance |
|---|---|---|
| `RegressionDiscontinuity` | ✅ Done | Delivered — clean win, fully `ggplot`. |
| `PrePostNEGD` | ✅ Done | Delivered — clean win, two-facet `ggplot`. |
| `DifferenceInDifferences` | 🟡 To do — component-wise | See the worked example above under Guardrails. Migrate scatter + control/treatment ribbons/lines (and the counterfactual ribbon/line for the multi-period case) to plotnine; keep the single-period violin and the arrow+text annotation on matplotlib by reusing the existing snippets unchanged after `.draw()`. Expect a modest reduction, not a full `ggplot` return type. |
| `RegressionKink` | 🟡 To do | Same shape as RD (scatter + one ribbon/line + one `axvline` for the kink point), ~83-line method — do the Guardrail #0 assessment first, but this looks like a clean win similar to RD. |
| `InterruptedTimeSeries` | 🟡 To do — assess first | ~250-line, 3-panel method. Already has a `_draw_singleton_hdi_marker` helper that exists specifically to patch a single-post-period edge case (arviz's `plot_hdi` degenerates to a zero-area polygon there) — this is the same shape of problem that made DiD's counterfactual non-trivial. Do the construct inventory before assuming the "shared 3-panel helper" framing applies cleanly; the singleton edge case may need the same component-wise treatment as DiD's violin. |
| `SyntheticControl` | 🟡 To do — assess first | ~194-line method, structurally similar to ITS. Assess independently; don't assume it shares a helper with ITS until both have been individually inventoried. |
| `SyntheticDifferenceInDifferences` | 🟡 To do — assess first | ~169-line method. Assess independently. |
| `PiecewiseITS` | 🟡 To do — assess first | ~141-line method. Assess independently. |
| `StaggeredDifferenceInDifferences` | 🟡 To do — assess first | Two plot entry points (`plot`, `plot_group_time`), errorbar-heavy — check `geom_pointrange`/`geom_errorbar` fit for the event-study markers. |
| `PanelRegression` | 🔴 Low priority, assess last | Forest plot (`_bayesian_plot`) plus four separate large plotting methods (`plot_coefficients`, `plot_unit_effects`, `plot_trajectories`, `plot_residuals`) — the largest and most heterogeneous plotting surface in the codebase. Assess each of the five methods independently; do not treat this as one commit. |
| `InversePropensityWeighting` (`plot_ate`, `plot_balance_ecdf`) | ⛔ Out of scope (pre-existing decision) | Mosaic layout + ECDF, no plotnine equivalent. |
| `InstrumentalVariable` | n/a | `plot()` raises `NotImplementedError`, unchanged. |

**Cleanup (only once every migration above has landed or been explicitly marked out of scope):**
- Delete unused functions from `plot_utils.py` as they stop being called; don't force full deletion before every consumer is resolved.
- Simplify `_render_plot` in `base.py` only as a dedicated, separately-reviewed step (see Guardrail #4) — not bundled into an experiment commit.
- Update type annotations and docstrings per method as each one's actual final return type is settled.

---

## Validation workflow

**Commit 0 — baseline harness (before any migration code)**

Create `.scratch/validate_plots.py`. This script fits each experiment with a small synthetic dataset (`sample_kwargs={"draws": 100, "tune": 100, "chains": 2}`), calls `experiment.plot()`, and saves the output to `.scratch/plot_validation/{experiment_name}.png`. At this point it captures the existing matplotlib output as the visual baseline.

Run it once to generate the baseline images, then commit them to `.scratch/plot_validation/` (untracked, local only).

**After each subsequent commit**

Re-run `.scratch/validate_plots.py`. The script saves new images alongside the old ones (e.g. `rd_before.png` / `rd_after.png`). **Stop and wait for human review before proceeding to the next commit.** Use the image-reading tool to actually display each before/after pair inline in the response (not just mention that images exist), and also state the absolute file path for each one so the reviewer can open it directly regardless of whether the inline render comes through.

The gate is: tests pass AND human has visually approved the before/after images for that experiment. Do not chain commits without this approval, and do not commit or post PR comments before that approval is given.

**What to check in each before/after pair**
- HDI band is centred on the correct data (not shifted or on the wrong series)
- Pre/post split is correct for ITS-family plots
- Colour assignments match between groups and the legend
- Panel order is correct for multi-panel plots
- Treatment cutoff lines appear in the right position

---

## Tests

Run the existing plot test suite after each commit. Only update assertions for the return type that a method *actually* ends up with per its Guardrail #0 assessment — `ggplot` for fully-migrated methods, `(fig, ax)` unchanged for component-wise or not-yet-migrated ones. Do not delete tests — update them to match. If a test file covers multiple unrelated experiments (e.g. `test_plot_show_parameter.py`), see Guardrail #4 before editing it.

Use `required_permissions: ["all"]` when running pytest (PyMC requires filesystem access outside the sandbox).

---

## Out of scope for this PR

- lets-plot as a backend (future work)
- `stat_ecdf` implementation in plotnine (tracked in plotnine#1099)
- IV `plot()` — currently raises `NotImplementedError`, leave as-is

cetagostini:
## Problem
Issue #988 requires replacing imperative plotting where Plotnine has a faithful equivalent without weakening CausalPy's existing public rendering contracts.

## Change
This component-wise transition migration adds `plotnine`, `polars`, and `tidydraws` to main dependencies; `environment.yml` was regenerated by the `pyproject2conda-yaml` prek hook. `RegressionDiscontinuity` now materializes points, the OLS fit, and labels with a single Plotnine draw. Renderer-only frames use internal columns, so valid user columns named `series` or `prediction` cannot collide with the new layers. The code deliberately retains the existing Matplotlib posterior HDI/ETI/histogram/spaghetti snippet, Axes-resident threshold/donut rules, and final Axes legend path, so public `plot()` keywords and its `(Figure, Axes)` result remain unchanged. The fit-selection mask is retained positionally so duplicate data indices cannot misclassify excluded rows. This adapts the component-wise design from @drbenvincent's draft #990 commit `aaaa852e` to the transition branch's current local PyMC 6/ArviZ 1.x plotting helpers; the initial migration commit includes a co-author trailer.

## Guardrail 0 construct inventory and status
| Experiment | Current construct inventory | Status |
| --- | --- | --- |
| RegressionDiscontinuity | observed/excluded points; OLS fit; Bayesian posterior mean, HDI/ETI ribbon, histogram, spaghetti; threshold/donut rules; score title; legend | Migrated only points, OLS fit, and labels to Plotnine. Posterior uncertainty, rule artists, and legend remain the existing Matplotlib components because their 0.94 HDI/ETI/histogram/spaghetti, Axes-artist, and legend-kwargs behavior are contractual. |
| DifferenceInDifferences | group trajectories, posterior bands, continuous-x violin, causal-impact arrow | Deliberately unchanged: violin and arrow require Matplotlib components; follow-up needed for compatible base geometry. |
| InterruptedTimeSeries, PiecewiseITS, SyntheticControl, SyntheticDifferenceInDifferences | three-panel causal layouts, uncertainty bands, shading, datetime rules, legends | Deliberately unchanged: mosaic/facet-equivalent panel assembly and posterior/legend components require a tested hybrid design. |
| PrePostNEGD, RegressionKink, StaggeredDifferenceInDifferences, PanelRegression | posterior/OLS overlays, scalar/posterior components, event/cohort or coefficient layouts | Deliberately unchanged pending a separately validated component migration. |
| InversePropensityWeighting, InstrumentalVariable | excluded by issue scope | Unchanged. |

## Per-file diff stat
| File | + | - |
| --- | ---: | ---: |
| `causalpy/experiments/regression_discontinuity.py` | 111 | 62 |
| `causalpy/tests/test_input_validation.py` | 114 | 24 |
| `causalpy/tests/test_posterior_plot_contracts.py` | 66 | 0 |
| `pyproject.toml` | 4 | 0 |
| generated `environment.yml` | 2 | 0 |

## Visual review
Before: `/Users/carlostrujillo/Documents/GitHub/_worktrees/CausalPy-988-plotnine/.scratch/plot_validation/rd_pymc_before.png` and `/Users/carlostrujillo/Documents/GitHub/_worktrees/CausalPy-988-plotnine/.scratch/plot_validation/rd_ols_before.png`. After: `/Users/carlostrujillo/Documents/GitHub/_worktrees/CausalPy-988-plotnine/.scratch/plot_validation/rd_pymc_after.png` and `/Users/carlostrujillo/Documents/GitHub/_worktrees/CausalPy-988-plotnine/.scratch/plot_validation/rd_ols_after.png`. These images are intentionally untracked; both regenerated final OLS and PyMC images were inspected.

## Evidence
`/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-988-plotnine python -m pytest causalpy/tests/test_input_validation.py -k 'donut or isolated' --no-cov` — 9 passed. `/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-988-plotnine python -m pytest causalpy/tests/test_plot_utils.py causalpy/tests/test_plot_show_parameter.py causalpy/tests/test_posterior_plot_contracts.py causalpy/tests/test_hdi_prob_wiring.py --no-cov` — 113 passed. `/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-988-plotnine prek run --files causalpy/experiments/regression_discontinuity.py causalpy/tests/test_input_validation.py` — passed. `/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-988-plotnine python scripts/generate_plots.py --tag after --only rd --backend pymc ols` — regenerated both visual-review images.

## Review
The first implementation review found two high renderer-column collisions; both were fixed and are covered by `test_rd_plot_isolated_from_user_column_names`. The final adversarial, correctness, edge-case, and test reviews all approved. A source fact-check and workflow supervisor both passed with no high/critical blocker.

## Maintainer decisions not taken unilaterally
This PR intentionally does not close #988 because the remaining experiments need follow-up component migrations. It does not replace the established Matplotlib posterior helper or change any public plot signature/return type.

## Release notes needed
RegressionDiscontinuity now uses Plotnine for its base points, OLS fit, and labels while retaining the existing Matplotlib posterior uncertainty, threshold/donut rules, and legend components; no public plotting API changed.

Refs #988


drbenvincent:
FYI. Can't quite remember but I think I have a bug in progress or to do the plotnine conversion. Can't remember what state it's in, but finishing off the big reactor pr's was on my list for August

cetagostini:
> FYI. Can't quite remember but I think I have a bug in progress or to do the plotnine conversion. Can't remember what state it's in, but finishing off the big reactor pr's was on my list for August
> 
> 

I thing, maybe move that in the migration branch, and let agents fully handle

Issue #874, opened 2026-04-30
Promote plot_placebo_calibration into the library and auto-populate CheckResult.figures


drbenvincent:
## Summary

Promote the `plot_placebo_calibration` helper currently defined inline in `docs/source/notebooks/its_place_in_time_analysis.ipynb` (introduced in #826) into the library, and have `PlaceboInTime.run()` populate `CheckResult.figures` with it by default.

> [!NOTE]
> Depends on #826 landing first — that PR introduces the notebook and the helper. This issue tracks the follow-up cleanup discussed in [the inline review thread](https://github.com/pymc-labs/CausalPy/pull/826#discussion_r3133448350).

cc @cetagostini

## Why

The notebook defines a ~123-line helper (`plot_placebo_calibration(pit_check, original_result, title=...)`) and calls it three times across different datasets. The reuse pattern is a strong signal that the helper is general, not example-specific.

The helper is keyed entirely on the `PlaceboInTime` `CheckResult` contract:

- `pit_check.metadata["fold_results"]`
- `pit_check.metadata["null_samples"]`
- `pit_check.metadata["actual_cumulative_mean"]`
- `pit_check.metadata["p_effect_outside_null"]`
- `original_result.post_impact` (present on every `applicable_methods` member of `PlaceboInTime`: ITS and SyntheticControl)

It already handles both datetime and integer `pseudo_treatment_time` formatting, so it is general for every experiment type `PlaceboInTime` supports — not ITS-specific.

There is also a documented home for it in `CheckResult` in `causalpy/checks/base.py` that no check currently uses:

```python
figures : list
    Optional matplotlib figures produced by the check.
...
figures: list[Any] = field(default_factory=list)
```

Promoting this helper would make `PlaceboInTime` the first concrete user of that field, and would make `GenerateReport` strictly more useful for free.

## Proposed shape

Two complementary pieces:

1. **Method on the check class** — `PlaceboInTime.plot_calibration(check_result) -> Figure`, living next to the metadata schema it depends on in `causalpy/checks/placebo_in_time.py`. Mirrors how `BaseExperiment` exposes its plotting methods.
2. **Auto-populate `CheckResult.figures`** — call the plotter inside `PlaceboInTime.run()` (gated behind `make_figures: bool = True`) and append the `Figure` to the result.

## Cleanups while moving

[...]

PR #1140, opened 2026-08-04
Promote the placebo-in-time calibration plot into the library

anevolbap:
## Summary

Completes the placebo-in-time calibration plot from #874 and ports the maintainer-authorized exact operating-characteristics implementation into the transition branch.

- `PlaceboInTime` still attaches its calibration figure to every result, including an annotated placeholder for INCONCLUSIVE runs that deliberately omit learned-null metadata.
- Exact ROPE operating curves, MDE, false-positive rate, and assurance now consume the identified placebo null without posterior-decision Monte Carlo. They preserve the existing positive-first, inclusive decision rule for every threshold in `(0, 1)`.
- Arrays and frozen CDF/SF priors are evaluated deterministically; RVS-only prior draws retain seed-aware and legacy-diagnostic handling. RVS-only priors omit the optional plot strip rather than making a changing figure.
- Invalid threshold, ROPE, and random-seed configuration now fails before fitting; the default operating grid remains meaningful for a degenerate learned null.
- The reporting and sensitivity guidance, including the placebo notebook, now present ROPE/HDI, placebo calibration, exact operating characteristics, and prior-integrated assurance as one workflow.

## Validation

- `/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-765-rope python -m pytest causalpy/tests/test_placebo_in_time.py causalpy/tests/test_generate_report.py causalpy/tests/test_operating_characteristics.py --no-cov` — 188 passed, 1 deselected.
- `/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-765-rope make check-exports` — passed.
- `/opt/anaconda3/condabin/mamba run -p /Users/carlostrujillo/Documents/GitHub/_worktrees/.mamba/CausalPy-765-rope python scripts/validate_notebooks.py docs/source/notebooks/interrupted-time-series-placebo-in-time-analysis.ipynb` — passed.
- Changed-file `prek -c .pre-commit-config.yaml run --files ...` — passed.

## Maintainer decision

The historical `feature/power-analysis-sigmoid` `power_analysis` API is intentionally not restored. This PR instead includes the maintainer-authorized in-progress `operating_characteristics` rewrite as the clean transition target; it supersedes the draft’s old local power-analysis direction rather than adding a compatibility alias.

## Follow-up coverage

The review found worthwhile but non-blocking coverage improvements to add separately: independently derived numerical frozen-prior and heterogeneous-array-prior assurance references; package-level re-export identity assertions; and an HTML-report assertion for the assurance prose. The optional array-prior strip could also communicate mass beyond the plotted effect range more explicitly.

## Release notes needed

- Add exact placebo-calibrated operating characteristics and deterministic assurance for `PlaceboInTime`, alongside its calibration plot and ROPE guidance.
- `PlaceboInTime` now reports INCONCLUSIVE rather than fabricating a learned null when too few usable folds or unidentified between-fold spread prevent calibration.


causalpy/checks/placebo_in_time.py
```diff
@@ -340,6 +362,7 @@ def __init__(
         n_design_replications: int | None = None,
         random_seed: int | None = None,
         intervention_length: Any | None = None,
+        make_figures: bool = True,
```
anevolbap, line 339: The issue proposes gating this on `run()`. That flag would be unreachable from a pipeline: `SensitivityAnalysis.run` calls `check.run(experiment, context)` positionally, so nothing could ever pass it. Constructor it is.

Default is `True` per the acceptance criteria. One figure costs nothing next to `n_folds` experiment refits plus the hierarchical fit, and defaulting it off would leave `CheckResult.figures` as unused as it has been since it was documented.

causalpy/checks/placebo_in_time.py
```diff
@@ -1478,6 +1502,7 @@ def run(
             "fold_sds": fold_sds,
             "status_quo_idata": idata,
             "null_samples": theta_new_samples,
+            "actual_cumulative_samples": np.asarray(actual_cumulative.values).ravel(),
```
anevolbap, line 1457: This is the `original_result` argument the issue asks to drop. `run()` already computed these samples, so recording them here means the plotter reads one object and cannot be handed a mismatched pair.

It also fixes a latent bug: the notebook helper recomputed panel C as `original_result.post_impact.sum("obs_ind")`, without the `isel(treated_units=0)` that `_extract_cumulative_impact` applies, so it was wrong-shaped for `SyntheticControl`, which is in `applicable_methods`.

causalpy/checks/placebo_in_time.py
```diff
@@ -1516,12 +1541,208 @@ def run(
...
+
+    @staticmethod
+    def _format_fold_time(pseudo_treatment_time: Any) -> str:
+        """Format a pseudo treatment time for a fold label."""
+        if hasattr(pseudo_treatment_time, "strftime"):
+            return f"{pseudo_treatment_time:%Y}"
+        return f"{pseudo_treatment_time}"
+
+    @staticmethod
+    def _draw(plot: Any, title: str, figsize: tuple[float, float]) -> Figure:
```
anevolbap, line 1689: `ggplot.draw()` returns a plain matplotlib `Figure`, which is what makes the plotnine move invisible downstream: `CheckResult.figures` keeps its documented type and `GenerateReport` keeps using `savefig`.

It also matters for the `make_figures=True` default. A drawn plotnine figure is not registered with pyplot (`plt.get_fignums()` stays empty), so building one inside `run()` neither accumulates managed figures nor makes the inline backend display it at the end of whichever notebook cell called `run()`. That would have happened with `plt.subplots`.

causalpy/steps/report.py
```diff
@@ -79,22 +79,45 @@ def validate(self, context: PipelineContext) -> None:
...
-            fig.savefig(buf, format="png", dpi=100, bbox_inches="tight")
+            plots.append(self._encode_figure(fig))
             plt.close(fig)
-            buf.seek(0)
-            plots.append(base64.b64encode(buf.read()).decode("utf-8"))
         except Exception as exc:
             logger.debug("Could not render plot: %s", exc)
         return plots
 
+    def _render_check_figures(self, check_result: Any) -> list[str]:
```
anevolbap, line 105: Deliberately no `plt.close` here, unlike `_render_plot` just above. That figure is created and owned by the report; these belong to the `CheckResult` the caller still holds, so closing them would blank an object the user can still read.

derwells:
Linking this for the open-PR sweep (#1048). The operating-characteristics module added here in covers what #989 does.

Tagged #989 `migration:superseded` and leaving it open. Coordinating wrap-up on #1048.

PR #1144, opened 2026-08-04
Report MSPE ratios in PlaceboInSpace and add plot_mspe_ratio

### LINKED ISSUE #915: Promote RMSPE-ratio view to a first-class PlaceboInSpace plot method
## Summary

PR #819 added per-donor pre/post RMSPE and the post/pre RMSPE ratio to `PlaceboInSpace.run()`, and the `sc_pymc.ipynb` notebook now hand-rolls a forest plot of the resulting ratios as the canonical Abadie inference view. The plot is small but recurring boilerplate; promote it to a first-class method on `PlaceboInSpace` so users get the inferential view "for free" instead of copy-pasting from the notebook.

## Background

`PlaceboInSpace.plot()` currently shows raw cumulative impact per donor with the treated unit highlighted. The raw view is useful but is *not* the inferential statistic recommended by `{cite:t}`abadie2010synthetic`` §5.2. The canonical inference statistic is the ratio of post-period to pre-period RMSPE: a unit with a large ratio fits well in the pre-period and diverges in the post-period, which is the signature of either a real effect or a structural break. Inference is the permutation rank of the treated unit's ratio in the empirical donor distribution.

After PR #819, the data needed to draw this plot is already on the `CheckResult`:

- `pis_result.table["pre_rmspe"]`, `["post_rmspe"]`, `["rmspe_ratio"]` — per-donor ratios.
- `pis_result.metadata["baseline_rmspe"]` — actual treated unit's ratio.

The notebook cell that builds the plot today is short but non-trivial (ordering, treated highlighting, infinite-ratio filtering, permutation p-value). All of it is library work that should not live in user code.

## Proposed change

Two design options; pick one in implementation.

**Option A: a new method.**

```python
@staticmethod
def plot_rmspe_ratio(
    result: CheckResult,
    *,
    treated_label: str = "treated",
    figsize: tuple[float, float] | None = None,
    show_pvalue: bool = True,
) -> tuple[plt.Figure, plt.Axes]:
    ...
```

Pros: clearly distinct from the raw-effect `plot()`; easy to discover; backwards compatible. Cons: two methods to maintain.

**Option B: a `mode=` kwarg on the existing `plot()`.**

```python
PlaceboInSpace.plot(result, mode="rmspe_ratio", ...)
```

Pros: single entry point; can extend with more views later. Cons: signature gets cluttered; backwards-incompatible if we change the default mode.

Preference is Option A.

In both cases the method should:

- Order units by ratio.
- Highlight the actual treated unit in a distinct colour.
- Draw a reference line at ratio = 1.
- Optionally compute and display the permutation p-value `n_units_with_ratio_geq_treated / total_units` (default on, off via kwarg).
- Filter infinite ratios with a `UserWarning` rather than silently dropping.
- Use `causalpy/checks/_plot_helpers.py` if a generic "ranked bar plot" helper makes sense alongside `forest_plot()`.

## Deferred work: posterior RMSPE

The current implementation uses *posterior-mean* residuals to derive a scalar RMSPE per unit. The fully Bayesian alternative is to compute the RMSPE for every posterior sample, giving a posterior distribution over RMSPE (and hence the ratio) per unit. Pros: richer uncertainty quantification; matches the spirit of the Bayesian SC framework. Cons: more compute, larger result objects, fiddlier to plot.

If we go this way, the plot would show credible intervals on ratios rather than point bars, and the permutation p-value would become a probability statement about ranking. Treat this as a follow-up after Option A lands.

## Acceptance criteria

- [ ] New `PlaceboInSpace.plot_rmspe_ratio()` method (Option A) implemented and documented.
- [ ] Method consumes the existing `pre_rmspe` / `post_rmspe` / `rmspe_ratio` columns and `metadata["baseline_rmspe"]`.
- [ ] Permutation p-value displayed by default, suppressible via kwarg.
- [ ] Unit tests covering: basic rendering, treated-unit highlighting, p-value computation, infinite-ratio filtering, missing-baseline edge case.
- [ ] `sc_pymc.ipynb` replaces its hand-rolled RMSPE-ratio cell with a single `PlaceboInSpace.plot_rmspe_ratio(pis_result, treated_label=...)` call.
- [ ] Docstring cites `{cite:t}`abadie2010synthetic`` §5.2 and explains the inferential interpretation.

## Out of scope

- Posterior-RMSPE implementation (tracked above as deferred work; can become a child issue once Option A lands).
- Changing the default behaviour of the existing `plot()`.
- Wiring the RMSPE-ratio plot into `GenerateReport` (separate small follow-up once the method exists).

anevolbap:
Refs #915. The issue should stay open for the notebook criterion, see the caveat below.

`PlaceboInSpace` compared placebo units by raw effect size, which is not the statistic Abadie, Diamond and Hainmueller (2010) recommend in section 3.4 and Figure 8. A donor the synthetic control fitted badly before the intervention can diverge a lot after it without that meaning anything. Dividing the post-period MSPE by the pre-period MSPE removes that, and the treated unit's rank in the donor distribution is the inference.

## What changed

`run()` records `pre_mspe`, `post_mspe` and `mspe_ratio` per placebo unit in `table`, and the treated unit's own figures in `metadata["baseline_mspe"]`. Both are additive, the existing effect-summary columns are untouched. The ratio is the ADH quantity unchanged, not its square root, so values compare directly to the ones published there. `CheckResult.text` points to the `mspe_ratio` column when the table has one. Building a placebo's row and computing its MSPE now happen outside the fit-failure handler, so an error there raises instead of turning into an `error` row; a failed fit still becomes an `error` row. A control that is also listed as a treated unit is skipped as a placebo, since a unit cannot be its own placebo and the plot would otherwise rank it twice.

`PlaceboInSpace.plot_mspe_ratio(result)` draws every unit's ratio in rank order, highlights the treated unit, marks a reference line at 1, and reports the permutation p-value. Every argument after the result is keyword-only.

Each treated unit is ranked against the placebo pool plus itself, never against the other treated units, so one treated unit's p-value does not depend on how many others were fitted.

Degenerate ratios are separated rather than collapsed. A positive post-period error over a zero pre-period error is `inf`: defined and ranked, but undrawable on a finite axis, so it leaves the bars and stays in the p-value denominator. Zero over zero, a non-finite error, and a failed placebo fit are `nan`: outside both the figure and the reference set. Each case warns separately, and so does asking for a p-value when the result has no treated baseline. The reported p-value is conditional on the units that fitted successfully.

`PlaceboInTime._draw` moved to `causalpy/checks/_plot_helpers.py` as `draw_figure`, since it now has a second caller. `_draw` stays as a delegate, so no existing call site changed. This is the one part of the PR that is not purely additive: reviewing the `PlaceboInSpace` feature means also accepting a change to `PlaceboInTime`. It is a separate first commit if you prefer to take it on its own.

`docs/source/release_notes.md` gets a "New features" entry under 1.0.0.

## Deliberate omissions and deviations from the issue

`run()` does not populate `CheckResult.figures`. `PlaceboInTime` auto-populates it behind a `make_figures` flag since #1140, so the two checks now differ. The issue puts `GenerateReport` wiring out of scope, and adding the flag means giving `PlaceboInSpace` an `__init__` it currently does not have, so I left it. Happy to add it if you would rather the two checks match.

The full-posterior MSPE is not implemented. Residuals are averaged over the posterior before squaring, giving a scalar per unit, which is what the issue asked for. A distribution over the ratio is reachable from the same `pre_impact`/`post_impact` arrays and is worth a child issue.

The issue's proposed signature differs in three places. There is no `treated_label` argument: the treated bars are labelled with the unit names from `metadata["baseline_mspe"]`, so a separate label would repeat them. The method returns a `Figure` rather than `(Figure, Axes)`, matching `PlaceboInTime.plot_calibration`. The docstring names the paper in prose instead of a `{cite:t}` role, because no Python docstring in the package uses sphinxcontrib-bibtex roles today; the cited version is in `sensitivity_checks.md`.

## Caveat on the issue

The issue assumed the RMSPE columns already existed. They do not exist on any branch, and the `sc_pymc.ipynb` cell it refers to plots mean effect with HDI intervals rather than ratios. So this PR computes the errors as well as plotting them. The notebook acceptance criterion is not met: `sc_pymc.ipynb` only exists on the #819 branch. The prose documentation went into `docs/source/notebooks/sensitivity_checks.md` instead.

#819 also adds `causalpy/checks/_plot_helpers.py` (with `forest_plot` and `null_distribution_plot`) and edits `placebo_in_space.py`, so whichever of the two lands second will need a small merge.

## Tests

29 tests for this feature in `causalpy/tests/test_method_specific_checks.py`: the statistic's definition against hand-computed residuals, the posterior mean being taken before squaring, the three degenerate cases, the new columns and metadata, the `run()` text, rendering, ordering, treated highlighting, the per-unit p-value, its suppression and the warning when there is no baseline to rank, the keyword-only signature, the multi-treated reference set, the missing-baseline and unfitted-experiment paths, an error in the statistic that has to raise rather than count as a failed fit, and both `ValueError` paths.

The definition test is the one that would catch the statistic drifting: its fixture gives 4 as an MSPE ratio and 2 as an RMSPE ratio, so an implementation that reports the square root fails it.

Two tests cover the seam between `run()` and the plot, since every other plot test builds its `CheckResult` by hand and would not catch the two drifting apart. One plots a real `run()` result. The other fails a single placebo fit, so a `NaN` ratio reaches the plot, which is a different route into the undefined branch than a zero pre-period error produces.

`pytest -k PlaceboInSpace` is 37 passed. Full suite is 2524 passed, 20 skipped. `diff-cover` against `pymc6_and_pymcmarketing1_migration` reports 100% over 93 changed production lines; it counts lines, so Codecov, which also counts branches, can report partials that `diff-cover` does not. The gate excludes `causalpy/tests/*`.


causalpy/checks/placebo_in_space.py
```diff
@@ -22,17 +22,79 @@
...
 
+_DEFAULT_PLOT_TITLE = "Placebo-in-space: post/pre RMSPE ratio"
+_DEFAULT_FIGSIZE = (7.0, 8.0)
+# Grey for the donor pool, red for the actual treated unit, matching the
+# palette the other check figures use.
+_PLACEBO_COLOUR = "#94a3b8"
+_TREATED_COLOUR = "#E24A33"
+
+
+def _rmspe(impact: xr.DataArray) -> float:
```
anevolbap, line 58: Residuals are averaged over the posterior before squaring, so each unit gets a scalar RMSPE. The full-posterior version (an RMSPE distribution per unit, credible intervals on the ratio) is reachable from the same `pre_impact`/`post_impact` arrays. Follow-up issue, or do you want it here?

causalpy/checks/placebo_in_space.py
```diff
@@ -137,16 +200,151 @@ def run(
 
         table = pd.DataFrame(rows) if rows else None
 
+        # The config can name treated units the fitted experiment does not
+        # carry, so only units present in its impact coordinates get a
+        # baseline; the rest are simply absent from the metadata.
+        baseline_rmspe: dict[str, dict[str, float]] = {}
+        if isinstance(experiment, SyntheticControl):
```
anevolbap, line 207: `run()` is annotated `BaseExperiment` but needs `pre_impact`, so mypy needs narrowing here. I used `isinstance` over `cast` so a non-SC experiment yields no baseline rather than an `AttributeError`. `validate()` already rejects non-SC, so this branch should be unreachable in the pipeline. Prefer `cast`?

causalpy/checks/placebo_in_space.py
```diff
@@ -137,16 +200,151 @@ def run(
...
+            f"is the statistic to rank units by; see "
+            f"`PlaceboInSpace.plot_rmspe_ratio`."
         )
 
         return CheckResult(
             check_name="PlaceboInSpace",
             passed=None,
             table=table,
             text=text,
+            metadata={"baseline_rmspe": baseline_rmspe},
```
anevolbap, line 233: `text` gained a sentence and `metadata` went from empty to populated. Additive, but user-visible. Does the 1.0.0 release note want a line, or is that section only for breaking changes? #1140 added `plot_calibration` without one.

causalpy/checks/placebo_in_space.py
```diff
@@ -137,16 +200,151 @@ def run(
...
+    def plot_rmspe_ratio(
+        check_result: CheckResult,
+        title: str = _DEFAULT_PLOT_TITLE,
+        figsize: tuple[float, float] = _DEFAULT_FIGSIZE,
+        show_pvalue: bool = True,
+    ) -> Figure:
+        """Plot the post/pre RMSPE ratio of every unit, treated unit highlighted.
+
+        This is the inferential view recommended in Abadie, Diamond and
+        Hainmueller (2010), section 5.2.  A unit with a large ratio tracks its
```
anevolbap, line 246: The issue asked for a `{cite:t}` role here, but no Python docstring in the package uses sphinxcontrib-bibtex roles, so this is plain prose and the cited version went into `sensitivity_checks.md`. Should docstrings start carrying citations?

causalpy/checks/_plot_helpers.py
```diff
@@ -0,0 +1,54 @@
...
+
+from typing import Any
+
+from matplotlib.figure import Figure
+
+# Suptitle position, just above the figure box the panels fill.
+_SUPTITLE_Y = 1.02
+
+
+def draw_figure(plot: Any, title: str, figsize: tuple[float, float]) -> Figure:
```
anevolbap, line 26: A new module for one function. The alternative was importing `PlaceboInTime._draw` across check modules. Is a shared `_plot_helpers` the direction you want as more checks get plots?

drbenvincent:
## Automated triage

**Why:**
- Adds a new public method (`PlaceboInSpace.plot_rmspe_ratio`) and modifies `PlaceboInTime._draw` — both are user-facing API in the checks module.
- Implements a new statistical ratio (pre/post RMSPE) from Abadie et al. (2010), which is a substantive numerical behaviour change for an inference method.
- 202 new lines in the main check code, 306 new lines of tests, and a refactored shared helper. Large diff across 5 files.
- All CI checks pass across 3 Python/pandas variants; codecov/patch reports 100%.

**Review focus:**
1. Verify the RMSPE ratio formula matches the ADH (2010) reference and handles edge cases (zero pre-period RMSPE, failed placebos).
2. Confirm the `PlaceboInTime._draw` refactor to `_plot_helpers.py` preserves existing rendering.
3. Review the deliberate omissions (no `make_figures` flag, no full-posterior RMSPE) and decide whether to block or file follow-up issues.

**Confidence:** high

drbenvincent (changes requested):
Agent-drafted review.

## Executive Summary

Recommendation: request changes.

Value delivered:
- `PlaceboInSpace.run()` now carries pre-period, post-period, and post/pre fit-error diagnostics from each successful placebo fit into the existing `CheckResult`, while preserving the prior effect-summary columns and recording the actual treated baseline separately.
- `plot_rmspe_ratio()` gives users a ranked, treated-highlighted inferential view with a run-to-plot integration test, explicit warnings for unusable rows, and no regression in the extracted `PlaceboInTime` drawing path.
- The pre/post boundary is consistent with `SyntheticControl`: pre is `index < treatment_time` and post is `index >= treatment_time`; the plug-in Bayesian calculation intentionally takes the posterior-mean residual trajectory before reducing over time.

Review focus:
- The single-treated happy path works and the rank p-value includes ties and the treated unit, but the public statistical contract does not match the cited ADH quantity, and the p-value reference set is wrong for CausalPy's supported multi-treated Bayesian synthetic-control case.

## Findings

- [must-fix] `causalpy/checks/placebo_in_space.py:58-95`, `causalpy/checks/placebo_in_space.py:243-256`, and `docs/source/notebooks/sensitivity_checks.md:133`: the implementation computes `post_RMSPE / pre_RMSPE`, but Abadie, Diamond, and Hainmueller (2010) section 5.2 and Figure 8 use `post_MSPE / pre_MSPE`. These are not numerically the same: this PR reports the square root of the cited ratio (for example, the paper's ratio of about 130 would be about 11.4 here). The square root is monotone, so ordering and the finite-case permutation p-value are unchanged, but the published columns, axis values, and claim that this is “the statistic recommended” are inaccurate. Before exposing this API, either compute and name the ADH MSPE ratio or explicitly specify that CausalPy reports the square-root transform and explain why; add a known-residual test that pins the selected definition, because the current tests only assert `ratio == post_rmspe / pre_rmspe` and cannot detect an incorrect RMSPE implementation.
- [must-fix] `causalpy/checks/placebo_in_space.py:328-335`: with multiple actual treated units, every annotation uses the full plotted frame, which includes all treated baselines. Each treated unit's randomization reference set should be the placebo units plus that treated unit, not the other non-placebo treated units. A probe with placebo ratios `[1, 2, 3]` and treated ratios `5` and `10` produced current p-values `0.4` and `0.2`; each unit's correct placebo-plus-self p-value is `0.25`. `SyntheticControl` supports multiple treated units for Bayesian models, `run()` deliberately records multiple baselines, and the plot advertises an annotation for each, so this is a supported-path inference error. Build each annotation from the finite placebo ratios plus only the baseline being ranked, and add a multi-treated regression test.
- [must-fix] `causalpy/checks/placebo_in_space.py:237-242`: the new documented plotting method leaves `title`, `figsize`, and `show_pvalue` positional-capable. CausalPy's documented public-callable policy requires keyword-only optional controls, and issue #915's proposed signature also placed these behind `*`. This API is unshipped, so add `*` after `check_result` now and pin the signature; matching the older `PlaceboInTime.plot_calibration()` shape is not a reason to create another compatibility obligation.
- [should-fix] `causalpy/checks/placebo_in_space.py:70-86` and `causalpy/checks/placebo_in_space.py:307-335`: all zero or non-finite pre-period errors are collapsed to `inf`, then removed from both the plot and permutation denominator. This conflates at least three cases: positive-post/zero-pre (an extended-real infinite ratio), zero-post/zero-pre (`0/0`), and non-finite pre-error. It can make the result look more distinctive: with donor ratios `[inf, 1, 2]` and treated ratio `5`, the displayed p-value becomes `1/3`; retaining the ordered infinite donor gives `2/4`. Define these cases separately and document whether the reported p-value is conditional on finite, successfully fitted units. At minimum, compute inference from a clearly defined reference set independently of what a finite-axis bar chart can render.

## Merge Readiness

Verdict: request changes.

Branch status: 20 commits behind the current `pymc6_and_pymcmarketing1_migration` tip. GitHub reports the PR mergeable and clean, and a local `git merge-tree` against current base `a2f13029` found no conflict. The reviewed diff's merge base is `194343ca`.

CI status: green for the observed head `53bc222d` against the older base state, including three Python/pandas test jobs, prek, notebook jobs, Read the Docs, Codecov patch, environment sync, and zizmor. Syncing the 20 base commits should rerun CI before merge.

## PR Summary

This PR extends `PlaceboInSpace` with per-placebo and treated-baseline fit-error ratios plus a ranked plot and permutation annotation, extracts the plotnine-to-matplotlib handoff for reuse by `PlaceboInTime`, adds 17 focused tests, and documents interpretation in the sensitivity-check guide. The posterior-mean residual plug-in and omission of automatic `CheckResult.figures` population are explicit scope choices and are not blockers; the hard-to-change concerns are the quantity's public name/definition and the inference reference-set contract.

## Test Evidence

- Remote: all checks listed by `gh pr checks 1144 --repo pymc-labs/CausalPy` passed at head `53bc222d`.
- Local: `PYTHONPATH="$PWD" mamba run -n CausalPy python -m pytest causalpy/tests/test_method_specific_checks.py -k 'PlaceboInSpace' -q --no-cov` — 25 passed, 20 deselected.
- Local: `PYTHONPATH="$PWD" mamba run -n CausalPy python -m pytest causalpy/tests/test_placebo_in_time.py -k 'plot_calibration' -q --no-cov` — 6 passed, 134 deselected; this covers the shared drawing helper's three-panel, sizing, title, datetime-label, palette-cycling, and missing-null paths.
- Local: `PYTHONPATH="$PWD" mamba run -n CausalPy python -m pytest causalpy/tests/test_public_signatures.py -q --no-cov` — 518 passed. This inventory catches variadics but does not enforce keyword-only controls on check-specific plot methods, so its pass does not resolve the signature finding.
- Local statistical probe: the multi-treated denominator changed expected p-values from `0.25` per treated unit to `0.4` and `0.2`; dropping an infinite placebo ratio changed `2/4` to `1/3`.
- Diff and workspace: `git diff --check` passed; the isolated worktree remained clean; the primary worktree retained its pre-existing `.agents/skills/review-pr/SKILL.md` modification.

REVIEWED | /Users/benjamv/git/CausalPy/.scratch/pr-reviews/2026-09-08-author-anevolbap/pr-1144.md

anevolbap:
All four findings are fixed.

**MSPE, not its square root.** The columns are `pre_mspe`, `post_mspe` and `mspe_ratio`, the metadata key is `baseline_mspe`, and the method is `plot_mspe_ratio`. The quantity is now the ADH ratio unchanged, so values compare directly to the published ones. No alias for the old names: `rmspe` appears on neither `main` nor `pymc6_and_pymcmarketing1_migration`, so nothing had shipped. `test_mspe_stats_matches_hand_computed_errors` pins the definition against hand-computed residuals: pre-period residuals 3 and 4 give 12.5, post-period 6 and 8 give 50, ratio 4. The RMSPE version of the same fixture is 2, so the test fails if the statistic drifts back.

**Reference set.** Each treated unit is ranked against the placebo pool plus itself, never against the other treated units. `test_plot_ranks_each_treated_unit_against_itself_only` is the probe from the review: placebo ratios 1, 2 and 3 with treated 5 and 10 now report 0.250 each instead of 0.4 and 0.2.

**Signature.** `*` after `check_result`, and a test asserts the positional call raises.

**Degenerate ratios.** Three cases, separated. A positive post-period error over a zero pre-period error is `inf`: defined and ranked, but not drawable on a finite axis, so it leaves the bars and stays in the denominator. Zero over zero, and any non-finite error, are `nan`: outside both the figure and the reference set. The review's probe now gives 0.5 rather than 1/3. Each case warns separately, and the docstring states that the p-value is conditional on the units that fitted successfully.

One change beyond the review. `_mspe` now reduces with `skipna=False`. xarray skips missing values by default, so a unit with a missing residual reported an error computed from the remaining points instead of a non-finite one, which left the new `nan` branch unreachable from that route.

Five questions from the first round are still open. None of them blocks a re-review, and the first is probably a child issue rather than scope here.

- The full-posterior MSPE. Residuals are averaged over the posterior before squaring, so each unit gets a scalar. A distribution over the ratio is reachable from the same `pre_impact`/`post_impact` arrays.
- The `isinstance` narrowing in `run()`, where `cast` would also satisfy mypy. `validate()` already rejects non-SC experiments, so the branch is unreachable in the pipeline; `isinstance` yields no baseline rather than an `AttributeError` if it ever is reached.
- Whether `docs/source/release_notes.md` wants a line for this. `CheckResult.text` gained a sentence and `metadata` went from empty to populated, both additive but user-visible. The 1.0.0 (unreleased) section is organised around the migration and its breaking changes, with no section for additions, and #1140 added `plot_calibration` without touching the file.
- Whether docstrings should start carrying `{cite:t}` roles. No Python docstring in the package uses sphinxcontrib-bibtex today, so the citation went into `sensitivity_checks.md` and the docstring names the paper in prose.
- Whether `causalpy/checks/_plot_helpers.py` is the right home as more checks get plots, versus importing `PlaceboInTime._draw` across check modules.

Does the conditional-on-fitted p-value need a stronger statement in `sensitivity_checks.md` than the docstring gives it, or does this land the statistical contract?

## Test evidence

- `causalpy/tests/test_method_specific_checks.py -k PlaceboInSpace`: 31 passed.
- Full suite: 2319 passed, 18 skipped.
- `diff-cover` against `pymc6_and_pymcmarketing1_migration`: 100% over 78 production lines.
- `prek run --all-files`: clean.
- Rebased onto `pymc6_and_pymcmarketing1_migration` at `cb7fefad`, no conflicts. All of the above ran on the rebased branch.

anevolbap:
Rebased onto `c2ff5eed`. The lazy experiment lifecycle on the base branch removed `pre_impact` and `post_impact` from `SyntheticControl`, and the branch still merged without a conflict, so the break was silent: on the merge result, 8 of the 31 `PlaceboInSpace` tests failed with `AttributeError`. The MSPE stats now read `result.impact_pre` and `result.impact_post`, and `run()` only records a baseline for a fitted experiment. The statistic is also computed outside the fit-failure handler, so a bug in it raises instead of turning every placebo into an error row.

The citation was wrong too: ADH (2010) has no section 5. The post/pre MSPE ratio and the 1/39 p-value are in section 3.4, with the ratios plotted in Figure 8. The docstrings, `sensitivity_checks.md` and the PR body now say that.

Full suite is 2522 passed, 18 skipped, and `diff-cover` reports 100% over 80 changed production lines. Good for another look?

cetagostini:
two things:
1. @anevolbap if you are done, let's merge.
2. Don't want to be bad or anything. But the "space" was kinda a vibe code extension of "in time", my understanding. Space feels less formal and experimental. Reading here, looks like gpts agree, and we are doing some adjustments for it. Given the fact this is our invention and not actual use of a paper into the library, could we mentioned?

cetagostini:
@anevolbap same question here. Should we merge?

anevolbap:
> @anevolbap same question here. Should we merge?

Yes, let's merge and take it from there.

Issue #1127, opened 2026-08-03
mypy is configured twice, the two configurations disagree, and neither gates anything


anevolbap:
Part of a deeper hardening pass over CI, pre-commit, packaging and pinning, done with Opus 5. This one is a decision rather than a patch, so it is an issue and not a PR. Related prior context: #66.

## What is true today, on `main`

Running `mypy` with no arguments, which is what `[tool.mypy]` describes:

```
Found 131 errors in 38 files (checked 16 source files)
```

Two separate things are going on there.

**Only 16 files are checked.** `files = "causalpy/*.py"` does not recurse, and `causalpy/` has exactly 16 top-level modules. Every subpackage is outside the configured scope: `experiments/` (15 modules), `checks/` (12), `steps/` (4), `data/` (3), `skills/` (1). The 38 files carrying errors are reached by import-following, not by the configured file set.

**Nothing runs it.** There is no `mypy` in any workflow or in the `Makefile`. The only thing that runs mypy is the pre-commit hook, which uses a different flag set (`--ignore-missing-imports`) and a different dependency set (`numpy>=1.20`, `pandas-stubs`) from `[tool.mypy]`.

The two also disagree about scope in both directions. `[tool.mypy]` sets `exclude = "...|tests|..."`, but the hook's `files: ^causalpy/` matches test files and passes them on the command line, where mypy's `exclude` does not apply. `prek run mypy --files causalpy/tests/conftest.py` runs and passes, so tests are type-checked in practice and excluded in configuration.

Versions are pinned on one side only. The hook pins `mirrors-mypy` at `rev: v2.3.0`, and the hook env it builds carries mypy 2.3.0. The `lint` extra declares a bare `mypy`, so a dev environment gets whatever was current when it was built, 1.20.2 in mine. The two versions will not agree on an error set.

Separately, #1124 removes five `[tool.mypy-matplotlib]`-style tables that use `mypy.ini` syntax and were never read by anything. It merged into `pymc6_and_pymcmarketing1_migration`, not into `main`, so those tables are still on `main` today.

## Why the hook is green

`--ignore-missing-imports` is not the main reason. The hook runs in an isolated env that contains only numpy and pandas-stubs, so pymc, xarray, sklearn, patsy and arviz are not present to analyse at all. Give mypy the hook's own flags inside an environment that does have the project dependencies, over the whole package rather than 16 files:

```
mypy --ignore-missing-imports causalpy
[...]

PR #1145, opened 2026-08-04
chore: make mypy a real gate with a per-module allowlist

anevolbap:
Closes the decision in #1127 by taking the gate option.

## What was wrong

`[tool.mypy]` set `files = "causalpy/*.py"`, which does not recurse, so it reached the 18 top-level modules and nothing under `experiments/`, `checks/`, `steps/`, `data/` or `skills/`. Nothing ran it: no workflow, no Makefile target. The only runner was the pre-commit hook, and that hook's env holds numpy and pandas-stubs only, so pymc, pytensor, arviz, patsy and sklearn are not importable and it passes on code that fails a real check. The two also sat at different mypy versions, the hook pinned at `mirrors-mypy` v2.3.0 and the `lint` extra declaring a bare `mypy`.

## What this does

`files = "causalpy"` recurses, so 54 source files are checked instead of 16. `make typecheck` runs mypy from the project environment, where the dependencies actually resolve, and the test job in ci.yml calls that target so it gates on every matrix leg. The pre-commit hook is removed, leaving one contract instead of two. mypy is pinned at 2.3.0 in `dev` and `lint`, and `environment.yml` is regenerated from that.

A third commit corrects two comments this PR introduced, the module count above and a claim that `warn_unused_configs` fails rather than warns, and unwraps the comment blocks per the AGENTS.md one-paragraph-per-line rule. No behaviour change.

## Why an allowlist

There are 49 errors in 16 files today. Gating on zero would mean fixing all of them first, in four experiment modules that are moving on this branch, and would leave the ~38 already-clean modules unprotected until that lands. Instead each failing module gets a `[[tool.mypy.overrides]]` entry that disables only the error codes it actually produces, so every other code is still checked there. The gate is green on merge, new code is covered immediately, and removing an entry (or one code from one entry) is a self-contained follow-up PR.

The largest entry by far is `inverse_propensity_weighting` at 21 errors, 17 of them `index`. The rest is a long tail of one or two per module.

## Not in this PR

No source files change, so nothing user-visible moves and there is no release note. The 49 errors are not fixed here, only fenced.

Tests stay outside the checked set, as `[tool.mypy] exclude` already had them. Including them would add 73 errors in 19 files, which is its own decision.

Two known limits of a config-level allowlist. `ignore_missing_imports` is global, so an unresolvable first-party import is silenced along with the untyped third parties; the test suite imports every module, so that case fails there instead. And mypy has no `warn_unused_ignores` equivalent for `disable_error_code`, so an entry can outlive the errors it fences. `warn_unused_configs = true` catches only the rename and deletion case, and it warns rather than failing: mypy prints `note: unused section(s)` and still exits 0, so it surfaces a stale entry in the log without turning the gate red on its own.

## Verification

- `mypy` with this config: `Success: no issues found in 54 source files`, under `--python-version 3.12` and `--python-version 3.14`.
- `prek run --all-files`: clean, including the regenerated `environment.yml`.
- `micromamba create --dry-run -c conda-forge python=3.14 mypy==2.3.0` resolves (`py314h518bba1_0`), so the pin does not break the 3.14 leg.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,6 +251,12 @@ jobs:
                   f"expected pandas {expected}, got {pd.__version__}"
               )
           PY
+      - name: Type check
+        # Runs here rather than in prek because mypy only says anything useful
+        # when the real dependencies are importable, and the pre-commit env has
+        # none of them (issue #1127). Scope and the per-module allowlist that
+        # keeps this green live in [tool.mypy] in pyproject.toml.
+        run: make typecheck
       - name: Run doctests
         # -p causalpy.tests.doctest_sampling loads the plugin that mocks
         # pm.sample for doctests (an autouse fixture in causalpy/tests/conftest.py
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,10 +252,7 @@ jobs:
               )
           PY
       - name: Type check
-        # Runs here rather than in prek because mypy only says anything useful
-        # when the real dependencies are importable, and the pre-commit env has
-        # none of them (issue #1127). Scope and the per-module allowlist that
-        # keeps this green live in [tool.mypy] in pyproject.toml.
+        # Runs here rather than in prek because mypy only says anything useful when the real dependencies are importable, and the pre-commit env has none of them (issue #1127). Scope and the per-module allowlist that keeps this green live in [tool.mypy] in pyproject.toml.
         run: make typecheck
       - name: Run doctests
         # -p causalpy.tests.doctest_sampling loads the plugin that mocks
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,13 +53,10 @@ repos:
     hooks:
       - id: numpydoc-validation
         files: ^causalpy/(?!tests/|data/).*\.py$
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v2.3.0
-    hooks:
-      - id: mypy
-        args: [--ignore-missing-imports]
-        files: ^causalpy/
-        additional_dependencies: [numpy>=1.20, pandas-stubs]
+  # Type checking is not a hook (issue #1127). It ran here against an isolated env holding
+  # only numpy and pandas-stubs, so pymc, pytensor, arviz, patsy and sklearn were not present
+  # to analyse and the hook passed on code that fails a real check. It now runs as
+  # ``make typecheck`` in the project environment, from the test job in ci.yml.
   - repo: https://github.com/abravalheri/validate-pyproject
     rev: v0.25
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,10 +53,7 @@ repos:
     hooks:
       - id: numpydoc-validation
         files: ^causalpy/(?!tests/|data/).*\.py$
-  # Type checking is not a hook (issue #1127). It ran here against an isolated env holding
-  # only numpy and pandas-stubs, so pymc, pytensor, arviz, patsy and sklearn were not present
-  # to analyse and the hook passed on code that fails a real check. It now runs as
-  # ``make typecheck`` in the project environment, from the test job in ci.yml.
+  # Type checking is not a hook (issue #1127). It ran here against an isolated env holding only numpy and pandas-stubs, so pymc, pytensor, arviz, patsy and sklearn were not present to analyse and the hook passed on code that fails a real check. It now runs as ``make typecheck`` in the project environment, from the test job in ci.yml.
   - repo: https://github.com/abravalheri/validate-pyproject
     rev: v0.25
     hooks:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -81,7 +81,8 @@ Renamed or deleted how-to pages must keep a **permanent** entry in `rediraffe_re
 
 ## Code quality checks
 
-- **Before committing**: Use `prek run` during iterative edits and run `prek run --all-files` before committing to ensure all checks pass (linting, formatting, type checking)
+- **Before committing**: Use `prek run` during iterative edits and run `prek run --all-files` before committing to ensure all checks pass (linting, formatting)
+- **Type checking**: `mypy` is intentionally **not** in `prek` — it only says anything useful when the project dependencies are importable, and a hook environment does not have them (issue #1127). Run `$CONDA_EXE run -n CausalPy make typecheck`; CI runs the same target from the test job. Scope and the per-module allowlist live in `[tool.mypy]` in `pyproject.toml`.
 - **Patch coverage (milestones, not every commit)**: `make test-patch-cov` is intentionally **not** in `prek` — it runs the full test suite (~90s+) and needs the conda env. Run it at meaningful checkpoints when Python source or tests under `causalpy/` changed, not on every small commit:
   - A PR is ready for review, or you are about to push for CI
   - You finished a large or multi-file task touching production code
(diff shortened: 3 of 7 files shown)


anevolbap:
Is `daimon-pymclabs` something contributors can request a review from, or is it kept for maintainer-authored PRs? Asking because this is the base of a four-PR stack, and one adversarial pass over the whole thing may be more useful than four separate reads.

drbenvincent:
## Automated triage

**Why:**
- Cross-cutting change: modifies CI workflow (`.github/workflows/ci.yml`), Makefile, pre-commit config, `pyproject.toml`, `environment.yml`, and developer docs (AGENTS.md, CONTRIBUTING.md).
- Replaces the broken pre-commit-only mypy check with a real gating `make typecheck` target that runs in every CI matrix leg.
- Introduces a 49-error per-module allowlist — this is a scaffolding decision that affects all future PRs landing on this branch.
- All CI checks pass across 3 Python versions and both pandas versions.

**Review focus:**
1. Verify the allowlist entries cover only the error codes actually produced by each module (no missing or stale entries).
2. Confirm `--python-version` gating for both 3.12 and 3.14 is not overly permissive.
3. Check the Makefile target does not introduce a circular dependency or slow the CI pipeline unduly.

**Confidence:** high

drbenvincent:
Sorry for the delays - been occupied. You should receive a batch of agent generated reviews soon :)

drbenvincent (approved):
Agent-drafted review.

# PR 1145 review

- **PR:** https://github.com/pymc-labs/CausalPy/pull/1145
- **Title:** chore: make mypy a real gate with a per-module allowlist
- **Author:** anevolbap
- **Base:** `pymc6_and_pymcmarketing1_migration` (`a2f13029`)
- **Head:** `chore/mypy-gate` (`7f95809e`)
- **Classification:** tests / CI / packaging / infrastructure (process docs included)
- **Reviewed against:** declared base only (not `main`, not descendant mypy PRs)

## Executive Summary
Recommendation: approve.

Value delivered:
- Mypy now recurses over the package (`files = "causalpy"`), so all 54 production modules are in scope instead of the 18 top-level files matched by `causalpy/*.py`.
- The hollow pre-commit hook (isolated env with only numpy and pandas-stubs) is replaced by `make typecheck` in the project environment, invoked from every CI test-matrix leg, with mypy pinned at 2.3.0 in `dev`/`lint` and regenerated `environment.yml`.
- Existing failures are fenced with per-module `disable_error_code` (not blanket `ignore_errors`), so already-clean modules are protected immediately. Local measurement on this head: the same 16 files and the same error codes as the allowlist; 50 errors with local mypy 2.3.1 versus the comment’s 49 at `194343ca`.

Review focus:
- Highest-risk assumption: the allowlist is only as wide as today’s errors, and CI actually runs a useful mypy. Probe result: every listed code is produced; no extra files or codes; `make typecheck` is green locally (54 files) and the Type check step succeeded on all three CI legs (3.12/pandas 3.x, 3.12/pandas 2.3, 3.14/pandas 3.x), ~15–20s each.

## Findings

- [nit] `pyproject.toml`: the allowlist comment still says “49 errors … measured … at 194343ca”. On current head the error *codes and modules* match; the count is 50 under local mypy 2.3.1. Harmless drift; descendants that shrink the list will rewrite this anyway.
- [nit] `.github/workflows/ci.yml`: typecheck runs on both Python 3.12 pandas legs. 3.12 vs 3.14 is justified because the Makefile does not pin `--python-version`. The second 3.12 pandas leg is redundant for typing.
- [nit] `AGENTS.md`: “Before committing” no longer names typecheck. The new bullet explains how to run it, but an agent that only follows the prek sentence will wait for CI. Optional follow-up: mention `make typecheck` next to prek when `causalpy/` source changes.

What worked well:
- Third commit walked back the overclaim that `warn_unused_configs` *fails* the gate; it now correctly says mypy prints `note: unused section(s)` and exits 0.
- `disable_error_code` per module is the right ratchet (stricter than issue #1127’s `ignore_errors` sketch). Grouped tables only share codes every member actually emits.
- Hook removal plus CONTRIBUTING/`make typecheck` is an honest single contract; `environment.yml` is generated (check-environment-yml passed), not hand-edited as a source of truth.
- Typecheck is placed after `pip install --no-deps -e .` and the pandas-leg assertion, before doctests, so it uses the real env and fails cheaply.

## Merge Readiness
Verdict: approve.
Branch status: up to date with `pymc6_and_pymcmarketing1_migration` (0 behind, 3 unique commits); GitHub `mergeable` / `CLEAN`; no reviews posted.
CI status: green on observed checks (ci test matrix, prek, check-environment-yml, notebooks, zizmor, codecov/patch, RTD, pre-commit.ci). Last complete run is 2026-08-18; remote base SHA is unchanged since then.

Residual risk (accepted, documented in the PR):
- Config-level `disable_error_code` has no `warn_unused_ignores` equivalent, so a code can rot after a fix until a human removes it. `warn_unused_configs` only catches renamed/deleted *modules*, and does not fail CI.
- `ignore_missing_imports` is global (third-party stub noise). First-party missing modules would be silenced here and caught by tests instead.
- Tests remain excluded (73 extra errors called out as a separate decision).
- pandas-stubs from the old hook are not in the project env; with real pandas installed this is still a net stronger check than the hollow hook.

## PR Summary
This PR implements the “gate” decision from #1127 without fixing the existing mypy errors: recursive `[tool.mypy] files`, a Makefile target, CI invocation in the micromamba test job, removal of `mirrors-mypy`, a 2.3.0 pin, and a 16-module allowlist so the gate starts green. No production source changes and no user-facing API impact. The trade-off is a shrinking suppressions list rather than a red tree until the later mypy-stack PRs land.

## Test Evidence
Local (isolated worktree `/Users/benjamv/git/.worktrees/pr-1145`, not the primary tree):
- `mamba run -n CausalPy make typecheck` → `Success: no issues found in 54 source files` (env mypy 2.3.1; CI pin is 2.3.0).
- 54 `causalpy/**/*.py` files excluding tests, matching mypy’s count.
- Mypy with overrides stripped: 50 errors in 16 files; codes match the allowlist exactly (IPW 21 with 17 `index`; grouped `arg-type` / `attr-defined` modules each emit those codes).

Remote:
- Type check step success on all three Test jobs (run `32174233098`).
- `gh pr checks 1145`: all listed checks pass.
- `check-environment-yml` pass confirms `mypy==2.3.0` in `environment.yml` comes from pyproject2conda (`dev`+`test` extras).

Not run: full pytest, `prek --all-files` (no source change; remote prek already green).

## Open Questions
None that affect merge of this PR against its declared base. Shrinking or deleting allowlist entries belongs to later stack PRs, not this one.

## Review brief (process)

Value hypothesis: real recursive mypy, one runner, green-on-merge via a tight allowlist.
Risk thesis: over-broad suppressions or a typecheck step that does not actually see project deps.
Attack plan:
1. Allowlist wider than produced codes — falsified by a no-override mypy dump.
2. `--python-version` / Makefile mismatch leaving 3.12 or 3.14 ungated — falsified: `mypy` uses the job interpreter; both 3.12 and 3.14 Type check steps passed.
3. Hook removal plus prek-only local workflow weakens signal — partially true vs a *working* hook, but the old hook was a false green; CI test job is the real gate.
4. `environment.yml` hand-edit — falsified by the sync job and pyproject2conda extras.

No GitHub comments were posted.

anevolbap:
Quick question: once it's approved, should I merge it, or leave that to the maintainers?

drbenvincent:
I think merging into the v6 branch is fine to just go ahead with - as long as it's gone through a round of review then we can be reasonably confident with it. 

Not yet decided, but we might to a _pre_ 1.0.0 release with v6 compatibility, and then do the 1.0.0 release once _all_ the core API changes we want are done. But see how it pans out. Thanks for all the work on this so far :)

PR #1146, opened 2026-08-04
fix: keep plot_ate axes as a list so the panels type-check

anevolbap:
Stacked on #1145. First allowlist entry to shrink.

`plot_ate` called `plt.subplot_mosaic`, which returns `dict[str, Axes]`, then rebound the same name to a list of the panels on the next line. mypy keeps the dict type for the rest of the function, so every positional `axs[0]`, `axs[1]` and `axs[2]` was an `index` error and the declared `tuple[Figure, list[Axes]]` return was a `return-value` error. That single rebinding accounted for 19 of the 49 errors on the base branch.

The mosaic result now keeps its own name and the list takes the old one. Iteration over the mosaic dict is insertion order, A then B then C, exactly as before, so the panels keep their meaning and nothing renders differently.

`inverse_propensity_weighting` drops out of its dedicated `[[tool.mypy.overrides]]` entry and joins the plain `arg-type` group, which is the two remaining errors where `fit` is called with numpy arrays against a `DataArray` annotation. Those are the public-signature question and are left alone here.

Branch total: 49 errors in 16 files becomes 30 in 16.

## Verification

- `mypy` green under `--python-version 3.12` and `3.14`.
- `causalpy/tests/test_ipw_extreme_ps.py` and `test_ipw_get_ate.py`: 39 passed. The first is the #645 regression suite and drives `plot_ate` for every weighting scheme.
- `test_integration_pymc_examples.py::test_inverse_prop` passed; it asserts `isinstance(axs, list)` and that each element is an `Axes`, which is the contract this preserves.

Being stacked on a feature branch, this gets the repo-wide workflows but not the test matrix. That arrives when #1145 merges and this retargets.

### DIFF
--- a/causalpy/experiments/inverse_propensity_weighting.py
+++ b/causalpy/experiments/inverse_propensity_weighting.py
@@ -669,8 +669,9 @@ class InversePropensityWeighting(BaseExperiment):
         mosaic = """AAAAAA
                     BBBBCC"""
 
-        fig, axs = plt.subplot_mosaic(mosaic, figsize=(20, 13))
-        axs = [axs[k] for k in axs]
+        fig, mosaic_axes = plt.subplot_mosaic(mosaic, figsize=(20, 13))
+        # Keyed by mosaic label; the panels are used positionally from here on.
+        axs = [mosaic_axes[k] for k in mosaic_axes]
         axs[0].axvline(
             0.1, linestyle="--", label="Low Extreme Propensity Scores", color="black"
         )
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -236,12 +236,9 @@ warn_unused_configs = true
 
 # Allowlist of modules that already fail, so the gate starts green and ratchets down (49 errors in 16 files, measured on pymc6_and_pymcmarketing1_migration at 194343ca). Each entry disables only the error codes that module actually produces, so everything else is still checked there. Removing an entry, or a single code from one, is a self-contained follow-up PR.
 [[tool.mypy.overrides]]
-module = "causalpy.experiments.inverse_propensity_weighting"
-disable_error_code = ["arg-type", "assignment", "index", "return-value"]
-
-[[tool.mypy.overrides]]
 module = [
     "causalpy.experiments.instrumental_variable",
+    "causalpy.experiments.inverse_propensity_weighting",
     "causalpy.experiments.prepostnegd",
     "causalpy.experiments.regression_discontinuity",
     "causalpy.experiments.regression_kink",


drbenvincent:
## Automated triage

**Why:**
- Fixes a variable-rebinding issue in `plot_ate` (2 files changed: the IPW module and pyproject.toml allowlist). Mechanical, no behavioural change.
- 19 mypy errors resolved by keeping the `plt.subplot_mosaic` result under its own name and the list under the original variable.
- All CI checks pass. Stacked on #1145; tests pass and integration test confirms `isinstance(axs, list)`.

**Review focus:**
1. Confirm the mosaic dict iteration order (A, B, C) matches the original panel ordering.
2. Verify the remaining `arg-type` allowlist entry for IPW .fit() calls is correctly scoped.

**Confidence:** high

drbenvincent (approved):
Agent-drafted review.

# PR #1146 unit review

- PR: https://github.com/pymc-labs/CausalPy/pull/1146
- Title: fix: keep plot_ate axes as a list so the panels type-check
- Author: anevolbap
- Base: `chore/mypy-gate` (#1145) → Head: `chore/mypy-ipw-axes` (`38d2ef2`)
- Scope reviewed: stacked diff only (`origin/chore/mypy-gate...HEAD`), 2 files, +4/−6
- Classification: typing bug-fix plus allowlist ratchet (tests/CI infra)

Private review brief used before the diff pass:

- Value hypothesis: `plot_ate` type-checks as `tuple[Figure, list[Axes]]`; IPW loses `assignment`/`index`/`return-value` allowlist codes; runtime panel order and list contract stay the same.
- Risk thesis: mosaic-dict iteration order vs positional `axs[0..2]`, and an over/under-scoped allowlist move.
- Attack plan: (1) mosaic key order is not A, B, C; (2) remaining IPW errors are not only `arg-type`; (3) joining the shared `arg-type` group silences extra codes or leaves `index`/`return-value` live.

## Executive Summary
Recommendation: approve

Value delivered:
- Splitting `plt.subplot_mosaic` onto `mosaic_axes` and building `axs` as a new list removes the mypy rebinding trap. With the allowlist stripped, IPW `assignment` / `index` / `return-value` errors on `plot_ate` disappear; the declared `tuple[Figure, list[Axes]]` return matches how callers already unpack the result.
- `causalpy.experiments.inverse_propensity_weighting` leaves its dedicated override (`arg-type`, `assignment`, `index`, `return-value`) and joins the shared `arg-type`-only group. That is a real ratchet, not a silenced check: those three extra codes would now fail the gate if they came back.

Review focus:
- Highest-risk assumption was that iterating the mosaic dict still yields panels A, then B, then C (top histogram, outcomes, ATE). Local matplotlib reproduces `keys: ['A', 'B', 'C']` and `axs[i] is mosaic_axes[label]`. That is the same iteration the old one-name rebinding used, so this is not a new ordering dependency.

## Findings

No must-fix, should-fix, or blocking nits.

Probes that did not become findings:

- Mosaic order: `subplot_mosaic("""AAAAAA\\nBBBBCC""")` inserts A, B, C; positional indexing matches the docstring (top / bottom-left / bottom-right). Explicit `mosaic_axes["A"]` lookups would be slightly more obvious, but they would not change runtime behavior relative to the previous `[axs[k] for k in axs]`.
- Remaining IPW mypy: with IPW removed from the override, only two `arg-type` errors remain, both at `algorithm()` line 138 (`self.model.fit(X=self.X, t=self.t, ...)`). `X` is `ndarray` against `PyMCModel`/`PyMCForecastModel` `DataArray` annotations. The existing `# type: ignore[call-arg]` covers the extra `t=` keyword, not `arg-type`. Leaving that for a public-signature follow-up matches the PR claim.
- Sibling `plot_balance_ecdf` uses `plt.subplots` and already returns a list-like of Axes; no parallel rebinding bug.
- Consumers (`test_integration_pymc_examples.py` asserts `isinstance(axs, list)` and each element is `Axes`; `test_ipw_extreme_ps.py` indexes `axs[0]`) already pin the list contract. A new runtime test would not fail on the old code; mypy is the regression oracle here (MF-3 does not apply as a must-fix).
- Allowlist comment still cites the gate-PR snapshot (“49 errors in 16 files”). Historical, not a contract. Measured without any overrides in this environment: 50 errors on `chore/mypy-gate` vs 31 on this head (same 19-error drop the PR reports as 49→30). Drift is stub/tooling, not an extra unfixed IPW error.

What worked well:
- One-name split is the smallest change that actually changes mypy’s inferred type; no casts, no `# type: ignore` on the plot path.
- Allowlist edit is correctly narrower: IPW is no longer special-cased for plot-index errors.

## Merge Readiness
Verdict: approve
Branch status: up to date with `chore/mypy-gate` (0 behind, 1 commit ahead); GitHub `MERGEABLE` / `CLEAN`; no review comments to address.
CI status: green. Observed passing: Test (Python 3.12 pandas 2.3 / 3.x, Python 3.14 pandas 3.x), prek, pre-commit.ci, notebooks (all four jobs), check-environment-yml, zizmor, codecov/patch, Read the Docs. The PR text said the stacked branch would miss the test matrix; current `gh pr checks` shows the matrix ran and passed.

Depends on #1145 remaining mergeable; this review does not re-litigate that base.

## PR Summary
`plot_ate` rebound `axs` from `dict[str, Axes]` to `list[Axes]`. mypy kept the dict type, so every integer index and the annotated list return were errors (the bulk of IPW’s allowlisted codes). The mosaic dict now keeps its own name; the list is a separate variable. Runtime construction is the same comprehension over insertion-ordered keys. The only remaining IPW suppressions are the two `fit` `arg-type` mismatches, which belong to a later signature/adapter change.

## Test Evidence
Local (isolated worktree `/Users/benjamv/git/.worktrees/pr-1146`, not the primary checkout):

- `mamba run -n CausalPy mypy --python-version 3.12` and `3.14` on the PR tree with the committed allowlist: Success, 54 files.
- Same mypy with all `[[tool.mypy.overrides]]` stripped: 31 errors / 16 files on head vs 50 / 16 on `origin/chore/mypy-gate`.
- IPW-only with IPW dropped from the arg-type group: two `arg-type` errors at line 138.
- Matplotlib mosaic key-order check: A, B, C mapped to `axs[0:3]`.
- `PYTHONPATH` import of the worktree module confirmed `plot_ate` contains the `mosaic_axes` split and still annotates `tuple[Figure, list[Axes]]`.

Not run locally: pytest. The shared CausalPy env is an editable install of the primary worktree (`issue-1092-lazy-experiment-lifecycle`); `make setup` in the PR worktree would retarget that env. Runtime is a rename-equivalent of the old list build. Remote CI already ran the test matrix, including IPW-heavy jobs.

Remote: all `gh pr checks 1146` entries pass; Codecov reports no coverage change on the modified lines.

## Open Questions
None that affect merge readiness. Stack merge still waits on #1145.

## Draft comments (not posted)

Approve-style summary if a maintainer wants to post later:

This is a real mypy root-cause fix, not a suppression. Keeping the mosaic dict and the panel list on different names drops the `index`/`assignment`/`return-value` errors, preserves the public `list[Axes]` contract, and correctly leaves only the `fit(..., X=ndarray)` `arg-type` pair on the IPW allowlist. Mosaic insertion order is still A, B, C. Happy to approve once #1145 is in.

REVIEWED | /Users/benjamv/git/CausalPy/.scratch/pr-reviews/2026-09-08-author-anevolbap/pr-1146.md

PR #1148, opened 2026-08-04
fix: clear the mechanical type errors and shrink the allowlist to one open question

anevolbap:
Stacked on #1146. Takes the allowlist from nine entries to three, and the branch from 30 errors in 13 files to 10 in 5.

A final commit unwraps the comments this PR adds, per the AGENTS.md one-paragraph-per-line rule. No behaviour change.

Four clusters, not 20 separate bugs. In every case the annotation was narrower than the value that actually reaches it, and the runtime behaviour was already correct.

`round_num` declared `n: float` while four call sites pass a zero-dimensional DataArray from `.mean()` on posterior samples. xarray formats those like the scalar they wrap, so this was purely the annotation. Widened to `float | xr.DataArray`, matching its sibling `convert_to_string`, which already declared that union. Clears 4 errors in four modules.

Six constructors take `data: DataFrameLike` and rebind that same parameter to the pandas frame `to_pandas` returns. mypy keeps the declared parameter type, so `.index`, `.copy()` and `data[column]` were all reported against `NativeDataFrame`. The conversion helpers were already annotated `-> pd.DataFrame`; the information was lost at the rebinding. The converted frame now gets its own name, which is also what the existing comments in those constructors already claimed. Clears 9 errors.

Then five one-liners: `np.abs` on a DataArray replaced by the builtin `abs`, which xarray types correctly and which computes the same thing; xarray dimension names sorted as strings, since `Hashable` is not orderable and they only go into an error message; the ETI branch of `_interval_bound_values` given its own names so it stops colliding with the array-typed HDI branch; a dimension name cast to `str` for a coords key that PyMC requires to be a string anyway; and a list turned into the 4-tuple `tight_layout(rect=...)` documents.

One inline ignore, on iterating the "bodies" entry of a violinplot result. matplotlib types every entry as a single Collection rather than a list of them, so this is a stub limitation rather than our bug, and `warn_unused_ignores` will flag the ignore once matplotlib describes the return properly.

## What is left, and why

The remaining 10 errors are one question, not a backlog. Several attributes and parameters are declared wider than the values that reach them, and narrowing them means deciding what the model lifecycle promises:

- `self.model` is declared as the base model union on `inverse_propensity_weighting` and `instrumental_variable`, while both reach `PropensityScore.fit` and `InstrumentalVariableRegression.fit`, which take numpy arrays and extra keywords (6 errors). Both call sites already carry `type: ignore[call-arg]` for this, without `arg-type`. Narrowing the attribute to the subclass each experiment requires is the real fix, and it is a public type on the experiment classes.
- `self.idata` starts as `None` and is a `DataTree` at every point that returns it (2 errors), and `self.causal_impact` is the same shape (1 error).
- `Axes.get_figure` is Optional for a detached artist, which cannot happen at that call site (1 error).

I did not want to pick an answer to the first one inside a cleanup PR, and asserting non-None at the other three would be adding runtime checks for states that cannot occur just to satisfy the checker. Happy to do whichever you prefer.

Since opening this, #1149 takes a position on all four and gets the allowlist to empty, each decision marked in the code. It is a draft, and it stacks on this branch, so this PR still stands on its own if you would rather stop here.

## Verification

- `mypy` green under `--python-version 3.12` and `3.14`.
- Full suite: `2294 passed, 18 skipped, 3 deselected`, on the rebased branch.
- `prek run --all-files` clean, including the regenerated `environment.yml`.

Stacked on a feature branch, so this gets the repo-wide workflows but not the test matrix until the stack merges down.

### DIFF
--- a/causalpy/_arviz_compat.py
+++ b/causalpy/_arviz_compat.py
@@ -202,7 +202,9 @@ def hdi_bounds(
     ).squeeze(drop=True)
     remaining_dims = set(result.dims) - {"hdi"}
     if remaining_dims:
-        msg = f"Scalar HDI bounds require reduced draws; remaining dims={sorted(remaining_dims)!r}"
+        # Dimension names are Hashable in xarray's typing, so sort them as strings.
+        dim_names = sorted(str(dim) for dim in remaining_dims)
+        msg = f"Scalar HDI bounds require reduced draws; remaining dims={dim_names!r}"
         raise ValueError(msg)
     return float(result.sel(hdi="lower").item()), float(result.sel(hdi="higher").item())
 
--- a/causalpy/checks/operating_characteristics.py
+++ b/causalpy/checks/operating_characteristics.py
@@ -555,7 +555,7 @@ class OperatingCharacteristics:
         )
         ax.yaxis.set_major_formatter(plt.FuncFormatter(lambda value, _: f"{value:.0%}"))
         ax.legend(loc="lower right", fontsize=8, framealpha=0.9, edgecolor="#cbd5e1")
-        figure.tight_layout(rect=[0, 0, 1, 0.97] if has_prior else None)
+        figure.tight_layout(rect=(0, 0, 1, 0.97) if has_prior else None)
         return figure
 
 
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -119,8 +119,7 @@ class DifferenceInDifferences(BaseExperiment):
     ) -> None:
         super().__init__(model=model)
         self.causal_impact: xr.DataArray | float | None
-        # to_pandas returns a copy, so index metadata is normalized on an
-        # owned frame rather than the caller's.
+        # to_pandas returns a copy, so index metadata is normalized on an owned frame rather than the caller's.
         pandas_data = to_pandas(data)
         pandas_data.index.name = "obs_ind"
         self.data = pandas_data
@@ -506,8 +505,7 @@ class DifferenceInDifferences(BaseExperiment):
                     showmedians=False,
                     widths=0.2,
                 )
-                # violinplot types every entry as a single Collection, but the
-                # "bodies" entry is a list of them.
+                # violinplot types every entry as a single Collection, but the "bodies" entry is a list of them.
                 for pc in parts["bodies"]:  # type: ignore[attr-defined]
                     pc.set_facecolor("C0")
                     pc.set_edgecolor("None")
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -506,7 +506,9 @@ class DifferenceInDifferences(BaseExperiment):
                     showmedians=False,
                     widths=0.2,
                 )
-                for pc in parts["bodies"]:
+                # violinplot types every entry as a single Collection, but the
+                # "bodies" entry is a list of them.
+                for pc in parts["bodies"]:  # type: ignore[attr-defined]
                     pc.set_facecolor("C0")
                     pc.set_edgecolor("None")
                     pc.set_alpha(0.5)
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -121,9 +121,9 @@ class DifferenceInDifferences(BaseExperiment):
         self.causal_impact: xr.DataArray | float | None
         # to_pandas returns a copy, so index metadata is normalized on an
         # owned frame rather than the caller's.
-        data = to_pandas(data)
-        data.index.name = "obs_ind"
-        self.data = data
+        pandas_data = to_pandas(data)
+        pandas_data.index.name = "obs_ind"
+        self.data = pandas_data
         self.expt_type = "Difference in Differences"
         self.formula = formula
         self.time_variable_name = time_variable_name
--- a/causalpy/experiments/interrupted_time_series.py
+++ b/causalpy/experiments/interrupted_time_series.py
@@ -166,10 +166,10 @@ class InterruptedTimeSeries(BaseExperiment):
         self.post_design: xr.Dataset
         # to_pandas_with_time_index returns a copy, so index metadata is
         # normalized on an owned frame rather than the caller's.
-        data = to_pandas_with_time_index(data, time_column)
-        data.index.name = "obs_ind"
-        self.data = data
-        self.input_validation(data, treatment_time, treatment_end_time)
+        pandas_data = to_pandas_with_time_index(data, time_column)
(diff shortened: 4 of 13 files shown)


drbenvincent:
## Automated triage

**Recommendation:** `review:high` — maintainer decision needed.

**Why:**
- Reduces the mypy allowlist from 9 entries to 3 and errors from 30 in 13 files to 10 in 5. Clear mechanical improvements (variable rebinding, widened type unions).
- The author explicitly flags 3 unresolved clusters requiring public-type decisions on `self.model`, `self.idata`, and `self.causal_impact` — these are maintainer-significant API choices.
- Stacked on #1146 and target feature branch `pymc6_and_pymcmarketing1_migration`. CI is UNSTABLE (readthedocs pending).
- Cross-sections 13 source files across the core library, including experiment classes and plot utils.

**Review focus:**
1. Decide the `self.model` type narrowing strategy for IPW and IV experiment classes (the 6-call-arg error cluster).
2. Decide whether to add `assert` runtime checks for the `None`-on-`Optional` value errors or accept inline ignores.
3. Confirm the stacked branch order is correct before merging.

**Confidence:** high

drbenvincent (approved):
Agent-drafted review.

# PR #1148 Review

https://github.com/pymc-labs/CausalPy/pull/1148

Scope: `chore/mypy-ipw-axes` (`38d2ef2f`, reviewed #1146) .. `chore/mypy-shrink-allowlist` (`a758a563`). Isolated worktree `/Users/benjamv/git/.worktrees/pr-1148`. Primary tree left on `issue-1092-lazy-experiment-lifecycle` with its existing dirty skill file.

## Executive Summary
Recommendation: approve.

Value delivered:
- The mypy allowlist ratchets from eight override tables to three, and from **31 errors in 16 files** on the #1146 base to **11 errors in 5 files** on this head, with `make typecheck` green under the remaining overrides.
- Annotations now match values the code already accepted at runtime: `round_num` / `_format_sig_figs` take `float | xr.DataArray` like sibling `convert_to_string`; six constructors bind `to_pandas` / `to_pandas_with_time_index` results as `pandas_data` so mypy sees a real `pd.DataFrame`; `abs(effect)`, string-sorted dim names, ETI bounds with their own names, `str(dimension)` coord keys, and a `tight_layout(rect=...)` 4-tuple clear the rest without a new public contract.

Review focus:
- Highest-risk assumption is “no runtime behaviour change.” Traced every helper/caller seam below; existing tests and worktree probes did not show a user-visible change. Residual typing debt (model `fit` numpy vs `DataArray`, optional `idata` / `causal_impact`, optional `get_figure`) is named in the allowlist comment and deferred, including to draft #1149.

## Findings

No must-fix or should-fix items.

- [nit] `pyproject.toml` / PR body: remaining-error counts are slightly stale versus a current no-allowlist run (body: 30→10 in 13→5 files; measured: 31→11 in 16→5). The extra leftover is `diff_in_diff.py:594` `annotate(..., xy=)` `arg-type`, already covered by the DID `arg-type` override kept for `causal_impact: ... | None`. Does not affect the ratchet or merge readiness.

## Merge Readiness
Verdict: approve.
Branch status: 4 commits ahead of `chore/mypy-ipw-axes`, 0 behind; GitHub `MERGEABLE` / `CLEAN`.
CI status: green (tests on 3.12/3.14, prek, notebooks, docs, codecov/patch, zizmor, sync).

## PR Summary
This stack PR is a mechanical typing cleanup on top of #1146: widen `round_num` to the 0-d `DataArray` four call sites already pass from `.mean()`, stop rebinding converted frames onto a `DataFrameLike` parameter in six experiment constructors, and apply five one-line typing fixes plus one matplotlib stub `type: ignore[attr-defined]` on violinplot `bodies`. The trade-off is leaving model-lifecycle type narrowing (IV/IPW `fit` arrays, optional `idata` / `causal_impact`, optional figure) for a follow-up rather than asserting states the checker cannot prove. No public `__all__` signature change; `round_num` remains an internal helper.

## Test Evidence
Remote: all `gh pr checks` for #1148 passing, including the Python test matrix (the PR body’s “no test matrix until the stack merges down” note is outdated relative to current checks).

Local, on the isolated worktree with `PYTHONPATH=/Users/benjamv/git/.worktrees/pr-1148` (shared conda env was not retargeted with `make setup`):
- `mamba run -n CausalPy make typecheck` on head: Success, 54 files.
- Same on #1146 worktree: Success under that branch’s wider allowlist.
- mypy without overrides: base 31/16 files → head 11/5 files; remaining codes match the three leftover override groups (`arg-type` on DID/IV/IPW, `return-value` on `pymc_models`, `assignment` on operating characteristics). `call-overload` on `setdefault` and OC `tight_layout` `arg-type` are gone.
- Runtime: `round_num(float)` vs `round_num(0-d DataArray)` identical for several values including 0; `abs(effect) > min_effect` equals `np.abs(effect) > min_effect` and the same ROPE probability; string dim names sort and `setdefault` the same before/after `str(dimension)`.
- pytest `--no-cov` / targeted (54 collected + 2 extra): `test_round_num`, three `_compute_rope_probability` tests, HDI/ETI ribbon tests, ETI extra-dimension rejection, input `to_pandas` / time-column tests — all passed.

## Open Questions
None that affect merge of this PR. Whether to narrow `self.model` / `idata` / `causal_impact` / `get_figure` is explicitly out of scope here and is the subject of #1149.

## Probe notes (review brief)

Private hypotheses and outcomes:

1. `round_num` / `_format_sig_figs` on 0-d `DataArray` would format or truth-test differently than float — **falsified** at the four `.mean()` call sites (`regression_kink` ×2, `regression_discontinuity`, `prepostnegd`); `f"{n:.Ng}"` and zero handling matched floats.
2. Constructor rename would leave a later use of original `data` (skipping conversion) or fail to pass the converted frame into validation — **falsified**; ITS/SC/SDID pass `pandas_data` into `input_validation`; panel uses it for `.copy()` and `nunique`; DID/staggered only store `self.data`.
3. `abs` vs `np.abs` would change ROPE indicators — **falsified**; types stay `DataArray` and `test_compute_rope_probability_two_sided` still expects 0.4.
4. ETI `to_numpy()` before the shared tail would change dims/length checks — **falsified**; `.dims` is still read on the DataArray; `test_plot_posterior_over_x_eti_rejects_multiple_preserved_dimensions` still matches `exactly one preserved dimension`.
5. Allowlist shrink would drop a still-failing module or leave unused override codes — **falsified**; typecheck is green; remaining errors sit only in the three kept groups. Violinplot ignore is the real `Collection` / `__iter__` `attr-defined` error from the base inventory.

Draft GitHub review (not posted): none required beyond approval.

REVIEWED /Users/benjamv/git/CausalPy/.scratch/pr-reviews/2026-09-08-author-anevolbap/pr-1148.md

PR #1060, opened 2026-07-22
Canonicalize model coefficient containers

drbenvincent:
## Summary

- Normalize every supported model's coefficients to an xarray container with `chain`, `draw`, `coeffs`, and optional `treated_units` dimensions.
- Preserve PyMC posterior draws and embed sklearn point estimates as singleton draws.
- Replace backend-specific coefficient printing and maketables extraction with shared, draw-aware consumers.

Fixes #1050

## Changes

- Move PyMC coefficient variable and dimension normalization into `ModelAdapter.coefficients()`.
- Preserve predictor and treated-unit coordinates in sklearn coefficient containers.
- Migrate DiD, PrePostNEGD, panel plots/summaries, and maketables to the canonical contract.
- Keep pymc-forecast `coefficients()` unsupported (scalar/state-space params are not design-matrix coefficients); restore forecast-specific `print_coefficients()` so ITS `.summary()` still prints scalar posterior parameters.
- Add coefficient contract tests covering dimensions, coordinates, singleton embedding, posterior preservation, formatting, and invalid containers.
- Document the coefficient contract alongside prediction normalization.

## Review follow-up

Addressed review feedback (commit `59e660c1`): restored `PyMCForecastModel.print_coefficients()` and `PyMCForecastAdapter.print_coefficients()` so forecast-backed ITS `.summary()` no longer raises through the base `coefficients()` path. Rebased onto current `main`.

## Testing

- [x] `mamba run -n CausalPy prek run --all-files`
- [x] Full test suite: 1,230 passed, 5 skipped (pre-rebase run)
- [x] Patch coverage: 100% (pre-rebase run)
- [x] `mamba run -n CausalPy make html`
- [x] Forecast adapter smoke test after review fix

## Checklist

- [x] Backend variation is normalized at the adapter seam.
- [x] Downstream coefficient consumers dispatch on draw count, not backend type.
- [x] No new dependency was added.

## Release notes (for next version)

- **Internal:** `ModelAdapter.coefficients()` now returns a canonical xarray draw container for PyMC and sklearn backends. Downstream code branches on draw count, not backend type.
- **User-visible (minor):** OLS DiD `causal_impact` is now a 0-d `xarray.DataArray` instead of a Python `float` (same pattern as prior prediction/score canonicalization PRs). Plot/table paths handle this internally.
- **pymc-forecast:** `coefficients()` remains unsupported; `.summary()` behaviour unchanged (scalar parameter table).

## Follow-ups

[...]

PR #1149, opened 2026-08-04
fix: empty the mypy allowlist by settling the four lifecycle types

anevolbap:
Stacked on #1148. Draft on purpose: every one of these is a judgement call I made rather than one you asked for, and each is marked with a `DECISION (#1127)` comment at the line where it was made, so you can argue with any of them in place without reading the whole diff.

The package now type-checks clean with an empty allowlist.

## The one I got wrong before

I described the six `fit` errors in #1148 as "either the signature is too narrow, or those two experiments are misusing it". Neither is true. `PyMCModel.fit` rejects anything that is not a DataArray at runtime, and these call sites reach `PropensityScore.fit` and `InstrumentalVariableRegression.fit`, which take numpy arrays plus extra keywords. Both lines already carried `type: ignore[call-arg]` for exactly this reason; the ignores just did not list `arg-type`.

So the decision is: add `arg-type` to the ignores that were already there. The real fix is narrowing `self.model` on those two experiments from the base model union to the subclass they actually require, which would drop every one of these codes at once. That is a public type on the experiment classes, so I did not do it here.

## The others

`self.idata` is assigned `None` in `__init__` with no annotation, so mypy inferred it as exactly `None`, which is why returning it read as returning None. Declared `xr.DataTree | None`. One of the two fit methods then still failed because it guarded `self.idata` in place, leaving it Optional again where the branches rejoined; it now works on a local and assigns once, which is what the sibling methods that already type-checked do.

`causal_impact` was declared `| None` because the OLS branch could assign None when no coefficient matched the interaction term. That branch no longer exists on this base: the coefficient canonicalization from #1060 replaced the backend-identity branching with a single `coefficients().sel(...)`, and the `next(...)` there raises rather than yielding None. So the annotation drops `| None`, and the guard that had gone dead with it, `if self.causal_impact is not None else 0.0` in the plotting path, goes too. An earlier version of this PR added a `ValueError` in the OLS branch; rebasing onto the current base removed the code it lived in, so that is no longer part of the change.

One behaviour difference does remain, on an error path rather than the happy path. `PropensityScore.fit` now assigns `self.idata` once at the end instead of in stages, so if the predictive sampling raises, the attribute keeps its previous value rather than holding a posterior with no predictive groups attached. Same result whenever `fit` completes. It is noted in the code as well; say if you would rather keep the staged assignment.

`Axes.get_figure` is typed `Figure | SubFigure | None`. A caller-supplied Axes always belongs to a figure, and this function returns a Figure, so a SubFigure would be wrong here too. Narrowed with `assert isinstance`, the idiom already used in `checks/convex_hull.py`.

`Axes.annotate` wants `tuple[float, float]` and `np.mean` returns `np.floating`. Older numpy and matplotlib stubs report that as an `arg-type` error and newer ones do not, so the same commit under the same pinned mypy passed in one project environment and failed in the other. Wrapping the value in `float()` removes the version sensitivity, so the gate answers from the code rather than from whichever stubs the environment resolved.

## Not in this PR

`warn_unused_ignores` would make the inline ignores self-cleaning, and I wanted it on, since it is the thing that stops an ignore outliving its cause. It currently reports 33 pre-existing unused ignores across 10 modules, so it needs its own change rather than riding along here. There is a note to that effect next to the setting.

Tests remain outside the checked set, which would add 73 errors in 19 files.

## Verification

- `mypy` with an empty allowlist: `Success: no issues found in 54 source files`, against both project environments (numpy 2.3.5 with matplotlib 3.10.8, and numpy 2.4.6 with matplotlib 3.11.1).
- Full suite: `2294 passed, 18 skipped, 3 deselected`.

Stacked on a feature branch, so the test matrix only arrives once the stack merges down.

### DIFF
--- a/causalpy/checks/operating_characteristics.py
+++ b/causalpy/checks/operating_characteristics.py
@@ -501,7 +501,10 @@ class OperatingCharacteristics:
             figure, ax = plt.subplots(figsize=(8, 5))
             ax.set_title(title, fontweight="bold", fontsize=11)
         else:
-            figure = ax.get_figure()
+            parent_figure = ax.get_figure()
+            # DECISION (#1127): narrowing assert, the idiom already used in checks/convex_hull.py. get_figure is typed `Figure | SubFigure | None`; None only happens for an artist never added to a figure, and this function returns a Figure, so a SubFigure would be wrong here too.
+            assert isinstance(parent_figure, plt.Figure)
+            figure = parent_figure
             ax.set_title(title, fontweight="bold", fontsize=11)
         ax.fill_between(
             effects, 0, nondetection, color="#94a3b8", alpha=0.30, label="Non-detection"
--- a/causalpy/experiments/diff_in_diff.py
+++ b/causalpy/experiments/diff_in_diff.py
@@ -118,7 +118,8 @@ class DifferenceInDifferences(BaseExperiment):
         model: PyMCModel | RegressorMixin | None = None,
     ) -> None:
         super().__init__(model=model)
-        self.causal_impact: xr.DataArray | float | None
+        # DECISION (#1127): no ``| None``. This is a bare annotation, so nothing ever assigns None; ``input_validation`` guarantees the interaction term that ``algorithm`` reads back, and reading before that already raises AttributeError rather than returning None.
+        self.causal_impact: xr.DataArray | float
         # to_pandas returns a copy, so index metadata is normalized on an owned frame rather than the caller's.
         pandas_data = to_pandas(data)
         pandas_data.index.name = "obs_ind"
@@ -593,7 +594,7 @@ class DifferenceInDifferences(BaseExperiment):
             "causal\nimpact",
             xy=(
                 arrow_x,
-                np.mean([y_pred_counterfactual_scalar, y_pred_treatment_scalar]),
+                float(np.mean([y_pred_counterfactual_scalar, y_pred_treatment_scalar])),
             ),
             xycoords="data",
             xytext=(5, 0),
@@ -614,11 +615,7 @@ class DifferenceInDifferences(BaseExperiment):
                 fontsize=LEGEND_FONT_SIZE,
             )
         else:
-            causal_impact_value = (
-                _as_scalar(self.causal_impact)
-                if self.causal_impact is not None
-                else 0.0
-            )
+            causal_impact_value = _as_scalar(self.causal_impact)
             ax.set(
                 xlim=[-0.05, 1.1],
                 xticks=[0, 1],
--- a/causalpy/experiments/instrumental_variable.py
+++ b/causalpy/experiments/instrumental_variable.py
@@ -206,10 +206,11 @@ class InstrumentalVariable(BaseExperiment):
                     "eta": 2,
                     "lkj_sd": 1,
                 }
+        # DECISION (#1127): the ignores below have one cause. `self.model` is declared as the base model union, while this experiment requires `InstrumentalVariableRegression`, whose `fit` takes numpy arrays plus `Z`, `t` and `priors`. Narrowing the attribute is the real fix and would drop every one of these codes, but that is a public type on the experiment classes, so it is left as a separate decision. mypy reports argument mismatches against the argument's own line, hence the per-argument ignores.
         self.model.fit(  # type: ignore[call-arg,union-attr]
-            X=self.X,
+            X=self.X,  # type: ignore[arg-type]
             Z=self.Z,
-            y=self.y,
+            y=self.y,  # type: ignore[arg-type]
             t=self.t,
             coords=COORDS,
             priors=self.priors,
--- a/causalpy/experiments/inverse_propensity_weighting.py
+++ b/causalpy/experiments/inverse_propensity_weighting.py
@@ -135,7 +135,8 @@ class InversePropensityWeighting(BaseExperiment):
         Delegates to ``self.model.fit`` with the covariate matrix ``self.X``,
         treatment vector ``self.t``, and coordinate metadata ``self.coords``.
         """
-        self.model.fit(X=self.X, t=self.t, coords=self.coords)  # type: ignore[call-arg]
+        # DECISION (#1127): `arg-type` joins the `call-arg` ignore that was already here. Both come from the same cause: `self.model` is declared as the base model union, while this experiment requires `PropensityScore`, whose `fit` takes numpy arrays and the extra `t` argument. Narrowing the attribute is the real fix and would drop both codes, but that is a public type on the experiment classes. `warn_unused_ignores` will flag this line if that lands.
+        self.model.fit(X=self.X, t=self.t, coords=self.coords)  # type: ignore[call-arg,arg-type]
 
     def input_validation(self) -> None:
         """Validate the input data and model formula for correctness.
--- a/causalpy/pymc_models.py
+++ b/causalpy/pymc_models.py
@@ -252,7 +252,8 @@ class PyMCModel(pm.Model):
             case default priors are used.
         """
         super().__init__()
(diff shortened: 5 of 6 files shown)


drbenvincent (approved):
Agent-drafted review.

# PR #1149 Review

https://github.com/pymc-labs/CausalPy/pull/1149

Scope: `chore/mypy-shrink-allowlist` (`a758a563`, reviewed #1148) .. `chore/mypy-zero` (`d14a69f9`). Isolated worktree `/Users/benjamv/git/.worktrees/pr-1149`. Primary tree left on `issue-1092-lazy-experiment-lifecycle` with its existing dirty skill file. Ancestors and descendants excluded.

## Executive Summary
Recommendation: approve.

Value delivered:
- The mypy allowlist is gone. `make typecheck` is `Success: no issues found in 54 source files` with no `[[tool.mypy.overrides]]` left, so the gate now fails on any new package error instead of on a shrinking suppression table.
- Lifecycle types now match what the code already did at runtime: `PyMCModel.idata` is `xr.DataTree | None`, DiD `causal_impact` is `xr.DataArray | float` (not optional), and `OperatingCharacteristics.plot` narrows `Axes.get_figure()` to `plt.Figure` before returning it.

Review focus:
- Highest-risk assumption is that emptying the allowlist is only type bookkeeping. Probed the six named seams below. One real runtime difference exists (`PropensityScore.fit` assigns `idata` once at the end), it is documented in the PR and at the call site, and it is the better failure behavior. No must-fix.

## Findings

No must-fix items.

- [nit] `causalpy/experiments/instrumental_variable.py`: the touched `self.model.fit` line still carries `union-attr`, which is unused. Stripping the ignores produces `call-arg` (extra kwargs vs `PyMCModel` / `PyMCForecastModel.fit`) and `arg-type` on `X` and `y` only — no `union-attr`. Dropping that code on this line would be free; it is already in the 33 unused ignores deferred with `warn_unused_ignores`.
- [nit] `causalpy/pymc_models.py` `PropensityScore.fit`: `if idata is not None:  # pragma: no branch` is dead. `pm.sample` is typed `DataTree | MultiTrace | ZarrTrace` (confirmed on pymc 6.0.1). Deleting the guard still type-checks. `_fit_with_validated_data` already uses a raise-if-None narrow instead. The pragma is honest coverage accounting, not a hidden hole, but the branch does not need to exist.
- [nit] `causalpy/checks/operating_characteristics.py`: the comment cites `checks/convex_hull.py` as the same idiom; that file uses `assert isinstance(experiment, SyntheticControl)`, not `get_figure`. The assert itself is the right local pattern given `-> plt.Figure`.

What worked well:
- Each judgement is marked `DECISION (#1127)` at the line, which is the right shape for a draft of type-policy choices.
- Not narrowing `self.model` on IV/IPW in this PR is the correct split. Constructors already take `InstrumentalVariableRegression | None` and `PropensityScore | None`, but `BaseExperiment` assigns `self.model = adapter.model` typed `PyMCModel | RegressorMixin | PyMCForecastModel`. Changing that is a public attribute type, not an ignore tweak.
- DiD dropping `| None` and the `else 0.0` plot fallback matches the post-#1060 path: `input_validation` requires the group×post interaction, `algorithm` does `next(...)` + `coefficients().sel(...)`, and `next` raises rather than yielding None. Plotting before `algorithm` was already `AttributeError` (bare annotation, no assignment).
- `float(np.mean(...))` is required in this environment too, not only on “older” stubs: without it, local numpy 2.4.6 + matplotlib 3.10.9 still reports `tuple[Any, floating[Any]]` vs `tuple[float, float]` on `Axes.annotate`.

## Merge Readiness
Verdict: approve.
Branch status: 2 commits ahead of `chore/mypy-shrink-allowlist`, 0 behind; GitHub `MERGEABLE` / `CLEAN`.
CI status: green (Python 3.12/3.14 test matrix, prek, notebooks, docs, codecov/patch, zizmor, check-environment-yml, pre-commit.ci).

## PR Summary
This stack PR finishes the #1127 ratchet: empty the per-module allowlist by settling four lifecycle types (`idata`, `causal_impact`, IV/IPW `fit` ignores, `get_figure`) plus a stub-stable `float()` on the DiD annotate y-coordinate. The only user-visible trade-off is on an error path: if `PropensityScore` predictive sampling raises after `pm.sample` succeeds, `self.idata` now keeps its previous value instead of a posterior with no predictive groups. Happy-path `fit` is unchanged. `warn_unused_ignores` and test-file checking stay out of scope on purpose.

## Test Evidence
Remote: all `gh pr checks 1149` passing, including codecov/patch (report: all modified coverable lines covered; partials −1 vs base, matching the `pragma: no branch`).

Local, isolated worktree `/Users/benjamv/git/.worktrees/pr-1149` with `PYTHONPATH` at that tree (shared conda env not retargeted with `make setup`):
- `mamba run -n CausalPy make typecheck` on head: Success, 54 files (env mypy 2.3.1, numpy 2.4.6, matplotlib 3.10.9, pymc 6.0.1).
- `mypy --warn-unused-ignores`: 33 unused ignores in 10 files, matching the PR claim. One of them is the unused `union-attr` on the IV fit line this PR edits.
- Ignore-stripped mypy: IPW needs `call-arg` + `arg-type`; IV needs `call-arg` on extra kwargs plus `arg-type` on `X`/`y` (also vs `PyMCForecastModel.fit`); DiD annotate needs `float()`; OC `get_figure` is `assignment` without the assert; restoring `causal_impact | None` reopens `convert_to_string` `arg-type`.
- `idata` annotation dropped: return-value errors at `InstrumentalVariableRegression.fit` and `PropensityScore.fit` only. Annotation kept with the old in-place `if self.idata is not None`: PropensityScore `return-value` (`DataTree | None` vs `DataTree`). Assign-once, or deleting the None guard, both type-check.
- Not run: full pytest (remote matrix already green; this PR does not change happy-path sampling).

## Open Questions
None that affect merge of this PR against its declared base. Whether to keep the staged-`idata` error path is already answered in the code comments; say so only if you prefer the old partial posterior. Narrowing IV/IPW `self.model` and turning on `warn_unused_ignores` remain follow-ups, as stated.

## Probe notes (review brief)

Private hypotheses and outcomes:

1. `PropensityScore.fit` assign-once would leave callers with a missing posterior after a PPC failure — **partially true, accepted**. Direct callers are `InversePropensityWeighting.algorithm` and `fit_outcome_model` via `az.extract(self.idata)`. A raised PPC still fails the experiment; the difference is leftover state (previous complete `idata` or `None`) vs a posterior without predictive groups. Documented; better than a half-fit. No test pins the exception path (and should not, given MCMC).
2. DiD `causal_impact` could still be None on OLS — **falsified**. Backend-agnostic `coefficients().sel(coeffs=treatment_coefficient)` after `next` on labels; validation requires the interaction; the old `else 0.0` plot guard was dead.
3. IV/IPW ignore codes would be wrong or too wide — **mostly falsified**. Needed codes are exactly `call-arg` and `arg-type`. `union-attr` is unused leftover. Cause is `adapter.model`’s union, not the constructor parameter types.
4. `assert isinstance(..., plt.Figure)` would break SubFigure callers — **no in-repo callers**. `plot(..., ax=)` is untested (method is `pragma: no cover`; tests call `plot()` without `ax`). Return type already promised `Figure`. Detached axes would fail later on `tight_layout` anyway.
5. Cross-environment `float()` would be unnecessary on newer stubs — **falsified here**. This env still errors without `float()`.
6. `pragma: no branch` would hide a live branch — **falsified** against pymc’s return type. The nicer fix is deleting the `if`, not the pragma.

Draft GitHub review (not posted): none required beyond approval.

REVIEWED | /Users/benjamv/git/CausalPy/.scratch/pr-reviews/2026-09-08-author-anevolbap/pr-1149.md

PR #1166, opened 2026-08-10
fix: parse micromamba list --json output from micromamba 2.9

anevolbap:
The `test` job on `pymc6_and_pymcmarketing1_migration` has been red since 2026-08-08. All three matrix legs fail at `Fingerprint the resolved PyTensor C-cache runtime` with `AttributeError: 'str' object has no attribute 'get'`, before any test runs. The last green run on the branch is 194343ca (2026-08-04).

Cause: the step parses `micromamba list --json` and iterates the result as a list of package dicts. micromamba 2.9.0 wraps that list in an object (`{"log_history": [...], "packages": [...]}`), so iteration yields the two key strings instead. `setup-micromamba` is SHA-pinned but the micromamba binary it downloads is not, so CI moved to 2.9.0 on its own.

Fix: take `packages` from the object when the parsed JSON is a dict, keep the bare list otherwise, and pin `micromamba-version: 2.9.0-0` so the binary stops floating.

Verified by extracting the step's script and running its parsing block against both binaries in the local `CausalPy` env: 437 packages parsed under 2.9.0 and under 2.5.0, while the pre-fix code returns a `dict` under 2.9.0. Downloaded 2.6.0 and 2.8.0 as well, both print a bare list, so 2.9 is where the shape changed. `prek run --all-files` passes.

CI on this PR confirms the fix: `Fingerprint the resolved PyTensor C-cache runtime` is green on all three legs. The matrix is still red one step further along, at `Run tests`, and those ten failures are pre-existing rather than anything this change introduced. They reproduce locally on the base branch source (10 failed, 2283 passed) and were hidden behind the fingerprint break; #1167 fixes them.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,9 @@ jobs:
       - name: Setup micromamba environment
         uses: mamba-org/setup-micromamba@ce51e99f4bb8a82ab7158c4dc59ef4634c59c4f9 # v3.1.0
         with:
+          # Pinned: the micromamba binary is resolved at run time, so an
+          # unpinned version changes CI behaviour on an untouched branch.
+          micromamba-version: 2.9.0-0
           environment-file: environment.yml
           environment-name: CausalPy
           # Pin python to the matrix version. All other deps (including
@@ -154,7 +157,7 @@ jobs:
           ).stdout
           with compiler.open("rb") as source:
               compiler_sha256 = hashlib.file_digest(source, "sha256").hexdigest()
-          packages = json.loads(
+          listing = json.loads(
               subprocess.run(
                   ["micromamba", "list", "--json"],
                   check=True,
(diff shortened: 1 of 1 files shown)


drbenvincent:
## Automated triage

**Why:**
- Changes `.github/workflows/ci.yml` — the CI workflow that all PRs depend on. A bug here blocks the entire project.
- Fixes a silent breakage where micromamba 2.9 changed its `--json` output shape, causing all 3 test matrix legs to fail at the fingerprint step before any tests run.
- Adds a `micromamba-version` pin to prevent the same class of floating-dependency breakage in future.
- Verified locally but CI tests are still in-progress; prek + sync + zizmor + readthedocs all pass.

**Review focus:**
1. Confirm the JSON-shape parsing guard (`dict` vs `list`) handles both old and new micromamba output correctly.
2. Verify the `micromamba-version: 2.9.0-0` pin is the right target and matches the version the test matrix actually needs.
3. Check that the combined change doesn't break any of the 3 test matrix legs once they complete.

**Confidence:** high

PR #1172, opened 2026-08-19
fix: clear the stale type: ignore comments and gate warn_unused_ignores

anevolbap:
Branched on top of #1173, which is itself stacked on #1149. Clears the 33 unused `type: ignore` comments that #1149 left as a follow-up, and turns `warn_unused_ignores` on.

The base field still reads `chore/mypy-zero` because GitHub refuses to retarget a PR that is part of a stack, so the diff currently shows #1173's commit as well. That resolves itself: once #1173 merges into `chore/mypy-zero`, this PR narrows to its own single commit. **Merge #1173 first.**

All 33 carried explicit error codes, none bare, so removing them can't unmask a hidden error class.

CausalPy and CausalPy-pymc6 resolve different numpy and matplotlib stubs, so I checked both before deleting anything: same 33 lines, same 10 files, in both. Nothing needed to stay for one env only.

No source behaviour changes. Every deletion is a clean unused-ignore removal.

## Why it sits on #1173 rather than #1149

One of the 33 was only partly unused: `instrumental_variable.py` carried `type: ignore[call-arg,union-attr]` and mypy flagged only `union-attr`, so this PR originally narrowed it to `[call-arg]`. #1173 narrows `self.model` to the concrete subclass, which removes that ignore and its neighbours entirely, so the two PRs were editing the same line in incompatible ways. Rebasing onto #1173 drops that hunk from this PR: it now touches 10 files instead of 11, and the remaining 32 deletions are independent of it.

The dependency runs the other way too, which is why these cannot be siblings. Turning `warn_unused_ignores` on requires that line to already be clean, and only #1173 makes it clean.

## Verification

- `mypy==2.3.0 --warn-unused-ignores`: `Success: no issues found in 54 source files`, in both envs, after the rebase.
- Full suite: `2294 passed, 18 skipped, 3 deselected`.

Stacked on a feature branch, so the test matrix only arrives once the stack merges down.

### DIFF
--- a/causalpy/checks/mccrary.py
+++ b/causalpy/checks/mccrary.py
@@ -91,7 +91,7 @@ class McCraryDensityTest:
         rd = experiment
         threshold = rd.treatment_threshold  # type: ignore[attr-defined]
         running_var = rd.running_variable_name  # type: ignore[attr-defined]
-        data = rd.data  # type: ignore[attr-defined]
+        data = rd.data
 
         x = data[running_var].values
         below = x[x < threshold]
--- a/causalpy/checks/placebo_in_time.py
+++ b/causalpy/checks/placebo_in_time.py
@@ -528,7 +528,7 @@ class PlaceboInTime:
             return self.intervention_length
 
         treatment_time = experiment.treatment_time  # type: ignore[attr-defined]
-        data = experiment.data  # type: ignore[attr-defined]
+        data = experiment.data
 
         treatment_end = getattr(experiment, "treatment_end_time", None)
         if treatment_end is not None:
@@ -1042,13 +1042,9 @@ class PlaceboInTime:
             raise ValueError("expected_effect_prior is not set.")
         if hasattr(prior, "rvs"):
             if self.random_seed is None:
-                return np.asarray(prior.rvs(n))  # type: ignore[union-attr]
+                return np.asarray(prior.rvs(n))
             if self._rvs_accepts_random_state(prior):
-                return np.asarray(
-                    prior.rvs(  # type: ignore[union-attr]
-                        n, random_state=self._rng_for_stage(0)
-                    )
-                )
+                return np.asarray(prior.rvs(n, random_state=self._rng_for_stage(0)))
 
             prior_type = f"{type(prior).__module__}.{type(prior).__qualname__}"
             if unseeded_custom_priors is not None:
@@ -1065,7 +1061,7 @@ class PlaceboInTime:
                 "marks its type as unseeded.",
                 stacklevel=2,
             )
-            return np.asarray(prior.rvs(n))  # type: ignore[union-attr]
+            return np.asarray(prior.rvs(n))
         raise TypeError(
             f"expected_effect_prior must have an .rvs(n) method, got "
             f"{type(prior).__name__}."
@@ -1147,7 +1143,7 @@ class PlaceboInTime:
         unseeded_custom_priors: list[dict[str, str]] = []
         factory = self._get_factory(context)
         treatment_time = experiment.treatment_time  # type: ignore[attr-defined]
-        data = experiment.data  # type: ignore[attr-defined]
+        data = experiment.data
         intervention_length = self._compute_intervention_length(experiment)
         required_pre_period_rows = self._get_intervention_window_observation_count(
             data, treatment_time, intervention_length
--- a/causalpy/experiments/interrupted_time_series.py
+++ b/causalpy/experiments/interrupted_time_series.py
@@ -274,13 +274,13 @@ class InterruptedTimeSeries(BaseExperiment):
             # Validate treatment_end_time > treatment_time
             # Type check: we've already validated both match the index type, so they're compatible
             # NOTE: Both treatment_time and treatment_end_time are INCLUSIVE (>=) in their respective periods
-            if treatment_end_time <= treatment_time:  # type: ignore[operator]
+            if treatment_end_time <= treatment_time:
                 raise ValueError(
                     f"treatment_end_time ({treatment_end_time}) must be greater than treatment_time ({treatment_time})"
                 )
             # Validate treatment_end_time is within data range
             # NOTE: treatment_end_time is INCLUSIVE, so it can equal data.index.max()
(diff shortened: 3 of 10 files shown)


drbenvincent (approved):
Agent-drafted review.

# PR #1172 review

https://github.com/pymc-labs/CausalPy/pull/1172

**Scope:** own commit `8fa0517a0c98adeeca7f4d2bec1d87aee8a30596` only (`3338e2a8..8fa0517a`). Ancestor `3338e2a8` is #1173 (already approved) and is excluded from this review. GitHub’s cumulative diff is 11 files because the declared base is still `chore/mypy-zero`; the own-commit delta is 10 files.

**Classification:** mixed cleanup + infrastructure. Dominant risk is unused `type: ignore` deletion plus `warn_unused_ignores` gating. No runtime behaviour change.

## Executive Summary

Recommendation: **approve**.

Value delivered:

- 32 stale, code-tagged `type: ignore` comments are gone across 9 library modules. Parent-commit mypy with `--warn-unused-ignores` reports exactly those 32 `unused-ignore` errors in the same files; HEAD typecheck is clean, so none of the deletions were still suppressing a real error in the CausalPy env (mypy 2.3.1 locally; CI uses the pinned 2.3.0).
- `[tool.mypy] warn_unused_ignores = true` is now on, and `make typecheck` / CI `make typecheck` read that table. A later fix that removes an underlying error will fail typecheck unless the ignore is deleted with it. That is the self-cleaning property #1149 left as follow-up.

Review focus:

- Highest-risk assumption: a deleted ignore is still required in some supported stub set (pandas 2.3 vs 3, numpy/matplotlib). Evidence against: (1) every deleted ignore is unused at the #1173 parent in this env; (2) HEAD mypy is `Success: no issues found in 54 source files` with the new gate; (3) remaining ignores in the same files (`attr-defined` on RD/ITS-specific attributes) still fire, so the gate is not a no-op; (4) remote Test matrix (Python 3.12 pandas 2.3, 3.12 pandas 3.x, 3.14 pandas 3.x) all ran `make typecheck` and passed.

## Findings

Must-fix: none.

- [nit] `causalpy/formula_utils.py`: the comment above `isetitem` still says pandas-stubs omit a Series argument. That was the rationale for `type: ignore[arg-type]`, which this commit correctly deletes as unused. The comment is now slightly stale; optional follow-up to drop or reword it. Not merge-blocking.

What worked well:

- Stacking on #1173 so `instrumental_variable.py` is not in this commit. Parent unused-ignore output has **no** `instrumental_variable.py` lines; GitHub’s extra 11th file is only the #1173 ancestor.
- Neighbour ignores that are still needed were left in place: `McCraryDensityTest` and `PlaceboInTime` still ignore `treatment_threshold` / `running_variable_name` / `treatment_time` / `post_impact` (`attr-defined` on `BaseExperiment`), while `.data` ignores were removed because `BaseExperiment.data: pd.DataFrame` is already declared.
- Every removed ignore carried an explicit error code. Enabling `warn_unused_ignores` at the parent without deletions fails typecheck (`Found 32 errors in 9 files`), which is the pinning check for the gate itself.

## Merge Readiness

Verdict: **approve**

Branch status: MERGEABLE / CLEAN. Head is **1 behind** `origin/chore/mypy-zero` (`d14a69f9`, `pragma: no branch` on a different `pymc_models.py` hunk) and 2 ahead (`3338e2a8` + `8fa0517a`). The behind commit does not conflict with this commit’s `model_coords["coeffs"]` ignore deletion. Author instruction stands: **merge #1173 first**, then this PR’s GitHub diff collapses to the own commit. Do not re-review #1173 as part of merging this one.

CI status: **green**. Observed checks all pass: Test matrix (3.12/pandas 2.3, 3.12/pandas 3.x, 3.14/pandas 3.x), `prek`, `pre-commit.ci`, `check-environment-yml`, notebooks, `codecov/patch`, Read the Docs, `zizmor`, `sync`. Codecov’s PR comment still shows a 96.875% patch figure with a partial on `pymc_models.py`; the required `codecov/patch` check is pass (threshold 96%). The partial is the pre-existing `if self._exog_var_names:` branch, not a new uncovered line from this ignore deletion.

No human review comments to address. No secrets, public API, or docs contract changes.

## PR Summary

This commit deletes 32 unused line-level `type: ignore` comments (10 files including `pyproject.toml`) and turns on `warn_unused_ignores` so remaining ignores stay honest. It exists because #1149 typed the package cleanly but left unused ignores that would immediately fail that gate. It is stacked on #1173 because one of the original 33 ignores lived on an IV `fit()` line that #1173 eliminates entirely; turning the gate on without #1173 would have been incompatible with that line. There is no user-facing behaviour change. Trade-off: reviewers looking at GitHub still see #1173 until that PR merges and this one is retargeted.

## Test Evidence

Local (isolated worktree `/Users/benjamv/git/.worktrees/pr-1172`, `mamba run -n CausalPy`, mypy 2.3.1, numpy 2.4.6, matplotlib 3.10.9, pandas 3.0.5):

- At `8fa0517a` (`warn_unused_ignores = true`): `make typecheck` → `Success: no issues found in 54 source files`.
- At parent `3338e2a8` with CLI `--warn-unused-ignores`: 32 `unused-ignore` errors in 9 files, matching the own-commit deletions 1:1 (transforms 11, IPW 4, staggered DiD 4, panel 3, ITS 2, placebo 5, McCrary 1, formula_utils 1, pymc_models 1). No leftover unused ignore in `instrumental_variable.py`.
- Remaining production ignores under `causalpy/` (excluding tests) are still live under the new gate; if any of those were unused, HEAD typecheck would have failed.

Remote: all `gh pr checks 1172` jobs pass, including typecheck on every Test matrix leg.

Not run: full pytest (no runtime change); no local `CausalPy-pymc6` env, so the author’s dual-stub claim was not reproduced here. Residual risk is low given pandas 2.3 and 3.x CI typecheck and the identical unused set at the #1173 parent.

## Open Questions

None that affect merge readiness of this commit. Process only: merge #1173 before this PR so GitHub’s file list and Codecov patch stop mixing the two.

REVIEWED /Users/benjamv/git/CausalPy/.scratch/pr-reviews/2026-09-08-author-anevolbap/pr-1172.md

PR #1173, opened 2026-08-19
fix: narrow self.model on InstrumentalVariable and InversePropensityWeighting

anevolbap:
Stacked on #1149. Settles the `self.model` narrowing that #1149 left as a separate decision, marked by the two `DECISION (#1127)` comments in `instrumental_variable.py` and `inverse_propensity_weighting.py`.

Both experiments require a specific `PyMCModel` subclass with a non-standard `fit()` signature: `InstrumentalVariableRegression` (extra `Z`, `t`, `priors`, ...) and `PropensityScore` (extra `t`). `self.model` is typed as `BaseExperiment`'s full model union (`PyMCModel | RegressorMixin | PyMCForecastModel`), so mypy checked the `fit()` calls against every member of that union instead of the concrete subclass, which is what produced the `call-arg`/`union-attr`/`arg-type` ignores.

Both subclasses already declare their constructor's `model` parameter and `_default_model_class` as that specific subclass, so nothing about the runtime contract changes. Adding `model: InstrumentalVariableRegression` / `model: PropensityScore` on the subclass states that in the type system too, so mypy resolves `fit()` against the concrete override and every ignore at that call site goes. `BaseExperiment.model` itself is untouched, so no other experiment's public type changes.

## What happens if the wrong model is passed

Nothing new. `make_model_adapter` only checks `isinstance(model, PyMCModel)` and `supports_bayes`, never the exact subclass. A caller who bypasses the type hints and passes the wrong `PyMCModel` still fails at the `fit()` call with the same runtime `TypeError` as before. No new validation added.

## Relationship to #1172

#1172 clears the 33 unused ignores and turns `warn_unused_ignores` on. It originally also narrowed the `instrumental_variable.py` ignore from `[call-arg,union-attr]` to `[call-arg]`, which is the same line this PR deletes outright. #1172 is now rebased on top of this branch and that hunk is gone from it, so the two no longer collide: this PR removes the ignores, #1172 assumes they are already gone. Merge order is this one first, then #1172.

## Verification

- `mypy`, full package, both project environments: `Success: no issues found in 54 source files` in each. The two resolve different stubs (numpy 2.3.5 / matplotlib 3.10.8, and numpy 2.4.6 / matplotlib 3.11.1), which is what makes checking both worthwhile.
- Full suite: `2294 passed, 18 skipped, 3 deselected`, plus the `InstrumentalVariable`/`InversePropensityWeighting` tests re-run clean in the second environment.

Being stacked on a feature branch, this gets the repo-wide workflows but not the test matrix. That arrives when the stack merges down.

### DIFF
--- a/causalpy/experiments/instrumental_variable.py
+++ b/causalpy/experiments/instrumental_variable.py
@@ -123,6 +123,7 @@ class InstrumentalVariable(BaseExperiment):
     supports_ols = False
     supports_bayes = True
     _default_model_class = InstrumentalVariableRegression
+    model: InstrumentalVariableRegression
 
     def __init__(
         self,
@@ -206,11 +207,10 @@ class InstrumentalVariable(BaseExperiment):
                     "eta": 2,
                     "lkj_sd": 1,
                 }
-        # DECISION (#1127): the ignores below have one cause. `self.model` is declared as the base model union, while this experiment requires `InstrumentalVariableRegression`, whose `fit` takes numpy arrays plus `Z`, `t` and `priors`. Narrowing the attribute is the real fix and would drop every one of these codes, but that is a public type on the experiment classes, so it is left as a separate decision. mypy reports argument mismatches against the argument's own line, hence the per-argument ignores.
-        self.model.fit(  # type: ignore[call-arg,union-attr]
-            X=self.X,  # type: ignore[arg-type]
+        self.model.fit(
+            X=self.X,
             Z=self.Z,
-            y=self.y,  # type: ignore[arg-type]
+            y=self.y,
             t=self.t,
             coords=COORDS,
             priors=self.priors,
--- a/causalpy/experiments/inverse_propensity_weighting.py
+++ b/causalpy/experiments/inverse_propensity_weighting.py
@@ -83,6 +83,7 @@ class InversePropensityWeighting(BaseExperiment):
     supports_ols = False
     supports_bayes = True
     _default_model_class = PropensityScore
+    model: PropensityScore
 
     def __init__(
         self,
@@ -135,8 +136,7 @@ class InversePropensityWeighting(BaseExperiment):
         Delegates to ``self.model.fit`` with the covariate matrix ``self.X``,
         treatment vector ``self.t``, and coordinate metadata ``self.coords``.
         """
-        # DECISION (#1127): `arg-type` joins the `call-arg` ignore that was already here. Both come from the same cause: `self.model` is declared as the base model union, while this experiment requires `PropensityScore`, whose `fit` takes numpy arrays and the extra `t` argument. Narrowing the attribute is the real fix and would drop both codes, but that is a public type on the experiment classes. `warn_unused_ignores` will flag this line if that lands.
-        self.model.fit(X=self.X, t=self.t, coords=self.coords)  # type: ignore[call-arg,arg-type]
+        self.model.fit(X=self.X, t=self.t, coords=self.coords)
 
     def input_validation(self) -> None:
         """Validate the input data and model formula for correctness.


drbenvincent (approved):
Agent-drafted review.

# PR 1173 Review

- PR: https://github.com/pymc-labs/CausalPy/pull/1173
- Author: anevolbap
- Head: `chore/mypy-narrow-iv-ipw-model` (`3338e2a8`)
- Scoped base: `chore/mypy-zero` (GitHub `baseRefOid` `aa2ae596`; current `#1149` / `origin/chore/mypy-zero` tip `d14a69f9`)
- Stack: follows approved `#1149`; `#1172` excluded from this review except merge-order notes
- Isolated worktree: `/Users/benjamv/git/.worktrees/pr-1173`
- Primary worktree left unchanged (`/Users/benjamv/git/CausalPy` on `issue-1092-lazy-experiment-lifecycle`)

## Review Brief (internal)

Value hypothesis:

- State on `InstrumentalVariable` / `InversePropensityWeighting` that `self.model` is the concrete PyMC subclass those experiments already require.
- Let mypy resolve `fit()` against `InstrumentalVariableRegression.fit` / `PropensityScore.fit` instead of `BaseExperiment`'s model union.
- Delete the `#1127` DECISION comments and the `call-arg` / `union-attr` / `arg-type` ignores those comments were holding.
- Leave `BaseExperiment.model` and every other experiment's public type unchanged.

Risk thesis:

- Class-level narrowing of a public instance attribute can lie relative to `BaseExperiment.__init__` → `make_model_adapter` → `adapter.model`, which still types and accepts any `PyMCModel`.

Attack plan:

1. Constructor / `_default_model_class` already restrict the type, but `super().__init__` assigns `adapter.model` typed as the full union / `PyMCModel` — narrowing is an assertion, not a checked assignment.
2. `make_model_adapter` still only checks `isinstance(model, PyMCModel)` + `supports_bayes`; a sibling `PyMCModel` still reaches `fit()` and fails there.
3. Public `.model` narrowing could break typed writers who assigned a different union member; readers should be compatible.
4. Removing ignores could hide remaining union issues if `fit()` still does not match the concrete signature.

## Executive Summary
Recommendation: approve.

Value delivered:

- `InstrumentalVariable.model` is now `InstrumentalVariableRegression` and `InversePropensityWeighting.model` is now `PropensityScore`, matching the constructor `model` parameter, `_default_model_class`, and docstring on both classes. A local mypy probe revealed exactly those types on instances.
- The only production `self.model.fit(...)` call sites in those modules type-check without ignores, because mypy now uses the subclass `fit` overrides (`Z`/`t`/`priors` on IV; extra `t` on IPW) instead of `PyMCModel.fit(X, y)` / the sklearn/forecast union.

Review focus:

- Highest-risk assumption: that the new annotations describe a runtime invariant rather than a wish. Evidence: constructors already typed the parameter this way; defaults instantiate those classes; `make_model_adapter` still does not check the exact subclass, which the PR states and which is unchanged. The annotation is the same contract the constructor already claimed, now applied to the attribute mypy actually uses at the `fit` call.

## Findings

No must-fix, should-fix, or nit findings.

Probes that did not become findings:

- **Producers.** `InstrumentalVariable.__init__` and `InversePropensityWeighting.__init__` still call `super().__init__(model=model)` with `model: <Concrete> | None`. `BaseExperiment.__init__` still does `self.model = adapter.model` after `make_model_adapter(...)`. For these two experiments, `model is None` instantiates `_default_model_class` (`InstrumentalVariableRegression` / `PropensityScore`). PyMC instances are wrapped in `PyMCModelAdapter` without cloning, so `adapter.model` is the same object.
- **Consumers.** The only experiment-side uses of `self.model` on these classes are the two `fit` calls this PR touches. Coefficient printing, idata, and IPW plotting go through `self._model_backend`, not `self.model`. Other experiments' `self.model` uses (`synthetic_control` pin, `panel_regression` predict) are untouched.
- **Wrong model at runtime.** `make_model_adapter` still accepts any `PyMCModel` when `supports_bayes` is true. OLS is already rejected (`test_olsmodel_and_iv`, `test_olsmodel_and_ipw`). A sibling PyMC class still fails at `fit()` (`TypeError` on unexpected keywords such as `Z`/`t`), as the PR describes. No new validation, no behavior change.
- **Public typing.** `BaseExperiment.model` is still inferred from `adapter.model` (`PyMCModel | RegressorMixin | PyMCForecastModel`). Narrowing is subclass-only, so `BaseExperiment`-typed code is unchanged. For the two experiment types, `.model` is a narrower read type (backwards-compatible for consumers). Typed assignment of a non-concrete model onto `iv.model` / `ipw.model` is now stricter, which is the intended static contract.
- **Unsoundness vs invariant.** The class annotation is not re-checked against the parent assignment (standard mypy subclass attribute pattern). Runtime can still store a sibling `PyMCModel` if the caller bypasses the constructor type. Under current mypy + `PyMCModel(pm.Model)`, sibling PyMC subclasses also did not error when assigned to `InstrumentalVariableRegression` in a probe — so the constructor annotation was already a weak static filter. That is pre-existing, not introduced here. The narrowing still does the job that motivated the DECISION comments: it selects the correct `fit` overload so the ignores can go.
- **Ignores.** Both DECISION (`#1127`) comments and every ignore on those two `fit` lines are gone. Remaining IPW ignores (`make_overlap_adjustments` / `make_doubly_robust_adjustment`) are unchanged and belong to `#1172`, not this PR.
- **Override / LSP.** mypy on the two production files plus `base.py` reported no incompatible-override error. Other experiments keep the wide attribute type.

## Merge Readiness
Verdict: approve.
Branch status: one commit behind current `#1149` / `origin/chore/mypy-zero` (`d14a69f9`, coverage pragma on `PropensityScore.fit` in `causalpy/pymc_models.py` only). No overlap with this PR's two files. GitHub reports `MERGEABLE` / `CLEAN`. Rebase onto the latest `#1149` tip is optional stack hygiene, not a merge blocker.
CI status: green on the checks this stacked PR actually runs (Read the Docs, pre-commit.ci, sync, welcome, zizmor). No pytest matrix, as the author notes for a non-`main` base.

## PR Summary

This is a two-file typing-only refactor: declare `model: InstrumentalVariableRegression` and `model: PropensityScore` on the two experiments that already require those classes and call their non-standard `fit` signatures directly. That is the decision `#1149` deferred. Runtime adapter validation, public `BaseExperiment.model`, and user-visible fitting behavior are unchanged. The trade-off is accepting a slightly more precise (and still not runtime-enforced) subclass attribute type so mypy can drop fit-site ignores; adding `isinstance` checks would be a separate behavior change.

## Test Evidence

Remote:

- `gh pr checks 1173`: docs/readthedocs.org:causalpy pass; pre-commit.ci pass; sync pass; welcome pass; zizmor pass.
- No human reviews or inline review comments. One Read the Docs bot issue comment (stack-wide HTML churn vs `latest`, not this two-file delta).
- Author claims: mypy clean in 54 files (two envs); `2294 passed, 18 skipped, 3 deselected`; `prek run --all-files` clean. Not re-run in full here.

Local (isolated worktree, scoped `origin/chore/mypy-zero...HEAD`):

- Diff is exactly the two experiment files, +6/-6 (git: 8 and 4 lines across the two files including context).
- `mamba run -n CausalPy mypy causalpy/experiments/instrumental_variable.py causalpy/experiments/inverse_propensity_weighting.py causalpy/experiments/base.py` → `Success: no issues found in 3 source files`.
- Temporary in-tree mypy probe (deleted afterward): `reveal_type(iv.model)` → `InstrumentalVariableRegression`; `reveal_type(ipw.model)` → `PropensityScore`. Full suite not run (typing-only, no runtime change; stacked PR has no test matrix).

`#1149` is `APPROVED` and still open. `#1172` GitHub still lists base `chore/mypy-zero`, but `origin/chore/mypy-zero-cleanup` already contains this PR's IV/IPW narrowing plus extra ignore deletions. Merge `#1173` before `#1172` as the author says so GitHub's `#1172` file list does not double-count this hunk.

## Open Questions

None that affect merge readiness.

## Draft GitHub comment (not posted)

Approve-shaped summary if a human wants to post later:

```markdown
Thanks — this is the right, minimal settlement of the `#1149` DECISION comments.

`InstrumentalVariable` / `InversePropensityWeighting` already typed the constructor `model` parameter and `_default_model_class` as `InstrumentalVariableRegression` / `PropensityScore`. Declaring the instance attribute the same way lets mypy resolve the non-standard `fit()` calls, so the `call-arg` / `union-attr` / `arg-type` ignores can go. `BaseExperiment.model` and adapter validation are untouched; a wrong `PyMCModel` still fails at `fit()` as before.

No request-changes items. Optional: rebase onto the latest `#1149` tip (`d14a69f9`) before merge; it does not touch these files.

Merge order remains this PR, then `#1172`.
```

REVIEWED /Users/benjamv/git/CausalPy/.scratch/pr-reviews/2026-09-08-author-anevolbap/pr-1173.md

Issue #475, opened 2025-05-26
Showcase use of splines in interrupted time series


drbenvincent:
At the moment we only have one example of using splines in CausalPy, and that's in a non-Bayesian example of the [regression discontinuity design](https://causalpy.readthedocs.io/en/stable/notebooks/rd_skl.html#using-basis-splines).

I would be interesting to see splines used more, particular in a Bayesian notebook. The easiest way to do this would be to extend the existing [Excess deaths due to COVID-19](https://causalpy.readthedocs.io/en/stable/notebooks/its_covid.html) notebook example which just has a basic linear trend. For example, we could modify the equation to `standardize(deaths) ~ 0 + cr(t, df=6) + C(month) + temp` to model the trend with a natural cubic spline.

I did give that a quick go, and was able to get a better pre-treatment fit ($R^2=0.75$) than the linear trend model ($R^2=0.71$), however there were $\hat{R}$ convergence problems. This could be also be related to the mismatch between the scale of the data and the (not cleverly chosen priors), and that is what probably lead me to use the `standardize` function in the `patsy` model formula. One solution to that might end up being better / automated priors (related to, but not the same as, #387).

It would also be interesting to optionally add in cyclic splines to model annual seasonality.

PR #1180, opened 2026-08-30
Spline showcase in Covid-19 ITS example

milesalanmoore:
This PR updates the COVID interrupted time-series notebook to show a spline-based alternative to a simple linear baseline. It adds a more flexible specification, demonstrates the diagnostics needed to assess model fit and convergence, also demonstrates a custom RW2 prior and Bayesian P-spline approach that leverages `pymc-extra`'s `VariableFactory` protocol and walks through the resulting counterfactual trend and estimated intervention effect. This addresses #475.

This is just a proof of concept / rough draft! I recognize the narrative still needs revision. For now: I wonder if something like this is what you had in mind for #475 @drbenvincent ? 

Thank you for thoughtfully maintaining a great library!


drbenvincent:
Nice. Sorry about the slight delay. Looking forward to getting to this soon @milesalanmoore

